PR #43, opened 2020-04-08
DJL UI prototype

mgubaidullin:
DJL UI prototype for training visualization

[README](https://github.com/mgubaidullin/djl/tree/djl-ui-prototype/ui)


frankfliu:
@mgubaidullin Thank you for creating such a great demo.

This implementation contains two parts:
1. UITrainListener - this looks like a candidate to be party of DJL library
2. Web serverice - this should be a separated. User can deploy it as a micro service, or use the library in their existing server.

I think the best way to moving forward is:
1. merge this code into djl-demo repo
2. we can write a blogpost about this feature.
3. We split UITrainListener out and see how can this better fix into DJL visualization plan, or create a standalone package.

Let me know what do you think.

mgubaidullin:
Do you mean something like that?

DJL UI

mgubaidullin:
mm

PR #59, opened 2020-04-28
Added a scaled dot product attention block to build transformers with.

chenkelmann:
## Description ##
Adds a new Block to perform scaled dot product attention as used by transformer models.

### Essentials ###

### Changes ###
- [x] ScaledDotProductAttentionBlock & integration test
- [x] A base block to inherit from when building new blocks, TransformerBaseBlock. 

## Comments ##
- Testing is currently only done by reproducing the response of the block when initialized with hardcoded values. The reference result is from the Tensorflow implementation of Bert. More tests will follow as soon as the necessary other transformer block are added
- The new base class is a suggestion on how to expand the AbstractBlock. IMHO directly inheriting from the AbstractBlock still forces devs of new blocks to write too much boilerplate. In the other Blocks for the "transformer topic" I need blocks with Parameters as well as other child blocks, so the other existing base blocks like ParameterBlock still result in too much code repetition.


chenkelmann:
The new NDArray.matMul seems not to be compatible to the the Tensorflow matMul. It did work with the manually called mx net "_npx_batch_dot", which seems to have the semantics of the TensorFlow matMul. I will add the batch dot operation to NDArray and then update this PR.

api/src/main/java/ai/djl/nn/transformer/ScaledDotProductAttentionBlock.java
```diff
@@ -0,0 +1,373 @@
...
+ * </ul>
+ *
+ * <p>Attention masks must contain a 1 for positions to keep and a 0 for positions to mask.
+ */
+public final class ScaledDotProductAttentionBlock extends TransformerBaseBlock {
+
+    private static final int VERSION = 1;
+
+    /** Size of the Word-/Token-embeddings we use the attention on. */
+    private final int embeddingSize;
```
frankfliu, line 65: Just FYI: We don't encourage add final for member/local and parameter variable. It's OK the leave the final, but it's a bit verbose and doesn't provide much value.

frankfliu: I fixed continuous build issue, rebase latest code should pass the tests.

chenkelmann: I did a merge this time as I already merged before and didn't want to run into conflicts due to the rebase. I will work with rebases instead of merges for future PRs.

chenkelmann: As far as the `final` modifier is concerened: You are right, in this case the benefits are not so huge. On whether it generally provides much value - I am afraid we have to agree to disagree here ;) 
I consider using final on members, parameters and local variables extensively one of the best coding practices. It helps avoid a lot of problems that come with mutable state and IMHO it encourages an immutable, more bug-resilient style. However, I will limit my use of final to more tricky control flow and classes with a more state and a more complex lifecycle. I removed the modifier in this case so the new Block fits in better with the existing code. It is a pity that the `final` behaviour isn't the default in Java - I wish there was  a `mutable` keyword instead...

api/src/main/java/ai/djl/nn/transformer/TransformerBaseBlock.java
```diff
@@ -0,0 +1,256 @@
...
+    public List<Parameter> getDirectParameters() {
+        return new ArrayList<>(parameters.values());
+    }
+
+    @Override
+    public void saveParameters(DataOutputStream os) throws IOException {
+        os.write(version);
+        for (Parameter parameter : parameters.values()) {
+            parameter.save(os);
+        }
```
zachgk, line 229: You may need to save/load some instance variables as well depending on the block (e.g. Embedding). Maybe add an additional saveInstanceVariables/load methods that can be overridden but default to nothing?

chenkelmann: Yes, I thought that as well, as looking at other blocks I saw that they store additional state. If nobody else is working on the merge and you are OK with it, I'd volunteer to do it and add methods for saving / loading instance state. Also I'd retrofit the existing blocks.

zachgk: That sounds great. Go ahead and update it

PR #61, opened 2020-05-02
Fixed typo

payne911:
I've fixed a typo in your `README.md` file : it was written "putput" instead of "output".
No need to fill out the automatic form that comes with PRs for this, afaik, since no code was affected.


frankfliu:
@payne911 thanks for your contribution.

PR #64, opened 2020-05-05
Expand abstract block api

chenkelmann:
## Description ##
This PR merges the functionality of `TransformerBaseBlock` into `AbstractBlock` and tries to use the newly available base API where possible to reduce code duplication in existing blocks. 

### Essentials ###
- [x] All changes have test coverage: (I am not sure about this, the existing tests should make sure the changes to parameter persistence should work)

### Changes ###
- [x] Merged `TransformerBaseBlock` into `AbstractBlock`
- [x] Changed return type of `Block.getDirectParameters()` from `List<Parameter` to `ParameterList` to be consistent with `Block.getParameters()`
- [x] Removed code duplication in child class
- [x] Changed usage of `Linear` block in `ScaledDotProductAttentionBlock` to `flatten=false` 

## Comments ##
I took great care to make changes backward compatible, but especially the serialization needs to be checked carefully when reviewing as well as the `MxNetSymbolBlock`, as these two topics were mostly affected by the change.


frankfliu:
@chenkelmann We are in the middle in 0.5.0 release, we'd better hold this PR after 0.5.0 release. There are some design choice we need dive deep in this PR. Will get back to you a bit later.

frankfliu:
@chenkelmann I rebased your changes on top of master, and made some minor changes. I pushed the code into block_refactor branch.

Most of changes looks good to me. Here is design choice I like to discuss.

In our origin design, we defined two type of block:
- Container, a container block only has children but has no parameters, it represent the topology of the network. In most case, user will just extends SequentialBlock as their base of custom block.
- ParameterBlock, it has no children and may or may not contains parameter. It represent basic operation/function. Implementing ParameterBlock usually is more complicated, and we expect contributor will keep enriching ParameterBlocks. and most end user will just consume them.

In your change, you make AbstractBlock can do both. It makes API more generic, but I'm not sure which is one is more clear for beginner.

zachgk:
@frankfliu I can see the argument that we need both parameters and children combined. The parameter block works well for operators that are fully implemented in the Engine. However, if you have an operator that could have been in the Engine and you are instead writing it at the DJL level, then it is likely to use both parameters and children. As long as we can keep the costs of creating a new Block class down, I think merging this would be good.

chenkelmann:
@frankfliu This is of course a tricky question and depends on who you want to adress. I can only give some feedback on my own experience here as someone who was new to the framework but has knowledge on deep learning and java. I actually found that separation into multiple base classes very confusing, but that may also be as I am very much used to the Tensorflow 1.x way of things, where you have an enormous flexibility. It took me several attempts to write my own blocks and I always failed because I would always need features from multiple base classes. I switched from one approach to another and there was always something missing. Also, there is already a lot of base classes, interfaces etc., so I think one base class you always use and need to know  and learn is easier to deal with than five or more (`AbstractBlock`, `ParameterBlock`, `SequentialBlock`, `ParallelBlock`, `LambdaBlock`...). Especially when that separation prohibits certain models, unless you go down to "bare metal". When I finally gave up and worked directly with `AbstractBlock` things started to work but I ended up with a ridiculous amount of boiler plate. I get the feeling that so far all architecture choices where made mainly with Image Processing in mind, where the current API is a good fit. However the latest transformer / NLP papers come up with more and more complex architectures which are hard to build with the current API. On a very personal note (this is just gut feeling), I think judging from the latest papers and tool developments the trend is more towards "derivable computing" and less "layer based deep learning". Then again I am not the typical consumer for your API, as I want to implement complex new models from scratch and train them myself...

api/src/main/java/ai/djl/nn/AbstractBlock.java
```diff
@@ -122,7 +124,7 @@ public int getVersion() {
      * @return the block given as a parameter - that way the block can be created and reassigned to
      *     a member variable more easily.
      */
-    protected <B extends Block> B addChildBlock(String name, B block) {
+    protected final <B extends Block> B addChildBlock(String name, B block) {
```
frankfliu, line 127: These method is called in constructor, they must be final

chenkelmann: @frankfliu Do you want me to change that?

frankfliu: @chenkelmann I already added final to those methods. I just add comment here to show why this final is necessary.

PR #72, opened 2020-05-19
Fixed leak of native memory caused by closing pointers of ndarrays st…

chenkelmann:
…ill in use.

## Description ##
The mxnet engine performs operations asynchronously. Before freeing an NDArray it is necessary to wait for operations on the array to finish, otherwise we get continuous memory leaks and random crashes.

### Essentials ###

### Changes ###
- Re-Added waiting for operations on NDArrays before closing

## Comments ##
There obviously was a reason this was commented out in the first place - is that reason fixed? But without this call, the engine simply crashes randomly, which probably is worse than what it was commented for.

### DIFF
--- a/mxnet/mxnet-engine/src/main/java/ai/djl/mxnet/engine/MxNDArray.java
+++ b/mxnet/mxnet-engine/src/main/java/ai/djl/mxnet/engine/MxNDArray.java
@@ -1534,8 +1534,7 @@ public class MxNDArray extends NativeResource implements LazyNDArray {
         }
         Pointer pointer = handle.getAndSet(null);
         if (pointer != null) {
-            // TODO: remove after fixing multi-thread data loading issue
-            // JnaUtils.waitToRead(pointer);
+            JnaUtils.waitToRead(pointer);
             JnaUtils.freeNdArray(pointer);
             manager.detach(getUid());
             manager = null;


mxnet/mxnet-engine/src/main/java/ai/djl/mxnet/engine/MxNDArray.java
```diff
@@ -1712,8 +1712,7 @@ public void close() {
         }
         Pointer pointer = handle.getAndSet(null);
         if (pointer != null) {
-            // TODO: remove after fixing multi-thread data loading issue
-            // JnaUtils.waitToRead(pointer);
+            JnaUtils.waitToRead(pointer);
```
frankfliu, line 1715: Does this fix memory leak?
MXNet backend execute NDArray operation asynchornously. waitToRead is a blocking call, this has significant performance impact.
If we can have a repo, we can work with MXNet to to resolve this issue.

frankfliu: We are aware of waitToRead() can prevent some random crash :(

Would you please invite @stu1130 to your repo as well. We will try to work with MXNet team and see if we can resolve it at MXNet level.

stu1130: Back to the date we debugged this issue, the conclusion is that when the MXNet C++ engine try to write the data back to the NDArray which is freed by DJL frontend, the crash happen. So the waitToRead() should be added to the MXNet C++ engine side. Now we have the reproducible script, as @frankfliu said, we can work with MXNet to fix this issue.

chenkelmann: An alternative to this PR for now is `LazyNDArray.waitForAll()` before a close. That way, this behaviour does not need to become default.

frankfliu: waitForAll() actually is more problematic:
* This affect all ndarray globally in MXNet engine
* This all suppress all native exceptions, will make debug super hard.

chenkelmann: @frankfliu thanks for the advice, I have replaced the `waitForAll` with a single `waitToRead` on a "bottleneck tensor" (e.g. the output of the previous transformer layer). This also does the trick. Should we keep this PR open or should we close it and create a Bug Issue instead, as this seems not to be the preferable solution?

keerthanvasist: We tested with the different examples we have, it doesn't look like there is a noticeable impact on performance of this change. We can go ahead and merge this change.

PR #110, opened 2020-06-24
Fixing Validation Epoch Result bug.

aksrajvanshi:
Addition: 
- When performing EasyTrain.fit() and gathering the validation epoch metrics, the number of occurrences of a particular evaluator collected are more than the number of epochs run. This change fixes the bug in the EvaluatorTrainingListener class.

### DIFF
--- a/api/src/main/java/ai/djl/training/listener/EvaluatorTrainingListener.java
+++ b/api/src/main/java/ai/djl/training/listener/EvaluatorTrainingListener.java
@@ -79,13 +79,16 @@ public class EvaluatorTrainingListener implements TrainingListener {
         Metrics metrics = trainer.getMetrics();
         for (Evaluator evaluator : trainer.getEvaluators()) {
             float trainValue = evaluator.getAccumulator(TRAIN_EPOCH);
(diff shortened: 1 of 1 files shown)


api/src/main/java/ai/djl/training/listener/EvaluatorTrainingListener.java
```diff
@@ -79,13 +79,16 @@ public void onEpoch(Trainer trainer) {
...
             if (metrics != null) {
                 String key = metricName(evaluator, TRAIN_EPOCH);
                 metrics.addMetric(key, trainValue);
+                String validateKey = metricName(evaluator, VALIDATE_EPOCH);
+                metrics.addMetric(validateKey, validateValue);
             }
 
             latestEvaluations.put("train_" + evaluator.getName(), trainValue);
-            float validateValue = evaluator.getAccumulator(VALIDATE_EPOCH);
+            validateValue = evaluator.getAccumulator(VALIDATE_EPOCH);
```
zachgk, line 91: The two definitions of validateValue are the same. You should delete the second

PR #113, opened 2020-06-29
LeNet Section 6.6 implemented.

aksrajvanshi:
Addition: LeNet Section 6.6 implementation.


jupyter/d2l-java/chapter_convolutional-neural-networks/lenet.ipynb
```diff
@@ -0,0 +1,577 @@
...
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "NDManager manager = NDManager.newBaseManager();\n",
+    "\n",
+    "SequentialBlock block = new SequentialBlock();\n",
+    "\n",
+    "block.add(\n",
```
zachgk, line 191: Can you run this through a code formatter in IntelliJ?

jupyter/d2l-java/chapter_convolutional-neural-networks/lenet.ipynb
```diff
@@ -0,0 +1,577 @@
...
+    "                    .addEvaluator(new Accuracy()) // Model Accuracy\n",
+    "                    .addTrainingListeners(TrainingListener.Defaults.logging()); // Logging\n",
+    "\n",
+    "    Model model = Model.newInstance(Device.defaultDevice().toString());\n",
+    "    model.setBlock(block);\n",
+    "    Trainer trainer = model.newTrainer(config);\n",
+    "    \n",
+    "    float epochLoss = 0f;\n",
+    "    float accuracyVal = 0f;\n",
+    "    long totalTime = 0;\n",
```
zachgk, line 418: I'm thinking it might be better to use EasyTrain.fit here. I think we are past the point of training from scratch at this stage in the book and we should focus on practical training and the model designs. You could also use the DJL Accuracy instead of reimplementing it above

jupyter/d2l-java/chapter_convolutional-neural-networks/lenet.ipynb
```diff
@@ -0,0 +1,577 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "NDArray X = manager.randomUniform(0f, 1.0f, new Shape(1, 1, 28, 28));\n",
+    "block.initialize(manager, DataType.FLOAT32, X.getShape());\n",
```
zachgk, line 254: Can you use getOutputShapes() in this section instead of forward? It would mean that you can skip the initialization and move it down with the training

jupyter/d2l-java/chapter_convolutional-neural-networks/lenet.ipynb
aksrajvanshi: So the only reason I went for the normal scratch method was to get the examples/sec on gpu stats. If that isn't required, I was initially moving to the EasyTrain method only. 
If we can get that stat in some other way, I can do that also.

zachgk: We record the time per batch to a metric in the TimeMeasureTrainingListener. You can use metrics.mean and divide by batch size to get the time per example. This information might also be printed by the logging training listener

jupyter/d2l-java/chapter_convolutional-neural-networks/lenet.ipynb
```diff
@@ -246,19 +228,20 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 7,
+   "execution_count": 18,
    "metadata": {},
    "outputs": [],
    "source": [
     "NDArray X = manager.randomUniform(0f, 1.0f, new Shape(1, 1, 28, 28));\n",
     "block.initialize(manager, DataType.FLOAT32, X.getShape());\n",
```
zachgk, line 236: You don't need this or the `block.setInitializer` because you initialize down in the trainer. Also, Xavier (or a more recent variant of it) is the default initializer in the TrainingConfig so you don't need to explicitly set it

jupyter/d2l-java/chapter_convolutional-neural-networks/lenet.ipynb
```diff
@@ -180,49 +179,32 @@
...
-    "    Linear.builder()\n",
-    "    .setOutChannels(84)\n",
-    "    .build())\n",
-    ".add(Activation::sigmoid)\n",
-    ".add(\n",
-    "    Linear\n",
-    "    .builder()\n",
-    "    .setOutChannels(10)\n",
-    "    .build());\n",
+    "block.add(Conv2D.builder()\n",
```
zachgk, line 190: This still looks a bit off. The block level methods are at the same indentation level as the Conv2D level methods.

jupyter/d2l-java/chapter_convolutional-neural-networks/lenet.ipynb
```diff
@@ -402,68 +389,50 @@
     "    Loss loss = Loss.softmaxCrossEntropyLoss();\n",
     "\n",
     "    LearningRateTracker lrt = LearningRateTracker.fixedLearningRate(lr);\n",
-    "    Optimizer sgd = Optimizer.sgd().setLearningRateTracker(lrt).setRescaleGrad(1.0f/batchSize).build();\n",
```
zachgk: Just to clarify: You might see some of these rescaleGrad by 1.0/batchSize in the book. You don't have to worry about it as much in DJL because we take the mean in the loss instead of the sum. Either method is to ensure your loss is constant with respect to the batch size. We just divide within the loss instead of the optimizer

aksrajvanshi: Yes, I have fixed those :)

jupyter/d2l-java/chapter_convolutional-neural-networks/lenet.ipynb
aksrajvanshi: So the block gives me an error in the next code block (where I'm printing the output shapes) if I don't initialize it.

PR #135, opened 2020-07-17
DenseNet Implementation Section 7.7

aksrajvanshi:
Addition: Implemented DenseNet Section 7.7

### DIFF
--- a/jupyter/d2l-java/chapter_convolutional-modern/densenet.ipynb
+++ b/jupyter/d2l-java/chapter_convolutional-modern/densenet.ipynb
@@ -85,7 +85,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%load ../utils/plot-utils.ipynb"
+    "%load ../utils/plot-utils.ipynb\n",
+    "%load ../utils/Training.java"
    ]
   },
   {
@@ -152,8 +153,7 @@
    "source": [
     "public SequentialBlock convBlock(int numChannels) {\n",
     "\n",
-    "        SequentialBlock block = new SequentialBlock();\n",
-    "        block\n",
+    "        SequentialBlock block = new SequentialBlock()\n",
     "                .add(BatchNorm.builder().build())\n",
     "                .add(Activation::relu)\n",
     "                .add(Conv2d.builder()\n",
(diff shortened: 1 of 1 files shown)


jupyter/d2l-java/chapter_convolutional-modern/densenet.ipynb
```diff
@@ -0,0 +1,632 @@
+{
```
zachgk, line 1: See my comments in https://github.com/awslabs/djl/pull/136

jupyter/d2l-java/chapter_convolutional-modern/densenet.ipynb
```diff
@@ -0,0 +1,632 @@
...
+    "trainer.initialize(new Shape(1, 1, 96, 96));"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "public void trainingChapter6(ArrayDataset trainIter, ArrayDataset testIter,\n",
```
zachgk, line 521: Not really for this PR, but we should consider moving this utility function out of these chapters instead of repeating it (shows up for the first usage in the book, otherwise reuse it from a utility class)

jupyter/d2l-java/chapter_convolutional-modern/densenet.ipynb
```diff
@@ -0,0 +1,606 @@
...
+    "    @Override\n",
+    "    public String toString() {\n",
+    "        return \"DenseBlock()\";\n",
+    "    }\n",
+    "\n",
+    "    @Override\n",
+    "    public NDList forward(ParameterStore parameterStore, NDList X, boolean training, PairList<String, Object> pairList) {\n",
+    "\n",
+    "        NDArray Y;\n",
+    "        for (Block block : this.net.getChildren().values()) {\n",
```
lanking520, line 213: What does this line do? Not sure if we can replace by using Parallel block since it seemed to parallel run all elements

lanking520: image
After checking, it seemed you can use a parallel block to represent each layer to do the execution work.

stu1130: The for loop is doing the following thing (+ means concat)
```
iter 0: f1(X) + X
iter 1: f2(f1(X) + X) + X
iter 2: f3(f2(f1(X) + X) + X)
...
```

stu1130: I would say there is no easier solution than the current one

PR #199, opened 2020-10-09
Return null device if requesting gpu but non available on system

waicool20:
## Description ##
Intuitively I thought that `Device.gpu()` would either throw an exception or return null in the case of no gpu on the system, apparently not. Here is a small patch to fix that. It also makes cases like `Device.defaultIfNull(Device.gpu())` work (Although we do have `Device.defaultDevice()` for that

### Essentials ###

### Changes ###
- [x] `Device.of()` returns null if requesting gpu but none are available on the system

### DIFF
--- a/api/src/main/java/ai/djl/Device.java
+++ b/api/src/main/java/ai/djl/Device.java
@@ -59,18 +59,20 @@ public final class Device {
         if (Type.CPU.equals(deviceType)) {
             return CPU;
         }
-        if (Type.GPU.equals(deviceType)) {
-            if (getGpuCount() == 0) {
-                return null;
-            }
-            try {
-                CudaUtils.getComputeCapability(deviceId);
-            } catch (Exception e) {
-                return null;
-            }
-        }
         String key = deviceType + '-' + deviceId;
-        return CACHE.computeIfAbsent(key, k -> new Device(deviceType, deviceId));
+        return CACHE.computeIfAbsent(key, k -> {
+            if (Type.GPU.equals(deviceType)) {
+                if (getGpuCount() == 0) {
+                    throw new IllegalArgumentException("Unable to create GPU Device, no GPUs available");
+                }
+                try {
+                    CudaUtils.getComputeCapability(deviceId);
+                } catch (Exception e) {
+                    throw new IllegalArgumentException("Unable to create GPU Device with ID " + deviceId + ": " + e.getMessage());
+                }
+            }
+            return new Device(deviceType, deviceId);
+        });
     }
 
     /**
--- a/api/src/main/java/ai/djl/Device.java
+++ b/api/src/main/java/ai/djl/Device.java
@@ -53,11 +53,14 @@ public final class Device {
      *
      * @param deviceType the device type, typically CPU or GPU
      * @param deviceId the deviceId on the hardware.
-     * @return a {@code Device} instance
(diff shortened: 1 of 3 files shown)


stu1130 (commented):
Thanks for your contribution!
Could you add a test on `DeviceTest.java`

waicool20:
Added another check to see if the gpu deviceId is valid by checking capability.

One part I'm not too sure about is whether or not the assignment of deviceId is related to the gpu count, check whether this is fine.

```
        if (Device.getGpuCount() > 2) {
            Assert.assertEquals(Device.gpu(3), Device.of("gpu", 3));
        } else {
            Assert.assertNull(Device.gpu(3));
        }
```

There also seems to be a mismatch between the way `CudaUtilsTest` and `Device` check for gpu count, `Device` checks if the engine is cuda capable, `CudaUtilsTest` does not. This was changed in 24afeab

api/src/main/java/ai/djl/Device.java
```diff
@@ -53,12 +53,22 @@ private Device(String deviceType, int deviceId) {
...
      * @param deviceId the deviceId on the hardware.
-     * @return a {@code Device} instance
+     * @return a {@code Device} instance, null if not available
      */
     public static Device of(String deviceType, int deviceId) {
         if (Type.CPU.equals(deviceType)) {
             return CPU;
         }
+        if (Type.GPU.equals(deviceType)) {
+            if (getGpuCount() == 0) {
```
frankfliu, line 63: getGpuCount() function is not free, it need make a JNA call to cuda driver, we can move into CACHE.computeIfAbsent().

I'm not a fan of returning null in this case, I think throwing IllegalArgumenetException is a better choice. User should check GPU availability before creating device

api/src/main/java/ai/djl/Device.java
```diff
@@ -53,12 +53,22 @@ private Device(String deviceType, int deviceId) {
...
+                return null;
+            }
+            try {
+                CudaUtils.getComputeCapability(deviceId);
```
frankfliu, line 67: I don't think this is necessary.
Do you see any case that getGpuCount() success but CudaUtils.getComputeCapability(deviceId) failure?

frankfliu (changes requested):

api/src/main/java/ai/djl/Device.java
waicool20: I think if deviceId is invalid, say -1 or a value greater than the number of gpu then it could fail.

waicool20:
@frankfliu Well now that I reference other frameworks, it seems like they allow the creation of devices that aren't valid (e.g don't exist on the system). For consistencies sake, I guess we don't need to merge this in, but I'll let you decide.

frankfliu:
> @frankfliu Well now that I reference other frameworks, it seems like they allow the creation of devices that aren't valid (e.g don't exist on the system). For consistencies sake, I guess we don't need to merge this in, but I'll let you decide.

I also don't think this improvement is necessary:
1. In most cases, user doesn't need create Device object manually. If user need manually create Device, I assume they know what they are doing.
2. Conceptually, a Device object doesn't have to be valid. It's caller's responsibility to make sure device is valid.
3. And I would prefer to keep it single.

Thanks for your effort. Let's keep it as is for now.

PR #215, opened 2020-10-14
Workaround JavaCPP bug

frankfliu:
## Channel for questions ##
Before or while filing an issue please feel free to join our [Slack channel](https://join.slack.com/t/deepjavalibrary/shared_invite/zt-ar91gjkz-qbXhr1l~LFGEIEeGBibT7w) to get in touch with development team, ask questions, find out what's cooking and more!

## Description ##
(Brief description of what this PR is about)

### Essentials ###
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- [ ] Code is well-documented: 
- [ ] To the my best knowledge, [examples](https://github.com/awslabs/djl/tree/master/examples) and [jupyter notebooks](https://github.com/awslabs/djl/tree/master/jupyter) are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, Java doc)
- [ ] Feature2, tests, (and when applicable, Java doc)

## Comments ##

### DIFF
--- a/tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/LibUtils.java
+++ b/tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/LibUtils.java
@@ -48,6 +48,8 @@ public final class LibUtils {
             logger.debug("Loading TensorFlow library from: {}", libName);
             String path = new File(libName).getParentFile().toString();
             System.setProperty("org.bytedeco.javacpp.platform.preloadpath", path);
+            // workaround javacpp physical memory check bug
+            System.setProperty("org.bytedeco.javacpp.maxPhysicalBytes", "0");
         }
         // defer to tensorflow-core-api to handle loading native library.
     }


zachgk (approved):

lanking520 (approved):

saudet:
Could you let me know what are the issues you've been encountering with this? Thanks!

frankfliu:
There are two issues:
1. DJL already close all resources we don't rely on GC to release native memory, with JavaCPP GC kick in, we saw high GC CPU usage.
2. JavaCPP maxBytes is assigned as max java heap size by default, which should be the available physical memory instead. For example, java starts with -Xmx2g, and I have 96G of physical memory (or 16G of GPU) memory, I should be able allocate 16G of tensor in native code. But JavaCPP throws exception.

We have to disable maxBytes check. I don't think use max heap size as limit is a good default value.

saudet:
> There are two issues:
> 
> 1. DJL already close all resources we don't rely on GC to release native memory, with JavaCPP GC kick in, we saw high GC CPU usage.

That shouldn't happen, unless maxPhysicalBytes is reached that is. Is this what is happening, it reaches the limit and starts doing GC, but without a limit, it doesn't do GC, correct? If that's what it's doing, that's not a bug, it's working as intended.

> 2. JavaCPP maxBytes is assigned as max java heap size by default, which should be the available physical memory instead. For example, java starts with -Xmx2g, and I have 96G of physical memory (or 16G of GPU) memory, I should be able allocate 16G of tensor in native code. But JavaCPP throws exception.
> 
> 
> We have to disable maxBytes check. I don't think use max heap size as limit is a good default value.

Yeah, I'm not sure what would be a good default. It depends on the use cases for sure, but since the JDK itself uses Runtime.maxMemory as its default for off-heap buffers, I suppose that isn't such a bad default...

PR #281, opened 2020-11-06
test: setScalar with a scalar should update the ndarray correctly

kimim:
##  Description ##
setScalar with correct number should update the ndarray correctly. Thus in this PR, we will add an assertion.

### Essentials ###
- [X] Changes are complete (i.e. I finished coding on this PR)
- [X] Code is well-documented: 

### DIFF
--- a/integration/src/main/java/ai/djl/integration/tests/ndarray/NDIndexTest.java
+++ b/integration/src/main/java/ai/djl/integration/tests/ndarray/NDIndexTest.java
@@ -148,7 +148,9 @@ public class NDIndexTest {
     public void testSetScalar() {
         try (NDManager manager = NDManager.newBaseManager()) {
             NDArray original = manager.create(new float[] {1, 2, 3, 4}, new Shape(2, 2));
+            NDArray expected = manager.create(new float[] {0, 2, 3, 4}, new Shape(2, 2));
             original.setScalar(new NDIndex(0, 0), 0);
+            Assert.assertEquals(original, expected);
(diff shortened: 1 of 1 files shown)


integration/src/main/java/ai/djl/integration/tests/ndarray/NDIndexTest.java
```diff
@@ -148,7 +148,9 @@ public void testSetNumber() {
     public void testSetScalar() {
         try (NDManager manager = NDManager.newBaseManager()) {
             NDArray original = manager.create(new float[] {1, 2, 3, 4}, new Shape(2, 2));
+            NDArray expected = manager.create(new float[] {0, 2, 3, 4}, new Shape(2, 2));
             original.setScalar(new NDIndex(0, 0), 0);
+            Assert.assertEquals(original, expected);
```
stu1130, line 153: nice catch, could we also add one more Assert after line154?

kimim: Line 154 will throw an exception, because we try to setScalar to more than 1 elements. This behavior is indicated on line 147:

```java
@Test(expectedExceptions = IllegalArgumentException.class)
```

kimim: otherwise, we can split these two tests(the normal case and the exception case) in different tests.

lanking520 (approved):

stu1130 (approved):

PR #326, opened 2020-11-14
Add AmazonReview dataset to basicdata

frankfliu:
## Channel for questions ##
Before or while filing an issue please feel free to join our [Slack channel](https://join.slack.com/t/deepjavalibrary/shared_invite/zt-ar91gjkz-qbXhr1l~LFGEIEeGBibT7w) to get in touch with development team, ask questions, find out what's cooking and more!

## Description ##
(Brief description of what this PR is about)

### Essentials ###
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- [ ] Code is well-documented: 
- [ ] To the my best knowledge, [examples](https://github.com/awslabs/djl/tree/master/examples) and [jupyter notebooks](https://github.com/awslabs/djl/tree/master/jupyter) are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, Java doc)
- [ ] Feature2, tests, (and when applicable, Java doc)

## Comments ##

### DIFF
--- a/basicdataset/src/test/resources/mlrepo/dataset/nlp/ai/djl/basicdataset/amazon_reviews/metadata.json
+++ b/basicdataset/src/test/resources/mlrepo/dataset/nlp/ai/djl/basicdataset/amazon_reviews/metadata.json
@@ -1,0 +1,42 @@
+{
+  "metadataVersion": "0.2",
+  "resourceType": "dataset",
+  "application": "nlp",
+  "groupId": "ai.djl.basicdataset",
+  "artifactId": "amazon_reviews",
+  "name": "amazon_reviews",
+  "description": "A sentiment analysis dataset of Amazon reviews",
+  "website": "https://s3.amazonaws.com/amazon-reviews-pds/readme.html",
+  "artifacts": [
+    {
+      "version": "1.0",
+      "snapshot": false,
+      "name": "amazon_reviews_us_Digital_Software",
+      "properties": {
+        "dataset": "us_Digital_Software"
+      },
+      "files": {
+        "amazon_reviews": {
+          "uri": "https://s3.amazonaws.com/amazon-reviews-pds/tsv/amazon_reviews_us_Digital_Software_v1_00.tsv.gz",
+          "sha1Hash": "b8390100b92579ed814eede4112514417e339902",
+          "size": 18997559
+        }
+      }
+    },
+    {
+      "version": "1.0",
+      "snapshot": false,
+      "name": "amazon_reviews_us_Software",
+      "properties": {
+        "dataset": "us_Software"
+      },
+      "files": {
+        "amazon_reviews": {
+          "uri": "https://s3.amazonaws.com/amazon-reviews-pds/tsv/amazon_reviews_us_Software_v1_00.tsv.gz",
+          "sha1Hash": "e48346dd356698ce680e385e3ecf07501de695b8",
+          "size": 94010685
+        }
+      }
+    }
+  ]
+}
--- a/examples/src/main/java/ai/djl/examples/training/transferlearning/TrainAmazonReviewRanking.java
+++ b/examples/src/main/java/ai/djl/examples/training/transferlearning/TrainAmazonReviewRanking.java
@@ -120,7 +120,7 @@ public final class TrainAmazonReviewRanking {
     private static CsvDataset getDataset(
             Arguments arguments, BertFullTokenizer tokenizer, int maxLength) {
         String amazonReview =
-                "https://alpha-djl-demos.s3.amazonaws.com/dataset/nlp/amazon_reviews_us_Digital_Software_v1_00.tsv.gz";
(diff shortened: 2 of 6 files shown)


lanking520:
Same point I would like to make here. We may consider keep this data outside. The dataset is just a subset used to demo an application, not standard dataset. Plus this maybe our own advanced use case for how to load multiple rows in a CSV dataset

zachgk:
Where does the data come from? Is it https://jmcauley.ucsd.edu/data/amazon/?

frankfliu:
> Where does the data come from? Is it https://jmcauley.ucsd.edu/data/amazon/?

It this it come from: https://s3.amazonaws.com/amazon-reviews-pds/readme.html

lanking520:
> > Where does the data come from? Is it https://jmcauley.ucsd.edu/data/amazon/?
> 
> It this it come from: https://s3.amazonaws.com/amazon-reviews-pds/readme.html

https://s3.amazonaws.com/amazon-reviews-pds/readme.html

And only contains digital software category

examples/src/main/java/ai/djl/examples/training/transferlearning/TrainAmazonReviewRanking.java
```diff
@@ -120,7 +120,7 @@ public static TrainingResult runExample(String[] args)
     private static CsvDataset getDataset(
             Arguments arguments, BertFullTokenizer tokenizer, int maxLength) {
         String amazonReview =
-                "https://alpha-djl-demos.s3.amazonaws.com/dataset/nlp/amazon_reviews_us_Digital_Software_v1_00.tsv.gz";
+                "https://s3.amazonaws.com/amazon-reviews-pds/tsv/amazon_reviews_us_Digital_Software_v1_00.tsv.gz";
```
frankfliu, line 123: I revert the example back. But I still think we should use our dataset:
1. In CsvDataset, we don't store the dataset locally, each time user has to re-download the gz file
2. Replacing CvsDataset with AmazonReview has no impact for user to understand this demo

basicdataset/src/test/resources/mlrepo/dataset/nlp/ai/djl/basicdataset/amazon_reviews/metadata.json
```diff
@@ -0,0 +1,42 @@
...
+      "files": {
+        "amazon_reviews": {
+          "uri": "https://s3.amazonaws.com/amazon-reviews-pds/tsv/amazon_reviews_us_Digital_Software_v1_00.tsv.gz",
+          "sha1Hash": "b8390100b92579ed814eede4112514417e339902",
+          "size": 18997559
+        }
+      }
+    },
+    {
+      "version": "1.0",
```
frankfliu, line 27: Amazon reviews is a popular collection of datasets, we should provide them.
I only added two here for now.

examples/src/main/java/ai/djl/examples/training/transferlearning/TrainAmazonReviewRanking.java
zachgk: Worst case, we could at least treat it as a third-party model zoo dataset (with the URL pointing to our first party one).

I think we could use this as a standard dataset if we pull the full original one. Then, we can do the filter on the Java end to only use the digital software category (as an option to speed up the demo).

frankfliu: Amazon Reviews is a collection of datasets, all different sub datasets are already stored in each .gz file: https://s3.amazonaws.com/amazon-reviews-pds/tsv/index.txt

What we kept is just the metadata that pointing to the original files. TensorFlow is doing the same thing.

zachgk: All of the data does follow the same format. So, combining the pieces should be as easy as just concatenating the data.

I think we have two options:
- Treat each of the items as a separate `Dataset` object constructed using the same class. They would all be loaded from artifacts in the same metadata file. Using multiple sub-datasets requires users to load each individually and concatenate the datasets themselves.
- Treat all of them as a unified `Dataset` object. When constructing the object, you can specify 1+ of the sub-datasets to load.

frankfliu: I prefer option 1, we can all all of them into the same metadata.json file, and use `optDatasetName()` to load each of them as I did in this PR.

The whole dataset are huge, current CSVReader doesn't support streaming load, it load the whole CVS into memory. I think it make sense to load each one separately.

If user need to combine some them, they should be able to create a CombinedDataset class which is them effort as we do to implement option 2.

lanking520 (approved):

zachgk (approved):

PR #400, opened 2020-12-03
Include demos as part of main docs

zachgk:
## Description ##

Right now, the demo docs are a separate and easily missed section of the docs.
Instead, the demos can be combined into the main docs site. This means that we
only need to support a single site and users can find everything on that site
whether it is integrative or not. Also importantly, it unifies the docs search
between the two sites.


stu1130 (approved):

roywei (approved):

lanking520 (approved):

lanking520:
Please trigger a docs test once this is merged

PR #410, opened 2020-12-07
[DLR] Create a new DlrPredictor to copy the model

stu1130:
## Description ##

### DIFF
--- a/dlr/dlr-engine/src/main/java/ai/djl/dlr/engine/DlrModel.java
+++ b/dlr/dlr-engine/src/main/java/ai/djl/dlr/engine/DlrModel.java
@@ -13,11 +13,11 @@
 package ai.djl.dlr.engine;
 
 import ai.djl.BaseModel;
-import ai.djl.Device;
 import ai.djl.Model;
-import ai.djl.dlr.jni.JniUtils;
+import ai.djl.inference.Predictor;
 import ai.djl.ndarray.NDManager;
 import ai.djl.ndarray.types.DataType;
+import ai.djl.translate.Translator;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.Files;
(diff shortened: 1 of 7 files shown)


dlr/dlr-engine/src/main/java/ai/djl/dlr/engine/DlrModel.java
```diff
@@ -32,6 +33,7 @@
  */
 public class DlrModel extends BaseModel {
 
+    private Device device;
```
frankfliu, line 36: We don't need this here, just call manager.getDevice() when needed.

examples/src/main/java/ai/djl/examples/inference/benchmark/util/AbstractBenchmark.java
```diff
@@ -288,10 +288,7 @@ public NDList processInput(TranslatorContext ctx, Object input) {
                         /** {@inheritDoc} */
                         @Override
                         public Object processOutput(TranslatorContext ctx, NDList list) {
-                            FloatBuffer fb = list.get(0).toByteBuffer().asFloatBuffer();
-                            float[] ret = new float[fb.remaining()];
-                            fb.get(ret);
-                            return ret;
+                            return list.get(0).toType(DataType.FLOAT32, false);
```
frankfliu, line 291: toType() is not supported by Ort and Dlr

dlr/dlr-engine/src/main/java/ai/djl/dlr/engine/DlrPredictor.java
```diff
@@ -0,0 +1,50 @@
...
+
+import ai.djl.Device;
+import ai.djl.dlr.jni.JniUtils;
+import ai.djl.inference.Predictor;
+import ai.djl.translate.Translator;
+
+/**
+ * {@code DlrPredictor} is special implementation of {@link Predictor} for DLR.
+ *
+ * <p>The native Dlr doesn't support multi-threading feature, when creating a new DlrPredictor, we
```
zachgk, line 23: Can this be done in the DlrModel? So, the model would track whether it is the first created predictor in the .newPredictor(...), and if not give it a new handle.

stu1130: Yes we can, but we need to override the getBlock to do implicitly copy, which is not ideal for end user. Let's say if user call DlrModel.getBlock() to play around the object they don't realize we reload another new TVM model.

frankfliu (approved):

PR #413, opened 2020-12-08
Create DJL-Easy

zachgk:
Creates easy wrapper over DJL.


settings.gradle
```diff
@@ -3,6 +3,7 @@ include ':api'
 include ':basicdataset'
 include ':examples'
 include ':model-zoo'
+include ':easy'
```
frankfliu, line 6: It's a bit strange call a module easy. may call it starter

tools/gradle/publish.gradle
```diff
@@ -3,6 +3,7 @@ configure([
         project(':basicdataset'),
         project(':model-zoo'),
         project(':serving'),
+        project(':easy'),
```
frankfliu, line 6: Add to publish later when we ready for release the package.

settings.gradle
zachgk: I was thinking of it similarly to EasyTrain (DJL-Easy in full).

I don't know if I like starter, because this isn't necessarily the start to using DJL. It is a higher level of abstraction over all of the deep learning provided by DJL. The goal is not to introduce them to DJL concepts, but to hide as many concepts as possible. Furthermore, I am afraid that referring to it as a starter might discourage others from treating it as a serious product. Ultimately, I think we can expect many users of DJL-easy to never have the time/inclination to learn enough deep learning to produce better results than DJL-Easy and to just keep using it indefinitely.

frankfliu: I agree starter may not be a good name, but, '`easy` is too general, it didn't provide enough context. We can discuss a little bit more. Here is some name I can think of:
- simple-train
- djl-simple
- easy-train
- 4dummies

easy/README.md
```diff
@@ -0,0 +1,39 @@
+# DJL Easy
```
lanking520, line 1: I am more lean towards `starter`. Like SpringBoot starter template

roywei (commented):
do we want a poll in slack or discussion forum for names?

easy/README.md
zachgk: https://github.com/awslabs/djl/pull/413#discussion_r538947968

zachgk:
We should first figure out if this will be just training or inference because that affects the name. So, can we merge this for now and then rename it later once we decide on a better name

frankfliu (approved):
Let merge it first and refactor from this.

PR #420, opened 2020-12-11
[dlr] Add engine dependency to dlr-native

frankfliu:
## Description ##


stu1130 (commented):
why do we need this

frankfliu:
The dependency between engine and native are bi-directional.
Our BOM solved partial of the problem, it only tell the maven with versions are compatible.
If user included pytorch-native:1.7.0, maven doesn't know it requires pytorch-engine:0.9.0, if we add this dependency in the pom file, maven will upgrade to 0.9.0.

This change requires us release both engine and native at the same time, but can reduce some confusion.

PR #425, opened 2020-12-12
[PyToch] Optimize the PyTorch tensor creation

stu1130:
The PR only tackles the tensor creation.

## Description ##

### DIFF
--- a/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
+++ b/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
@@ -54,8 +54,12 @@ public class PtNDArray extends NativeResource<Long> implements NDArray {
     private PtNDManager manager;
     private PtNDArrayEx ptNDArrayEx;
 
+    // keep a reference to direct buffer to avoid GC release the memory
+    private ByteBuffer dataRef;
+
     /**
-     * Constructs an PyTorch from a native handle (internal. Use {@link NDManager} instead).
+     * Constructs a PyTorch {@code NDArray} from a native handle (internal. Use {@link NDManager}
+     * instead).
      *
      * @param manager the manager to attach the new array to
      * @param handle the pointer to the native PyTorch memory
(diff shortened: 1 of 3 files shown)


pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
```diff
@@ -193,37 +213,38 @@ public void set(Buffer data) {
                     "size mismatch! the NDArray has size " + size() + " but set with size " + size);
         }
         if (data.isDirect() && data instanceof ByteBuffer) {
+            dataRef = (ByteBuffer) data;
```
frankfliu, line 216: set() should do the same to keep the buffer

pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_torch_creation.cc
```diff
@@ -36,7 +36,7 @@ JNIEXPORT jlong JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_torchFromBlob(JNI
   }
   // Don't change device unless data on CPU
   if (!device.is_cpu()) {
-    result = result.to(device);
+    result = result.to(device, false, true);
```
frankfliu, line 39: What is change for?

frankfliu (approved):

pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
```diff
@@ -54,8 +54,12 @@
     private PtNDManager manager;
     private PtNDArrayEx ptNDArrayEx;
 
+    // keep a reference to direct buffer to avoid GC release the memory
+    private ByteBuffer dataRef;
```
lanking520, line 58: if we close the NDArray, and not shut down the dataRef. Will the dataRef (pointer to native memory) still exist and waiting for GC to clean up?

frankfliu: Yes.
1. DirectBuffer has a different lifecycle than normal GC, it doesn't use heap
2. We are creating DirectBuffer before, now just holding it a bit longer.

stu1130: when we close the NDArray, we should mark the DirectByteBuffer to be ready for GC. As long as Java hold the reference, the data would be still there.

pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_torch_creation.cc
stu1130: we don't need this after I checked the source code

pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
lanking520: DirectBuffer are using Phantom reference instead. It will be collected with cleaner thread instead of gc. Should be safe then...

lanking520 (approved):

PR #449, opened 2020-12-18
Release 0.9.0

stu1130:
## Description ##


.github/workflows/release.yml
```diff
@@ -8,7 +8,9 @@ jobs:
     runs-on: ubuntu-18.04
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
```
frankfliu, line 11: We don't need this in master branch.
We actually don't even need it in release branch, we can select branch in the actions UI

roywei (approved):

.github/workflows/release.yml
stu1130: done

lanking520 (approved):

frankfliu (approved):

PR #473, opened 2020-12-29
Move the common utilities to project root for all jni usage

stu1130:
## Description ##


utils/native/utils.h
```diff
@@ -0,0 +1,91 @@
+/*
```
frankfliu, line 1: util folder name is too general.

I'm not sure where is the best location for jni shared code. I think we should be able to put them under api module:

api/src/main/native/shared/

frankfliu (approved):

PR #475, opened 2020-12-29
Fix AbstractBlock inputNames and inputShapes mismatch bug

frankfliu:
## Description ##


api/src/main/java/ai/djl/nn/AbstractBlock.java
```diff
@@ -296,6 +297,13 @@ public ParameterList getDirectParameters() {
      * @param inputShapes the expected shapes of the input
      */
     protected void beforeInitialize(Shape[] inputShapes) {
+        if (inputNames.isEmpty()) {
+            // automatically assing input names
```
stu1130, line 301: typo

stu1130 (approved):

PR #487, opened 2021-01-04
rewrite paddle engine to JNI

lanking520:
## Description ##

rewrite the PaddlePaddle Engine into JNI implementation


paddlepaddle/paddlepaddle-engine/src/main/java/ai/djl/paddlepaddle/engine/PpSymbolBlock.java
```diff
@@ -44,13 +47,18 @@ public NDList forward(
             NDList inputs,
             boolean training,
             PairList<String, Object> params) {
+        if (inputNames.length != inputs.size()) {
+            throw new IllegalArgumentException(
+                    "Input number mismatch, requires: " + Arrays.toString(inputNames));
+        }
+        // TODO: always clones new predictor
```
frankfliu, line 54:
```suggestion
```

paddlepaddle/paddlepaddle-engine/src/main/java/ai/djl/paddlepaddle/engine/PpSymbolBlock.java
```diff
@@ -18,23 +18,26 @@
...
 import ai.djl.nn.SymbolBlock;
-import ai.djl.paddlepaddle.jna.JnaUtils;
+import ai.djl.paddlepaddle.jni.JniUtils;
 import ai.djl.training.ParameterStore;
 import ai.djl.util.PairList;
+import java.util.Arrays;
 
 /** {@code PpSymbolBlock} is the PaddlePaddle implementation of {@link SymbolBlock}. */
-public class PpSymbolBlock extends AbstractBlock implements SymbolBlock {
+public class PpSymbolBlock extends AbstractBlock implements SymbolBlock, AutoCloseable {
```
frankfliu, line 27: We don't need make this AutoCloseable.

The PaddlePredictor will be closed in PpPredictor and PpModel.

paddlepaddle/paddlepaddle-engine/src/main/java/ai/djl/paddlepaddle/engine/PaddlePredictor.java
```diff
@@ -10,25 +10,26 @@
...
+public class PaddlePredictor extends NativeResource<Long> {
 
-/** The Paddle Native size handler. */
-@SuppressWarnings("missingjavadocmethod")
-public class NativeSize extends IntegerType {
-
-    private static final long serialVersionUID = 1L;
-
-    public static final int SIZE = Native.SIZE_T_SIZE;
+    PaddlePredictor(long handle) {
```
stu1130, line 21: can it be a simple JniUtils function call rather than a class?

lanking520: We have two options. Make it as a long and do all raw use including its life cycle. Or we wrap it up as an object.

frankfliu (approved):

paddlepaddle/paddlepaddle-engine/src/main/java/ai/djl/paddlepaddle/engine/PaddlePredictor.java
stu1130: the line of code seems to be almost the same. let's just keep it as it is

PR #509, opened 2021-01-11
Refactor request parsing to request parser class

ebamberg:
## Description ##

tiny refactoring extracting request parsing into a separate RequestParser class.

part of a bigger PullRequest as discussed with Frank in December. 


serving/src/main/java/ai/djl/serving/http/RequestParser.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
frankfliu, line 2:
```suggestion
 * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

frankfliu (approved):

PR #514, opened 2021-01-11
add minor fix and version bump ups

lanking520:
## Description ##

Minor fix on some unneccesary manager settings for MNIST with some documentation and other version bump-ups.

### DIFF
--- a/djl-easy/README.md
+++ b/djl-easy/README.md
@@ -34,6 +34,6 @@ You can pull the module from the central Maven repository by including the follo
 <dependency>
     <groupId>ai.djl</groupId>
     <artifactId>easy</artifactId>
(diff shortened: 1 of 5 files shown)


basicdataset/src/main/java/ai/djl/basicdataset/Mnist.java
```diff
@@ -52,7 +53,7 @@
 
     private Mnist(Builder builder) {
         super(builder);
-        this.manager = builder.manager;
+        this.manager = Engine.getInstance().newBaseManager(Device.cpu());
```
frankfliu, line 56: We probably need to keep the manager, since nobody will close this manager.

frankfliu (approved):

PR #518, opened 2021-01-13
Organize BasicDatasets into folders

zachgk:
This divides them into the folders:
- cv
- cv.classification
- nlp
- tabular


basicdataset/src/main/java/ai/djl/basicdataset/cv/classification/AbstractImageFolder.java
```diff
@@ -139,7 +139,7 @@ private boolean isImage(File file) {
         int maxDepth;
 
         protected ImageFolderBuilder() {
-            flag = Image.Flag.COLOR;
+            super();
```
frankfliu, line 142: You don't need call super() here

frankfliu (approved):

PR #529, opened 2021-01-14
Remove snapshot PyTorch Android build from nightly

stu1130:
## Description ##
We should use separate pipeline to publish android snapshot as the action here didn't really recompile the JNI

### DIFF
--- a/.github/workflows/snapshot_pytorch_android.yml
+++ b/.github/workflows/snapshot_pytorch_android.yml
@@ -1,0 +1,36 @@
+name: Snapshot PyTorch Android
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


.github/workflows/snapshot_pytorch_android.yml
```diff
@@ -0,0 +1,36 @@
+name: Snapshot PyToech Android
```
zachgk, line 1: To match our current naming convention, should this name (and file name) be "Native Snapshot PyTorch Android" and native_snapshot_pytorch_android.yml?

stu1130: this is for ai.djl.android package not for pytorch native(libtorch) or pytorch jni

zachgk (approved):

PR #531, opened 2021-01-15
Use custom NDK version instead of action default NDK

stu1130:
PyTorch Android native build only works on NDK version 20.
Action uses version 22 by default.


lanking520:
Have you tested the android pipeline?

stu1130:
> Have you tested the android pipeline?

I tested it on my local repo https://github.com/stu1130/djl/runs/1705801591?check_suite_focus=true you can see it only failed on upload

lanking520 (approved):

PR #549, opened 2021-01-21
Add NDArrayAdapter for partially implemented NDArray implementations

zachgk:


stu1130 (approved):

stu1130 (commented):
How about call it NoOpNDArrayAdapter

zachgk:
I named it to match `TrainingListenerAdapater`

api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
```diff
@@ -0,0 +1,949 @@
...
+import ai.djl.ndarray.types.Shape;
+import ai.djl.ndarray.types.SparseFormat;
+import java.nio.Buffer;
+import java.nio.ByteBuffer;
+
+/**
+ * A base implementation of the {@link NDArray} that does nothing. This can be used for overriding
+ * the NDArray with only part of the interface implemented.
+ */
+public interface NDArrayAdapter extends NDArray {
```
frankfliu, line 27: In java convention, Adapter should be a class with empty implementation.
This should not be an interface.

zachgk: I originally had it as a class, but the PpNDArray already extends NativeResource

frankfliu: Should we implement NativeResource functionally in MxNDArray and PpNDArray instead?

zachgk: I am not sure, especially as the NativeResource includes data as well. It might be more difficult that way

stu1130: We can refactor other NDArray classes except for MxNDArray, PpNDArray and make it abstract class

zachgk: PpNDArray is one of the ones this should help. Another option is we make NativeResource an interface, add a BaseNativeResource class, and keep a `NativeResource<T> handle = new BaseNativeResource(...)` as instance variables inside the MxNDArray and PpNDArray and implement NativeResource instead of extending BaseNativeResource. Then, the Adapter can be the class

lanking520: For all different cases, the NDArray either holds a Java native object (TF Tensor, Onnx Tensor, DLR Tensor) or JNI/JNA handle (MXNet, PpNDArray, PtNDArray). Maybe we can make NativeResource applies to all of them, just holds different objects...

frankfliu (approved):

api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
zachgk: In a call, we looked into this a bit. In DJL, BaseXXX classes are used to provide a base implementation of an interface while XXXAdapter is used only to provide empty implementations for methods in an interface. In NDArrayAdapter, it should be used only for NDArrays that don't plan to implement much of the NDArray interface. Given this limited scope for the adapter, we think it should be fine as an interface as it is

PR #554, opened 2021-01-21
Make RNN op generic

stu1130:
## Description ##

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/internal/NDArrayEx.java
+++ b/api/src/main/java/ai/djl/ndarray/internal/NDArrayEx.java
@@ -19,6 +19,7 @@ import ai.djl.ndarray.index.NDArrayIndexer;
 import ai.djl.ndarray.types.DataType;
 import ai.djl.ndarray.types.Shape;
 import ai.djl.nn.Activation;
+import ai.djl.nn.recurrent.RNN;
 import ai.djl.util.PairList;
 import java.util.List;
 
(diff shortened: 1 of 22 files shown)


api/src/main/java/ai/djl/ndarray/internal/NDArrayEx.java
```diff
@@ -313,67 +314,90 @@ NDList batchNorm(
...
+            NDArray input,
+            NDArray state,
+            NDList params,
+            boolean hasBiases,
+            int numLayers,
+            RNN.Activation activation,
+            double dropRate,
+            boolean training,
+            boolean bidirectional,
+            boolean batchFirst);
```
roywei, line 344: `batch_first` seems a confusing flag only for PyTorch, can we only use one behavior? Is there a specific use case for batch not first? 
Ref: https://github.com/pytorch/pytorch/issues/4145

api/src/main/java/ai/djl/nn/recurrent/RecurrentBlock.java
```diff
@@ -297,45 +190,56 @@ public T setStateSize(int stateSize) {
...
-            this.useBidirectional = useBidirectional;
+        public T optHasBiases(boolean hasBiases) {
+            this.hasBiases = hasBiases;
             return self();
         }
 
         /**
-         * Sets the optional parameter that indicates whether to have the states as symbol outputs.
+         * Sets the optional flag that indicates whether to return state or not. This is typically
+         * useful when you use RecurrentBlock in Sequential block.
```
roywei, line 236: provide default value in comments

roywei (commented):
Prefer to remove batch first option unless there is a solid usecase. Otherwise LGTM!

roywei (approved):
After discussion with Jake offline, let's keep batchFirst option for potential speed ups by setting it to `false`, and set default value to `true` for ease of use.

zachgk (approved):

PR #556, opened 2021-01-24
Fix recurrent block memory leak and output shape calculation

mpskowron:
## Description ##

1. Recurrent block implementation has a memory leak in `opInputs(...)` function, line 240:
`parameterList.add(array.flatten());`
`array.flatten()` is creating new array in model's NDManager, thus this array lives as long as the model. The analogical problem is also present in LSTM block.
2. Recurrent block's `getOutputShapes(...)` function returns incorrect shapes. Block accepts NTC not TNC (which is wrongly assumed as input shape in `getOutputShapes(...)` function).

### DIFF
--- a/api/src/main/java/ai/djl/nn/recurrent/LSTM.java
+++ b/api/src/main/java/ai/djl/nn/recurrent/LSTM.java
@@ -136,9 +136,9 @@ public class LSTM extends RecurrentBlock {
         NDList result = new NDList(head);
         try (NDList parameterList = new NDList()) {
             for (Parameter parameter : parameters.values()) {
-                NDArray array = parameterStore.getValue(parameter, device, training).duplicate();
+                NDArray array = parameterStore.getValue(parameter, device, training).flatten();
                 array.attach(manager);
-                parameterList.add(array.flatten());
+                parameterList.add(array);
             }
             NDArray array = NDArrays.concat(parameterList);
             result.add(array);
--- a/api/src/main/java/ai/djl/nn/recurrent/LSTM.java
+++ b/api/src/main/java/ai/djl/nn/recurrent/LSTM.java
@@ -16,6 +16,7 @@ import ai.djl.Device;
 import ai.djl.ndarray.NDArray;
 import ai.djl.ndarray.NDArrays;
 import ai.djl.ndarray.NDList;
+import ai.djl.ndarray.NDManager;
 import ai.djl.ndarray.internal.NDArrayEx;
 import ai.djl.ndarray.types.DataType;
 import ai.djl.ndarray.types.Shape;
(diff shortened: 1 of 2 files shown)


api/src/main/java/ai/djl/nn/recurrent/RecurrentBlock.java
```diff
@@ -227,22 +227,24 @@ protected NDList opInputs(ParameterStore parameterStore, NDList inputs, boolean
...
         inputs = updateInputLayoutToTNC(inputs);
         NDArray head = inputs.head();
+        NDManager manager = head.getManager();
         Device device = head.getDevice();
 
         NDList result = new NDList(head);
         try (NDList parameterList = new NDList()) {
             for (Parameter parameter : parameters.values()) {
-                NDArray array = parameterStore.getValue(parameter, device, training);
+                NDArray array = parameterStore.getValue(parameter, device, training).duplicate();
```
frankfliu, line 236: I don't think we need dup it,
We can just attach flattened array to manager.
```
                NDArray array = parameterStore.getValue(parameter, device, training).flatten();
                array.attach(manager);
                parameterList.add(array);
```

api/src/main/java/ai/djl/nn/recurrent/RecurrentBlock.java
```diff
@@ -164,17 +164,18 @@ protected void resetBeginStates() {
     /** {@inheritDoc} */
     @Override
     public Shape[] getOutputShapes(NDManager manager, Shape[] inputs) {
-        // Input shape at this point is TNC. Output Shape should be NTS
+        // Input shape at this point is NTC. Output Shape should be NTS
         Shape inputShape = inputs[0];
+        Long nShape = inputShape.get(0);
```
frankfliu, line 169:
```suggestion
        long nShape = inputShape.get(0);
```

api/src/main/java/ai/djl/nn/recurrent/RecurrentBlock.java
```diff
@@ -164,17 +164,18 @@ protected void resetBeginStates() {
...
+        Long tShape = inputShape.get(1);
```
frankfliu, line 170:
```suggestion
        long tShape = inputShape.get(1);
```

api/src/main/java/ai/djl/nn/recurrent/LSTM.java
```diff
@@ -128,27 +129,28 @@ protected NDList opInputs(ParameterStore parameterStore, NDList inputs, boolean
...
-        NDArray head = inputs.singletonOrThrow();
+        NDArray head = inputs.head();
+        NDManager manager = head.getManager();
         Device device = head.getDevice();
 
         NDList result = new NDList(head);
         try (NDList parameterList = new NDList()) {
             for (Parameter parameter : parameters.values()) {
-                NDArray array = parameterStore.getValue(parameter, device, training);
+                NDArray array = parameterStore.getValue(parameter, device, training).duplicate();
```
lanking520, line 139: Why we do duplicate here? Does that cause any issues?

api/src/main/java/ai/djl/nn/recurrent/RecurrentBlock.java
mpskowron: You're right, changed.

api/src/main/java/ai/djl/nn/recurrent/LSTM.java
mpskowron: Changed.

stu1130 (approved):
Hi @mpskowron Thanks for your contribution. I am also doing refactoring on the model to make it more generic for latter PyTorch & TensorFlow integration. So I am going to merge your  PR and rebase onto it.

frankfliu (approved):

mpskowron:
Sounds great! Thank you guys for a quick review.

PR #569, opened 2021-01-29
Add Rotation

lanking520:
## Description ##

Add rotation op to the ImageUtils. Also enforced Shape check on Image operators


api/src/main/java/ai/djl/ndarray/NDArray.java
```diff
@@ -2664,6 +2664,17 @@ default NDArray mean(int[] axes) {
...
+    /**
+     * Rotates an array by 90 degrees in the plane specified by axes.
+     *
+     * <p>Rotation direction is from the first towards the second axis.
+     *
+     * @param times Number of times the array is rotated by 90 degrees.
+     * @param axes The array is rotated in the plane defined by the axes. Axes must be different.
+     * @return the rotated NDArray
+     */
+    NDArray rot90(int times, int[] axes);
```
frankfliu, line 2676: better use full name, or use:
```
NDArray rotate(int degree, int[] axes)
```

And only support 90/180/270 for now

lanking520: rotate is vague, it essentially just do rotate 90 degree. And numpy use `rot90` as the name. We should keep them consistent.

pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_torch_isjm.cc
```diff
@@ -55,6 +55,16 @@ JNIEXPORT jlong JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_torchUnsqueeze(
   API_END_RETURN()
 }
 
+JNIEXPORT jlong JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_torchRot90
+        (JNIEnv* env, jobject jthis, jlong jhandle, jlong jk, jlongArray jdims) {
+    API_BEGIN()
```
stu1130, line 60: the indent should be 2 spaces here

api/src/main/java/ai/djl/ndarray/NDArray.java
stu1130: Does java have naming convention for those image transformation function? I am ok with either one

api/src/main/java/ai/djl/modality/cv/util/NDImageUtils.java
```diff
@@ -62,9 +62,26 @@ public static NDArray resize(
...
+            return image.rot90(times, new int[] {1 + batchDim, 2 + batchDim});
+        } else {
+            return image.rot90(times, new int[] {batchDim, 1 + batchDim});
+        }
+    }
+
     /**
      * Normalizes an image NDArray of shape CHW or NCHW with a single mean and standard deviation to
-     * apply to all channels.
+     * apply to all channels. TensorFlow enforce HWC instead.
```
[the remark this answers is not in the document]
lanking520: @roywei I think some of the TF model require NHWC inputs? Like the Resnet50.

frankfliu (approved):

PR #574, opened 2021-02-01
add paddle 2.0 official packages

lanking520:
## Description ##

Add Paddle 2.0.0 official packages (without mkldnn, openblas only)

### DIFF
--- a/.github/workflows/continuous.yml
+++ b/.github/workflows/continuous.yml
@@ -120,3 +120,31 @@ jobs:
         with:
           name: tensorflow-model-zoo
           path: tensorflow/tensorflow-model-zoo/build/reports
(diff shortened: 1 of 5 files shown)


.github/workflows/continuous.yml
```diff
@@ -120,3 +120,31 @@ jobs:
         with:
           name: tensorflow-model-zoo
           path: tensorflow/tensorflow-model-zoo/build/reports
+
+  # Windows platform for testing minor engines
```
frankfliu, line 124: Let's call them hybrid engines

frankfliu (approved):

zachgk (approved):

PR #579, opened 2021-02-02
Adding norm operator (similar to numpy.linalg.norm.html)

markbookk:
Adding the norm operator. This is similar to https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html. 

Implemented using MXNet and Tensorflow engine. This is my first operator and I had to do a few tricks to make it work so any feedback or tips on the code is greatly appreciated!


tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDArray.java
```diff
@@ -529,6 +530,66 @@ public NDArray erfinv() {
...
+    public NDArray norm(int[] axes) {
+        try (Tensor<?> tensor =
+                tf.linalg.euclideanNorm(getOperand(), tf.constant(axes)).asTensor()) {
+            return new TfNDArray(manager, tensor);
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public NDArray norm(boolean keepDims) {
```
stu1130, line 556:
```suggestion
    public NDArray norm(boolean keepDims) {
    if (dataType != DataType.FLOAT64) {
            throw new UnsupportedOperationException("float64 is not supported");
    }
    TfNDArray flattenTensor = (TfNDArray) flatten();
    try (Tensor<?> tensor =
                tf.linalg
                        .euclideanNorm(
                                flattenTensor.getOperand(),
                                tf.constant(0),
                                EuclideanNorm.keepDims(keepDims))
                        .asTensor()) {
            float number = tensor.rawData().asFloats().getFloat(0);
            long[] shapes = LongStream.generate(() -> 1).limit(shape.dimension()).toArray();
            // close the temp NDArray
            flattenTensor.close();
            return manager.create(new double[] {number}, new Shape(shapes));
    }
```

stu1130 (commented):
Thanks for your contribution!

api/src/main/java/ai/djl/ndarray/NDArrays.java
```diff
@@ -1989,4 +1989,110 @@ public static NDArray logicalXor(NDArray a, NDArray b) {
     public static NDArray erfinv(NDArray input) {
         return input.erfinv();
     }
+
```
zachgk, line 1992: We don't need to add norm to NDArrays. NDArrays only has methods that involve multiple NDArrays where it may be odd to associate the method with just the first one. Since norm only uses a single NDArray, we can just keep it on NDArray.

api/src/main/java/ai/djl/ndarray/NDArray.java
```diff
@@ -4415,4 +4415,98 @@ default String toDebugString(int maxSize, int maxDepth, int maxRows, int maxColu
...
+     * 5.
+     * jshell&gt; NDArray array = manager.create(new float[] {1f, 2f, 3f, 4f}, new Shape(2, 2));
+     * jshell&gt; array.norm();
+     * ND: () cpu() float32
+     * 5.4472
+     * </pre>
+     *
+     * @return the norm of this {@code NDArray}
+     */
+    NDArray norm();
```
zachgk, line 4437: You can add a default here with keepDims=false. Then, there are fewer methods to implement on the other classes.

api/src/main/java/ai/djl/ndarray/NDArrays.java
markbookk: I see. I'll remove it.

tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDArray.java
markbookk: Is there a specific reason you used `new double[]` on line 572 instead of `new float[]`? `number` is a float

markbookk: Also, thanks for this suggestion! It is much more cleaner.

api/src/main/java/ai/djl/ndarray/NDArray.java
```diff
@@ -4415,4 +4415,98 @@ default String toDebugString(int maxSize, int maxDepth, int maxRows, int maxColu
...
+     *
+     * @param axes If axes contains an integer, it specifies the axis of x along which to compute
+     *     the vector norms. If axis contains 2 integers, it specifies the axes that hold 2-D
+     *     matrices, and the matrix norms of these matrices are computed.
+     * @param keepDims keepDims If this is set to True, the axes which are normed over are left in
+     *     the result as dimensions with size one. With this option the result will broadcast
+     *     correctly against the original x.
+     * @return the norm of this {@code NDArray}
+     */
+    NDArray norm(int[] axes, boolean keepDims);
```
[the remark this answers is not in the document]
markbookk: I could make a new default function and that default calls this one. Is that okay or do you prefer to just change this one?

tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDArray.java
stu1130: I updated the code, it should be float

api/src/main/java/ai/djl/ndarray/NDArray.java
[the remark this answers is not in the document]
stu1130: I think we need a norm with all arguments, and this one have implementation for others to call

tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDArray.java
markbookk: Alright, thanks again

api/src/main/java/ai/djl/ndarray/NDArray.java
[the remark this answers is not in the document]
markbookk: Alright, I'll do that

api/src/main/java/ai/djl/ndarray/NDArray.java
markbookk: Done

api/src/main/java/ai/djl/ndarray/NDArray.java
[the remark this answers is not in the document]
markbookk: Done

stu1130 (approved):

lanking520 (approved):

PR #583, opened 2021-02-03
[Paddle] clean up and fix doc site build issue

lanking520:
## Description ##

Update the jupyter notebook and some code refactor on Paddle Engine.


paddlepaddle/paddlepaddle-native/build.gradle
```diff
@@ -273,33 +273,6 @@ publishing.repositories {
     }
 }
 
-
```
frankfliu, line 276: We will need this eventually, right?

lanking520: We already have `prepareNativeLib` which essentially doing the same thing

frankfliu (approved):

PR #597, opened 2021-02-04
Fixing model view bugs

### LINKED ISSUE #596: DJL Model View crashes on some Model Tabs.
## Description
When picking almost any model on the ModelView page and you click the Synset tab the model view page will crash. Refreshing the page is the only way to get the page back to a usable state. 

### Expected Behavior
Expected behavior would be either not showing the tab at all or showing that the current attribute isn't being used.

## How to Reproduce?
(If you developed your own code, please provide a short script that reproduces the error. For existing examples, please provide link.)

### Steps to reproduce
(Paste the commands you ran that produced the error.)

1. In your terminal go into the central folder(*cd central *). Then run gradle(* ,/gradlew*). 2.Go to http://localhost:8080/ in your web browser.
3.Click CV.ACTION_RECOGNITION

4.Click inceptionv3_ucf101
image

5.Then press the synset tab.
image

Fix ...... Coming soon.

anfee1:
## Description ##

This Pr is to close the issue with the ModelView page failing when a model has a null attribute. It utilizes error handling on the front end to simply display a more information to the user.

Fix:
image

fix #596

### DIFF
--- a/central/src/main/webapp/components/ModelView.jsx
+++ b/central/src/main/webapp/components/ModelView.jsx
@@ -148,6 +148,12 @@ export default function ModelView(props) {
 
 	console.log(props.model);
 
+	let noParameters = {
+		uri: "N/A",
+		sha1Hash: "N/A",
+		size: 0
+	};
+
 	let noSynset = {
 		name: "N/A",
 		Hash: "N/A",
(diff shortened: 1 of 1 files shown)


roywei (approved):
Thanks for the fix. LGTM!

PR #613, opened 2021-02-05
Serving configure idle timing

ebamberg:
## Description ##

configuration of worker-pool parameters like max_idle_time, max_batch_delay through rest-endpoint.
Default configuration for workers gets pick up from ConfigManager instead of using hardcoded-values.

ModelInfo changed to immutable to fix potential raise-condition when setting for example min, max worker per setter in model.

some smaller code refactorings to improve maintainability and readability.


frankfliu (approved):
Really nice refactor!

PR #617, opened 2021-02-06
Various fixes and improvements for MultiThreadedBenchmark

zemei:
* Collects used memory instead of committed for the JVM. The committed memory amount is [guaranteed](https://docs.oracle.com/javase/7/docs/api/java/lang/management/MemoryUsage.html) to be available to the JVM and not the currently used amount.
* Runs memory collection on a separate thread. Memory collection can be slow compared to inference which significantly affects the throughput measurement of the first worker thread.
* Collects memory before and after model loading to print separate base, model, and inference memory usage.


frankfliu (approved):
This looks great.

PR #623, opened 2021-02-08
Model View Arguments Fix

anfee1:
This is just an addendum to PR #596. Doing the same fix except with the arguments attribute.

### DIFF
--- a/central/src/main/webapp/components/ModelView.jsx
+++ b/central/src/main/webapp/components/ModelView.jsx
@@ -223,19 +223,19 @@ export default function ModelView(props) {
 						<DynForm data={props.model.properties} />
 					</TabPanel>
 					<TabPanel value={index} index={2}>
-						{data = props.model.arguments
+						{props.model.arguments
 							? <DynForm data={props.model.arguments} />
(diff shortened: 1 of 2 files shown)


central/src/main/webapp/components/ModelView.jsx
```diff
@@ -217,7 +223,10 @@ export default function ModelView(props) {
 						<DynForm data={props.model.properties} />
 					</TabPanel>
 					<TabPanel value={index} index={2}>
-						<DynForm data={props.model.arguments} />
+						{data = props.model.arguments
```
zachgk, line 226: Should this be assigning to a variable data or just returning the ternary?

anfee1: I could change how it is written so that it would make more sense like:
```
{
    props.model.arguments //Is this a truthy value
      ? <DynForm data={props.model.arguments} />
      : <DynForm data={noArguments}/>
}
```
It supposed to check to see if the value is not null and then assign it to the variable data if it isn't

frankfliu (approved):

roywei (approved):

PR #624, opened 2021-02-08
List input and output classes in ModelZoo lookup

zachgk:
This will list the input and output classes when trying to find a model in the
model zoo. It works either when no classes are specified or invalid classes are
specified. This will help users who don't know what the possible classes are to
resolve the question without looking at the code.

### DIFF
--- a/api/src/main/java/ai/djl/repository/zoo/BaseModelLoader.java
+++ b/api/src/main/java/ai/djl/repository/zoo/BaseModelLoader.java
@@ -36,6 +36,7 @@ import java.lang.reflect.Type;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


stu1130 (approved):

api/src/main/java/ai/djl/repository/zoo/BaseModelLoader.java
```diff
@@ -199,12 +200,24 @@ public String toString() {
     @SuppressWarnings("unchecked")
     private <I, O> TranslatorFactory<I, O> getTranslatorFactory(Criteria<I, O> criteria) {
         if (criteria.getInputClass() == null) {
-            throw new IllegalArgumentException("The criteria must set an input class");
+            throw new IllegalArgumentException(
```
frankfliu, line 203:
```suggestion
            throw new IllegalArgumentException(getErrorMessage("The criteria must set an input class.\n");
```

api/src/main/java/ai/djl/repository/zoo/BaseModelLoader.java
```diff
@@ -199,12 +200,24 @@ public String toString() {
...
+                    "The criteria must set an input class.\n" + printTranslatorFactories());
```
frankfliu, line 204:
```suggestion
```

frankfliu (approved):

PR #626, opened 2021-02-09
Add a new section in community tab

lanking520:
## Description ##

Add development section


zachgk:
I am not sure if this is the right place. Users are likely to click this button without realizing that it is going to take them away from the docs (and the docs' navigation). We also just take them to the development page, but the community docs should cover everything in the community overall. What is the particular situation you are thinking of?

lanking520:
Lu just mentioned we need a spot for developers to find our RFC/proposals

frankfliu (approved):

PR #628, opened 2021-02-09
Fix duplicate module name for native package

frankfliu:
## Description ##


lanking520 (commented):
What is the impact of the changes?

frankfliu:
With automatic module, user can use DJL in java-9 modularized application directly.

lanking520 (approved):

PR #638, opened 2021-02-10
Use PyTorch index to improve the performance

stu1130:
## Description ##

### DIFF
--- a/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
+++ b/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
@@ -267,6 +267,12 @@ public class PtNDArray extends NativeResource<Long> implements NDArray {
 
     /** {@inheritDoc} */
     @Override
+    public NDArray get(long... indices) {
+        return JniUtils.getItem(this, indices);
+    }
+
+    /** {@inheritDoc} */
+    @Override
     public void copyTo(NDArray array) {
         throw new UnsupportedOperationException("Not implemented");
     }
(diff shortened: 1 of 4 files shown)


pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
```diff
@@ -265,6 +265,12 @@ public void set(Buffer data) {
         JniUtils.set(this, buf);
     }
 
+    /** {@inheritDoc} */
```
zachgk, line 268: If we put it here, it may not activate for some calls to get that use the NDIndex directly. The more precise place would be to integrate it into the NDArrayIndexer. Maybe add a new NDIndexFull type to handle this?

lanking520 (approved):

pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
stu1130: I checked it but can't come up with a good way to avoid NDIndexFullPick.fromIndex and NDIndexFullSlice.fromIndex. Maybe we can merge it for now and later optimize it?

zachgk: It should avoid fullpick on its own because pick requires a pick NDIndexElement, while this is a fixed. Avoiding the slice is difficult, but there is an order of trying the NDIndexFull* inside the `NDArrayIndexer`. Just make your new one before `NDIndexFullSlice` in the order. In fact, maybe just make your new one first in the order. You may also need to throw an exception in the FullFixed get and catch it in the NDArrayIndexer get to make the FullFixed handling optional.

If you prefer, go ahead and merge it and then I can take a look at it later.

stu1130: ok Thank, I will go ahead to merge it and you can come back to this when you have time

PR #639, opened 2021-02-10
fix softmax flag behavior

### LINKED ISSUE #520: LogSumExp seems to be missing in SoftmaxCrossEntropyLoss (sparseLabel=false)
## Description
On latest MasterBranch (commit 4b516196) in SoftmaxCrossEntropyLoss (sparseLabel=false) line 85 
```
            loss = pred.mul(lab).neg().sum(new int[] {classAxis}, true);
```
looks like the LogSumExp term (red) is missing.
grafik

## Proposed correction
```
            int[] axes = new int[] {classAxis};
            NDArray max = pred.max(axes, true);
            NDArray logSumExp = max.add((pred.sub(max)).exp().sum(axes, true).log());
            loss = logSumExp.sub(pred.mul(lab).sum(axes, true));
```

roywei:
fix #520


zachgk (approved):

stu1130 (approved):
I saw all rnn tests are failing due to wrong output. maybe we need to re-calculate the output value

roywei:
Investigating the RNN test output. Loss is not supposed to be negative numbers to begin with.

PR #641, opened 2021-02-11
Fix windows lib_name not found during the uploading

stu1130:
## Description ##

### DIFF
--- a/.github/workflows/native_jni_s3_pytorch.yml
+++ b/.github/workflows/native_jni_s3_pytorch.yml
@@ -15,10 +15,13 @@ jobs:
         include:
           - operating-system: macos-latest
             os: osx
+            lib: libdjl_torch.dylib
           - operating-system: ubuntu-16.04
             os: linux
+            lib: libdjl_torch.so
           - operating-system: windows-latest
             os: win
+            lib: djl_torch.dll
     steps:
       - uses: actions/checkout@v1
       - name: Set up JDK 1.8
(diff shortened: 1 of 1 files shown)


.github/workflows/native_jni_s3_pytorch.yml
```diff
@@ -45,7 +48,7 @@ jobs:
           aws-region: us-east-2
       - name: Copy files to S3 with the AWS CLI
         run: |
-          aws s3 cp pytorch/pytorch-native/build/*djl_torch.* s3://djl-ai/publish/pytorch-1.7.1/jnilib/0.10.0/${{ matrix.os }}-x86_64/cpu/
+          aws s3 cp pytorch/pytorch-native/build/${{ matrix.lib }} s3://djl-ai/publish/pytorch-1.7.1/jnilib/0.10.0/${{ matrix.os }}-x86_64/cpu/
```
frankfliu, line 51: Is this causing proble?

stu1130: the window-cpu actually failed but still show pass 
The user-provided path pytorch/pytorch-native/build/*djl_torch.* does not exist.

frankfliu: I see, windows doesn't expend `* ` as linux do. it can only do djl_torch.*

frankfliu (approved):

PR #657, opened 2021-02-16
Revert partial of multithreading benchmark changes

frankfliu:
## Description ##


lanking520:
Why do we revert the change?

frankfliu:
> Why do we revert the change?

This cause GPU throughput degraded

stu1130 (approved):

PR #658, opened 2021-02-17
Serving input output classes

ebamberg:
## Description ##

you an register models for different input/output types now. This required to move forward with central and plugin architecture, cause we need to ability to use the models shown in central to be deployed and tested in the serving component, so it makes sense to combine the 2 projects into 1 project.

for now the model filters are simply http-parameter which makes usage easy for REST beginners.
Next step would be to accept kind of GAV-style parameter like "groupid:artifactid:version {..addition filters...}"

Real world example of the change can be found in ReadMe.MD

see first example Object detection.
we can register a model:

curl -X POST "http://localhost:8080/models?model_name=detect&application=cv/object_detection&filter=backbone:resnet50&input_type=ai.djl.modality.cv.Image&output_type=ai.djl.modality.cv.output.DetectedObjects"

predict the object detection:
curl -X POST http://127.0.0.1:8080/predictions/detect -F "data=@../examples/src/test/resources/dog_bike_car.jpg"

### follow ups ###
-  allow GAV input to find model by group:artifactid:version so we can uniquely identify a model and trigger deployment from central.
- more refactoring in this area. simplify parameters in methods like "handleRegisterModel"


ebamberg:
@frankfliu I was just rethinking the whole change. Maybe this change is complete non-sense. or at least the part with the input/output-type.

Guess I am a bit confused here.

I think the ServingTranslatorFactory is meant to be generic enough to choose a factory based on application-type and just uses Input.class/Output.class as inputs and outputs. 
But there is somekind of hardcoded "if (ImageClassification) then....

what will this mean for future trained model ? Every Model need  a change in the core the get served.

stu1130:
@ebamberg really appreciate your contribution. just to let you know we are releasing 0.10 so will probably merge your PR after that

frankfliu:
> @frankfliu I was just rethinking the whole change. Maybe this change is complete non-sense. or at least the part with the input/output-type.
> 
> Guess I am a bit confused here.
> 
> I think the ServingTranslatorFactory is meant to be generic enough to choose a factory based on application-type and just uses Input.class/Output.class as inputs and outputs.
> But there is somekind of hardcoded "if (ImageClassification) then....
> 
> what will this mean for future trained model ? Every Model need a change in the core to be servable.

@ebamberg 
I agree with you that this PR may not the best experience. Exposing too many configuration parameters in API actually reduce the usability and make user feel complicate.

I'm trying to hide the complexity as much as possible. I prefer the mode itself can provide those information and ServingTranslatorFactory be the bridge. Unfortunately there are gaps at API level, that's why we have to hard code the application for the time being. One big missing part is connecting Model Zoo Translator with TranslatorFactory, I don't have clear picture how to connect them yet.

In the short term, I think we only need fulfill three use cases:
1. Load model from an archive file (.zip/.tar/.tar.gz), assume this .zip file contains
   1. A properties file that tells ServingTranslatorFactory how to create a Translator
   2. Contains A TranslatorFactory/Translator implementation in the .zip file
2. Load model from a raw archive file, if we are not able to infer Translator, we just serving this model as Tensor in Tensor out
3. Load a model from ModelZoo/DJL Central

We already support 1,  2, I don't think we need expend the support to REST API level

You PR can address 3, but I'm thinking something different:
1. We can create a type of URL that can locate a mode: djl://ai.djl.mxnet/resnet/0.0.1/resnet18_v1, or djl://ai.djl.mxnet/resnet/0.0.1?layer=18&dataset=imagenet
2. in DJL Serving, we only need user provide this url to load a model from model zoo
3. the djl:// url support should be added at ModelZoo/Repository level not in DJL serving
4. currently DJL model zoo translator information was not in the metadata.json file, and the model depends on ModelZoo jar file. We need find a way to connect them.

ebamberg:
@frankfliu I close this PR request then for now, cause we agreed that there is a better solution.

PR #664, opened 2021-02-18
Make XavierInitializer default value & Improve setInitializer

stu1130:
## Description ##

### DIFF
--- a/api/src/main/java/ai/djl/training/DefaultTrainingConfig.java
+++ b/api/src/main/java/ai/djl/training/DefaultTrainingConfig.java
@@ -13,23 +13,23 @@
 package ai.djl.training;
 
 import ai.djl.Device;
+import ai.djl.nn.Parameter;
 import ai.djl.training.evaluator.Evaluator;
 import ai.djl.training.initializer.Initializer;
-import ai.djl.training.initializer.XavierInitializer;
-import ai.djl.training.initializer.XavierInitializer.FactorType;
-import ai.djl.training.initializer.XavierInitializer.RandomType;
 import ai.djl.training.listener.TrainingListener;
 import ai.djl.training.loss.Loss;
 import ai.djl.training.optimizer.Adam;
 import ai.djl.training.optimizer.Optimizer;
+import ai.djl.util.PairList;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.function.Predicate;
 
 /** {@code DefaultTrainingConfig} is an implementation of the {@link TrainingConfig} interface. */
 public class DefaultTrainingConfig implements TrainingConfig {
 
-    private Initializer initializer;
+    private PairList<Initializer, Predicate<Parameter>> initializers = new PairList<>();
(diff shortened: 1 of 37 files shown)


api/src/main/java/ai/djl/training/DefaultTrainingConfig.java
```diff
@@ -39,15 +39,12 @@
...
-     * XavierInitializer} as initialiser, {@link Adam} as optimiser, and the given {@link Loss}. The
-     * evaluators and listeners are left to the user's discretion.
+     * DefaultTrainingConfig} creates a default {@link TrainingConfig}, {@link Adam} as optimiser,
+     * and the given {@link Loss}. The evaluators and listeners are left to the user's discretion.
      *
      * @param loss the loss to use for training
      */
     public DefaultTrainingConfig(Loss loss) {
-        // Defaults to initializer defined in https://arxiv.org/abs/1502.01852
-        this.initializer = new XavierInitializer(RandomType.GAUSSIAN, FactorType.IN, 2);
```
zachgk, line 50: Can you keep this as the default initializer?

api/src/main/java/ai/djl/training/TrainingConfig.java
```diff
@@ -70,6 +72,13 @@
      */
     Initializer getInitializer();
 
+    /**
```
zachgk, line 75: If the initializer requires a Predicate, then initializing the model will require multiple initializers. This function is not defined for multiple initializers.

zachgk (changes requested):

api/src/main/java/ai/djl/training/DefaultTrainingConfig.java
stu1130: XavierInitializer is already a default initializer. You can check Parameter Type weight

api/src/main/java/ai/djl/training/TrainingConfig.java
stu1130: good point, I will extend to a list

api/src/main/java/ai/djl/training/DefaultTrainingConfig.java
stu1130: I copied the parameter of XavierInitializer from TensorFlow. Let me know if it is really optimal

zachgk: I am not sure the best way to present this. Right now, there are two papers on Xavier initializer (that I know of at least). The default parameters in `new XavierInitializer()` are the first paper because it also named the initializer. The `new XavierInitializer(RandomType.GAUSSIAN, FactorType.IN, 2)` are from the second which improved upon it. We probably should make that a constant in XavierInitializer or something, but I think we should keep our default initializer to the second paper rather than reverting it to the first paper.

stu1130: done!

zachgk (approved):

PR #665, opened 2021-02-18
Run multiple engines integration test.

frankfliu:
## Description ##


.github/workflows/continuous.yml
```diff
@@ -145,6 +145,7 @@ jobs:
       - name: Compile PaddlePaddle JNI
         run: ./gradlew :paddlepaddle:paddlepaddle-native:compileJNI
       - name: Integration tests
+        # MXNet integration test on Github Action fail due to memory limitation on Windows
         run: |
-          ./gradlew :paddlepaddle:paddlepaddle-engine:test :paddlepaddle:paddlepaddle-model-zoo:test
```
lanking520, line 149: This will disable all paddle and onnx test coverage.

.github/workflows/continuous.yml
```diff
@@ -145,6 +145,7 @@ jobs:
...
-          ./gradlew :onnxruntime:onnxruntime-engine:test
+          ./gradlew --rerun-tasks -Dai.djl.default_engine=PyTorch build
```
lanking520, line 150: Do rerun on all tasks will make the test time tripled. I would still suggest we do test in parallel and for the sake of clearance.

lanking520 (changes requested):

.github/workflows/continuous.yml
frankfliu: 1. We are not collecting test coverage here, this run doesn't runt jRR
2. gradle build will run all tests except MXNet integration test

.github/workflows/continuous.yml
frankfliu: 1. The build itself is very fast, less than 1 minute, especially for TF and PT integration test
2. The majority time spend is on spin up the machine, download engine native binary and dependencies. 
3. Windows build is run parallel with Linux and mac, it should not increase total continuous build time

roywei (approved):

lanking520 (approved):

PR #669, opened 2021-02-18
DJL parameter saving

lanking520:
## Description ##

This PR will cover several major components to allow DJL to load/save parameter smoothly. Now you can do InputStream/OutputStream to Pytorch model.

Stages
- Stage 1: Create saving functions for MXNet and PyTorch [Done]
- Stage 2: Make model saving and loading all in one file [In next PR]

Here is the plan:

Assume you have a block architecture like this:

```
SequentialBlock
SymbolBlock
SequentialBlock
```

What we will do to save is adding SymbolBlocks original Artifacts (e.g MXNet symbol params, PyTorch .pt TorchScript) into the `.params` file we generate. It will make the whole thing like a zip.

So, in the end, we will expect to have `block.params` standalone file.

To load all of them back. Instead of the current detour way to load Symbolic model first and replace all the content later, we can now make the block all calling `loadParameters` functions to load the model. What internally it will do, is to separate these content into files (symbol, params in MXNet, torchscript in PyTorch) and load them during the `loadParameters` call from `SymbolBlock`.

This will ensure our loading/saving experience very smooth and we don't duplicate our parameters.

### DIFF
--- a/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/JniUtils.java
+++ b/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/JniUtils.java
@@ -46,7 +46,7 @@ public final class JniUtils {
 
     private static final int NULL_PTR = 0;
 
-    private static final byte[] BUFFER = new byte[4194304]; // preserved buffer
+    private static final int BYTE_LENGTH = 4194304;
 
     private JniUtils() {}
 
@@ -1379,6 +1379,7 @@ public final class JniUtils {
     }
 
     public static PtSymbolBlock loadModule(PtNDManager manager, InputStream is, Device device) {
+        byte[] buf = new byte[BYTE_LENGTH];
         long handle =
                 PyTorchLibrary.LIB.moduleLoad(
                         is,
@@ -1386,12 +1387,13 @@ public final class JniUtils {
                             PtDeviceType.toDeviceType(device),
                             device.equals(Device.cpu()) ? -1 : device.getDeviceId()
                         },
-                        BUFFER);
+                        buf);
         return new PtSymbolBlock(manager, handle);
     }
 
     public static void writeModule(PtSymbolBlock block, OutputStream os) {
-        PyTorchLibrary.LIB.moduleWrite(block.getHandle(), os, BUFFER);
+        byte[] buf = new byte[BYTE_LENGTH];
+        PyTorchLibrary.LIB.moduleWrite(block.getHandle(), os, buf);
     }
 
     public static NDList moduleGetParams(PtSymbolBlock block, PtNDManager manager) {
--- a/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/JniUtils.java
+++ b/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/JniUtils.java
@@ -22,6 +22,8 @@ import ai.djl.pytorch.engine.PtDeviceType;
 import ai.djl.pytorch.engine.PtNDArray;
 import ai.djl.pytorch.engine.PtNDManager;
 import ai.djl.pytorch.engine.PtSymbolBlock;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.file.Path;
@@ -43,6 +45,8 @@ public final class JniUtils {
     private static Set<String> configs;
 
     private static final int NULL_PTR = 0;
+
+    private static final byte[] BUFFER = new byte[4194304]; // preserved buffer
 
     private JniUtils() {}
 
@@ -1372,6 +1376,34 @@ public final class JniUtils {
                         extraFileKeys,
                         extraFileValues);
         return new PtSymbolBlock(manager, handle);
+    }
+
+    public static PtSymbolBlock loadModule(PtNDManager manager, InputStream is, Device device) {
+        long handle =
+                PyTorchLibrary.LIB.moduleLoad(
+                        is,
+                        new int[] {
+                            PtDeviceType.toDeviceType(device),
+                            device.equals(Device.cpu()) ? -1 : device.getDeviceId()
+                        },
+                        BUFFER);
+        return new PtSymbolBlock(manager, handle);
+    }
+
+    public static void writeModule(PtSymbolBlock block, OutputStream os) {
+        PyTorchLibrary.LIB.moduleWrite(block.getHandle(), os, BUFFER);
+    }
+
+    public static NDList moduleGetParams(PtSymbolBlock block, PtNDManager manager) {
+        long[] handles = PyTorchLibrary.LIB.moduleGetParams(block.getHandle());
+        String[] names = PyTorchLibrary.LIB.moduleGetParamNames(block.getHandle());
+        NDList list = new NDList(handles.length);
+        for (int i = 0; i < handles.length; i++) {
+            PtNDArray array = new PtNDArray(manager, handles[i]);
+            array.setName(names[i]);
+            list.add(array);
+        }
+        return list;
     }
 
     public static void enableInferenceMode(PtSymbolBlock block) {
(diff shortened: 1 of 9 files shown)


pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_inference.cc
```diff
@@ -46,6 +46,75 @@ JNIEXPORT jlong JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_moduleLoad(
...
+}
+
+JNIEXPORT void JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_moduleWriteToOS(
+        JNIEnv* env, jobject jthis, jlong module_handle, jobject jos) {
+    API_BEGIN()
+    auto* module_ptr = reinterpret_cast<torch::jit::script::Module*>(module_handle);
+    std::ostringstream stream;
+    module_ptr->save(stream);
+    auto str = stream.str();
+    jclass os_class = env->GetObjectClass(jos);
```
frankfliu, line 90: I think we can move class and method lookup into ai_djl_pytorch_jni_cache.h file

pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_inference.cc
```diff
@@ -99,3 +168,40 @@ JNIEXPORT void JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_torchDeleteModule(
...
+    module_ptr->save(djl::utils::jni::GetStringFromJString(env, jpath));
+    API_END()
+}
+
+JNIEXPORT jlongArray JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_moduleGetParams(
+        JNIEnv* env, jobject jthis, jlong jhandle) {
+    API_BEGIN()
+    auto* module_ptr = reinterpret_cast<torch::jit::script::Module*>(jhandle);
+    std::vector<jlong> jptrs;
+    for (const auto tensor : module_ptr->parameters())
```
frankfliu, line 185: format c code

pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/JniUtils.java
```diff
@@ -1374,6 +1376,36 @@ public static PtSymbolBlock loadModule(
         return new PtSymbolBlock(manager, handle);
     }
 
+    public static PtSymbolBlock loadModule(PtNDManager manager, InputStream is, Device device) {
+        byte[] buffer = new byte[4194304];
```
stu1130, line 1380: can we have constants declare somewhere

pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_inference.cc
```diff
@@ -46,6 +46,71 @@ JNIEXPORT jlong JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_moduleLoad(
...
+        return -1;
+    }
+    jmethodID method_id = env->GetMethodID(is_class, "read", "([BII)I");
+    if (method_id == nullptr) {
+        env->ThrowNew(ENGINE_EXCEPTION_CLASS, "The read method in InputStream is not found");
+        return -1;
+    }
+    std::ostringstream os;
+    int len = env->GetArrayLength(arr);
+    char* temp = new char[len];
```
stu1130, line 64: If it is temporary memory, just put it in stack 
```suggestion
    char temp[len];
```

pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_inference.cc
stu1130: tensor is considered as big object, will reference here
```suggestion
    for (const auto& tensor : module_ptr->parameters())
```

pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_inference.cc
```diff
@@ -99,3 +164,40 @@ JNIEXPORT void JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_torchDeleteModule(
...
+    return jarray;
+    API_END_RETURN()
+}
+
+JNIEXPORT jobjectArray JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_moduleGetParamNames(
+        JNIEnv* env, jobject jthis, jlong jhandle) {
+    API_BEGIN()
+    auto* module_ptr = reinterpret_cast<torch::jit::script::Module*>(jhandle);
+    std::vector<std::string> jptrs;
+    for (const auto named_tensor : module_ptr->named_parameters())
```
stu1130, line 197: same as above

pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/JniUtils.java
```diff
@@ -44,6 +46,8 @@
 
     private static final int NULL_PTR = 0;
 
+    private static final byte[] BUFFER = new byte[4194304]; // preserved buffer
```
frankfliu, line 49: this should not be static, the static buffer might be used in multiple threads.

pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_inference.cc
```diff
@@ -46,6 +46,71 @@ JNIEXPORT jlong JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_moduleLoad(
...
+        return -1;
+    }
+    std::ostringstream os;
+    int len = env->GetArrayLength(arr);
+    char* temp = new char[len];
+    int available = 0;
+    while(available != -1) {
+        available = env->CallIntMethod(jis, method_id, arr, 0, len);
+        if (available != -1) {
+            env->GetByteArrayRegion(arr, 0, available, reinterpret_cast<jbyte*>(temp));
```
stu1130, line 69: you can remove the temp char array and directly get data from arr

lanking520: No, the os.write requires char* inputs. There is no holder to take that piece of memory. I tried to create the pointer directly without allocating memory and it is just causing Seg fault.

stu1130: ok before we can come up with better solution, let's leave it as it is. add a delete[] arr after you done using arr

lanking520: arr is passed from the Java, you can only delete the local reference

pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_inference.cc
```diff
@@ -46,6 +47,71 @@ JNIEXPORT jlong JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_moduleLoad(
...
+    return -1;
+  }
+  jmethodID method_id = env->GetMethodID(is_class, "read", "([BII)I");
+  if (method_id == nullptr) {
+    env->ThrowNew(ENGINE_EXCEPTION_CLASS, "The read method in InputStream is not found");
+    return -1;
+  }
+  std::ostringstream os;
+  int len = env->GetArrayLength(arr);
+  auto temp = std::make_unique<char[]>(len);
```
frankfliu, line 65: boolean isCopy;
jbyte* b = GetByteArrayElements(env, arr, &isCopy);
char* ch = (char*)b;
...
ReleaseByteArrayElements(env, arr, b, JNI_ABORT);

frankfliu (approved):

stu1130 (approved):

PR #672, opened 2021-02-18
Adding evaluation for test dataset in EasyTrain

aksshi2830:
Added a utility evaluateDataset method so that users can run evaluation of their datasets seperately.

### DIFF
--- a/integration/src/main/java/ai/djl/integration/tests/training/EvaluateDatasetTest.java
+++ b/integration/src/main/java/ai/djl/integration/tests/training/EvaluateDatasetTest.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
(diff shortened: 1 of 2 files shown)


goswamig (commented):
May be I am missing something but current implementation can let users run evaluation of their datasets separately.

What am I missing here?

roywei (commented):
Can you add a test for `evaluateDataset`?

aksshi2830:
> Can you add a test for `evaluateDataset`?

Done

integration/src/main/java/ai/djl/integration/tests/training/EvaluateDatasetTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+        Trainer trainer = model.newTrainer(config);
+        trainer.initialize(new Shape(1, 784));
+
+        Metrics metrics = new Metrics();
+        trainer.setMetrics(metrics);
+
+        EasyTrain.evaluateDataset(trainer, testMnistDataset);
+
+        trainer.notifyListeners(listener -> listener.onEpoch(trainer));
```
roywei, line 75: It maybe a little unintuitive to call notify listeners and `onEpoch` to view the result because the training has finished.

roywei (approved):

roywei:
> May be I am missing something but current implementation can let users run evaluation of their datasets separately.
> 
> What am I missing here?

To clarify, users can do evaluation separately and define custom training/eval loop imperative style like PyTorch using normal Trainer API. DJL's `EasyTrain` API is a higher level abstraction that helps basic training use cases. We aim to provide a few helper functions to avoid writing training/eval loop manually. Kind of like Keras's `fit`, `evaluate`: https://www.tensorflow.org/api_docs/python/tf/keras/Model

lanking520 (approved):

PR #679, opened 2021-02-22
add codecov

lanking520:
## Description ##

Add Code Cov for DJL

### DIFF
--- a/.github/workflows/continuous.yml
+++ b/.github/workflows/continuous.yml
@@ -120,6 +120,14 @@ jobs:
         with:
           name: tensorflow-model-zoo
           path: tensorflow/tensorflow-model-zoo/build/reports
+      - name: upload to codecov
+        uses: codecov/codecov-action@v1
+        with:
+          files: ./build/reports/jacoco/jacocoRootReport/jacocoRootReport.xml
+          name: codecov-umbrella
+          fail_ci_if_error: true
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous.yml
```diff
@@ -120,6 +120,16 @@ jobs:
         with:
           name: tensorflow-model-zoo
           path: tensorflow/tensorflow-model-zoo/build/reports
+      - name: upload to codecov
+        if: always()
+        run: mkdir coverage && cp -t coverage/ $(find . -name "*jacocoTestReport.xml")
+      - uses: codecov/codecov-action@v1
+        with:
+          directory: ./coverage/
```
zachgk, line 128: With gradle task `jacocoRootReport`, it creates a joined jacocoTestReport. You should only need to use that one

.github/workflows/continuous.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       - name: Compile PaddlePaddle JNI
         run: ./gradlew :paddlepaddle:paddlepaddle-native:compileJNI
       - name: Build with Gradle
-        run: ./gradlew build jRR jRV
+        run: ./gradlew build jRR jRV jacocoRootReport
```
zachgk, line 56: I think the jRR is jacocoRootReport

zachgk (approved):

PR #683, opened 2021-02-23
Create ObjectDetectionDataset

zachgk:
This dataset simplifies the creation of new object detection datasets by
presenting a clearer and more standard class to extend.

This PR first creates an ImageDataset class that is the parent class of both the
ImageClassificationDataset and this ObjectDetectionDataset.

The ObjectDetectionDataset is then used as the new parent for the BananaDataset,
PikachuDataset, and CocoDetection dataset. They are all modified to use it.

In addition, some additional documentation is added to Rectangle to document
that it usually has coordinates from 0 to 1.

### DIFF
--- a/basicdataset/src/main/java/ai/djl/basicdataset/cv/ImageDataset.java
+++ b/basicdataset/src/main/java/ai/djl/basicdataset/cv/ImageDataset.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package ai.djl.basicdataset.cv;
+
+import ai.djl.basicdataset.cv.classification.ImageClassificationDataset;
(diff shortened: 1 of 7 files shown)


basicdataset/src/main/java/ai/djl/basicdataset/cv/ImageDataset.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * Returns the image at the given index in the dataset.
+     *
+     * @param index the index (if the dataset is a list of data items)
+     * @return the image
+     * @throws IOException if the image could not be loaded
+     */
+    protected abstract Image getImage(long index) throws IOException;
```
lanking520, line 61: Do we also need a getLabel function to get label for the image?

basicdataset/src/main/java/ai/djl/basicdataset/cv/ObjectDetectionDataset.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Returns the list of objects in the image at the given index.
+     *
+     * @param index the index (if the dataset is a list of data items)
+     * @return the list of objects in the image. The long is the class number of the index into the
+     *     list of classes of the desired class name. The rectangle is the location of the object
+     *     inside the image.
+     * @throws IOException if the data could not be loaded
+     */
+    public abstract PairList<Long, Rectangle> getObjects(long index) throws IOException;
```
lanking520, line 69: Can `DetectedObject` class helps here?

zachgk: Maybe, they are slightly different due to the probabilities. Or more formally, what we have here is the correct answer whereas Classifications/DetectedObjects are probability distributions over all answers. I figured not having the extraneous information to confuse users might be cleaner, but we could think about it in the future.

lanking520 (approved):

PR #685, opened 2021-02-23
fix nightly test

roywei:
windows cpu build is still having OOM issue even on CPU if we run test with 3 engines together.
We have PyTorch and TensorFlow integration covered in contineous GitHub Action.


integration/src/test/java/ai/djl/integration/IntegrationTests.java
```diff
@@ -29,7 +29,9 @@ public void runIntegrationTests() {
         String[] engines;
         String defaultEngine = System.getProperty("ai.djl.default_engine");
         if (defaultEngine == null) {
-            engines = new String[] {"MXNet", "PyTorch", "TensorFlow"};
```
zachgk, line 32: Can you add another condition so it runs the three engines when not on Windows?

stu1130 (approved):

zachgk (approved):

PR #691, opened 2021-02-24
Delete DataManager

zachgk:
Also done:
- Create a forwardInternal with labels and the implementations in the higher level
blocks SequentialBlock and ParallelBlock
- Removed duplicate testSeqToSeq test


stu1130 (approved):

lanking520:
LGTM when I see the title. Look into the code now

examples/src/main/java/ai/djl/examples/training/TrainSentimentAnalysis.java
```diff
@@ -143,8 +141,16 @@ public static TrainingResult runExample(String[] args)
         }
     }
 
-    private static Block getModel() {
+    private static Block getModel(ModelZooTextEmbedding embedding) {
         return new SequentialBlock()
+                .addSingleton(
+                        a -> {
+                            try {
+                                return embedding.embedText(a);
```
lanking520, line 149: The problem to put it here is: I would like to freeze this block and not use for training. If remove DataManager, I will not be able to freeze the block

lanking520 (changes requested):

examples/src/main/java/ai/djl/examples/training/TrainSentimentAnalysis.java
zachgk: This would be frozen because the ModelZooTextEmbedding freezes (it uses a separate predictor). In a different case, you should be able to just call embedding.freeze() to freeze just the embedding.

lanking520: Can you verify if PyTorch engine will work in this case? I am not sure if the training work are correctly handled

zachgk: There is no word embedding for pt in the model zoo so I can't test it. But, is there a reason that PT may not work?

lanking520: It should work by default. Just concerned if the NDArray inside will be impacted. In the DistiledBert training for Amazon Dataset, we throw the embedding layer into a lambda function to avoid parameter collections

lanking520 (approved):

PR #697, opened 2021-02-26
DJL BlockFactory

lanking520:
## Description ##

The block factory is a class that provides easy access to recover DJL based model for inference with DJLServing. It contains a single function `getBlock()`  and essentially just recover the block.

Apart from that, there is a new `SymbolBlock.newInstance()` method that only create a empty SymbolBlock holder. It can now be used in PyTorch and MXNet to recover the Symbolic model information from combined params file. In this case, we no longer needs to load the SymbolBlock from the model, instead, everything is loaded from 1 file.

### DIFF
--- a/api/src/main/java/ai/djl/nn/BlockFactory.java
+++ b/api/src/main/java/ai/djl/nn/BlockFactory.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package ai.djl.nn;
+
+import ai.djl.ndarray.NDManager;
+import ai.djl.repository.zoo.ModelZoo;
+import java.io.Serializable;
+
+/**
+ * Block factory is a component to make standard for block creating and saving procedure. Block
+ * factory design is intended to bypass the serialization of the blocks. This class can be used by
+ * {@link ModelZoo} or DJL Serving to recover the block to its uninitialized states. User should
+ * combine this method with the block.loadParameter to get the block with all parameters.
+ */
+public interface BlockFactory extends Serializable {
+
+    /**
+     * Constructs the uninitialized block.
+     *
+     * @param manager the manager to assign to block
+     * @return the uninitialized block
+     */
+    Block newBlock(NDManager manager);
+}
--- a/api/src/main/java/ai/djl/nn/SymbolBlock.java
+++ b/api/src/main/java/ai/djl/nn/SymbolBlock.java
@@ -12,6 +12,7 @@
  */
 package ai.djl.nn;
 
+import ai.djl.ndarray.NDManager;
 import ai.djl.ndarray.types.Shape;
 import ai.djl.util.PairList;
 
@@ -20,6 +21,16 @@ import ai.djl.util.PairList;
  * the engine in its native format.
  */
 public interface SymbolBlock extends Block {
+
+    /**
+     * Creates an empty SymbolBlock instance.
+     *
+     * @param manager the manager to be applied in the SymbolBlock
+     * @return a new Model instance
+     */
+    static SymbolBlock newInstance(NDManager manager) {
+        return manager.getEngine().newSymbolBlock(manager);
+    }
 
     /** Removes the last block in the symbolic graph. */
     default void removeLastBlock() {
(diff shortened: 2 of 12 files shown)


zachgk:
If the BlockFactory is for DJL serving, does it integrate with the existing ModelZoo or ModelLoader concepts?

integration/src/main/java/ai/djl/integration/tests/nn/BlockFactoryTest.java
```diff
@@ -0,0 +1,114 @@
...
+            Criteria.Builder<Image, Classifications> builder =
+                    Criteria.builder()
+                            .optApplication(Application.CV.IMAGE_CLASSIFICATION)
+                            .setTypes(Image.class, Classifications.class)
+                            .optProgress(new ProgressBar())
+                            .optArtifactId("resnet")
+                            .optEngine(name)
+                            .optGroupId("ai.djl." + name.toLowerCase())
+                            .optFilter("layers", "50");
+            Model model = ModelZoo.loadModel(builder.build());
```
goswamig, line 103: In this case it loads the model with just name, since you don't specify any url/file?

integration/src/main/java/ai/djl/integration/tests/nn/BlockFactoryTest.java
```diff
@@ -0,0 +1,114 @@
...
+        @Override
+        public Block getBlock(NDManager manager) {
+            SequentialBlock newBlock = new SequentialBlock();
+            newBlock.add(SymbolBlock.newInstance(manager));
+            newBlock.add(Blocks.batchFlattenBlock());
+            newBlock.add(Linear.builder().setUnits(10).build());
+            return newBlock;
+        }
+
+        public Model constructBlock()
```
goswamig, line 91: I felt function name is a bit misnomer since it loads the model and modifies the last layer. 
Something like "ChangeLastBlockInModel" ?

goswamig (approved):
Overall LGTM. I left a minor comment.

lanking520:
> If the BlockFactory is for DJL serving, does it integrate with the existing ModelZoo or ModelLoader concepts?

The blockfactory will be both used in ModelZoo and DJLServing. It can be the way for ModelLoader to load customized DJL trained model

integration/src/main/java/ai/djl/integration/tests/nn/BlockFactoryTest.java
lanking520: No, it does looking for the model stored in DJL model zoo.

zachgk:
> > If the BlockFactory is for DJL serving, does it integrate with the existing ModelZoo or ModelLoader concepts?
> 
> The blockfactory will be both used in ModelZoo and DJLServing. It can be the way for ModelLoader to load customized DJL trained model

Is the use of the BlockFactory mandatory or optional? It seems like it is a possible way for implementing the modelLoader, but doesn't really ease the process. It just adds an additional class and moves the work around.

I am trying to understand how the BlockFactory will be used and how it integrates with the rest of the DJL API:

- Are there any methods or classes that will accept a BlockFactory as an input?
- Will we update classes that are similar to a BlockFactory in that they produce Blocks such as ResNetV1?

lanking520:
> * Are there any methods or classes that will accept a BlockFactory as an input?
> * Will we update classes that are similar to a BlockFactory in that they produce Blocks such as ResNetV1?

@zachgk Here is my thoughts: This will only be mandatory for any DJL based model. If any model contains DJL imperative blocks should implement the `BlockFactory` to help ModelZoo reconstruct it. It may work in a similar fashion with the ModelLoader. 

`ModelFactory` will be used during the model loading time. The ModelZoo will try to find a `BlockFactory` implementation from the provided artifacts to reconstruct the block.

Ideally, All DJL block that can be used directly should implement the BlockFactory.

Here is an example of the case when we use it: Assume you have a symbolic model (e.g MXNet Symbol, TorchScript) and try to use it along with Imperative part (e.g fullyConnected). The model loading part is quite complicated: You have to Load the Symbolic part and then the imperative parts.

```
Imperative Block
SymbolBlock
Imperative Block
```
If we implement the block factory, it works like the `symbol.json` in MXNet to describe the model structure, but just written in code.

The imperative model stored in the model zoo will be like the followings:

- ZooBlockFactory.java: The BlockFactory for the Imperative Model
- parameters.param: The parameters for the model

These two files will be sufficient for us to serve the model in anywhere. Since the serialization steps is complete, all model information will be stored into parameters.param.

api/src/main/java/ai/djl/nn/BlockFactory.java
```diff
@@ -0,0 +1,27 @@
...
+/** Block factory is a component to make standard for block creating and saving procedure. */
+public interface BlockFactory {
+
+    /**
+     * Constructs the uninitialized block.
+     *
+     * @param manager the manager to assign to block
+     * @return the uninitialized block
+     */
+    Block getBlock(NDManager manager);
```
frankfliu, line 26: newBlock()

api/src/main/java/ai/djl/nn/SymbolBlock.java
```diff
@@ -21,6 +23,16 @@
...
 public interface SymbolBlock extends Block {
 
+    /**
+     * Creates an empty SymbolBlock instance.
+     *
+     * @param manager the manager to be applied in the SymbolBlock
+     * @return a new Model instance
+     */
+    static SymbolBlock newInstance(NDManager manager) {
+        return Engine.getInstance().newSymbolBlock(manager);
```
frankfliu, line 33: manager.getEngine().newSymbolBlock(manager);

lanking520:
Make factory serializable

stu1130:
maybe add an example use case and reason on the BlockFactory javadoc

frankfliu (approved):

zachgk (approved):

stu1130 (approved):

PR #704, opened 2021-02-27
Non zero operator for PyTorch

aksrajvanshi:
Added nonzero operator for PyTorch. Added a Scalar check in JniUtils as PyTorch behaves differently in comparison to MXNet when passed a scalar value for nonzero operator.


goswamig (commented):
LGTM, can we add unit test ?

frankfliu (approved):

frankfliu:
> LGTM, can we add unit test ?

We should already have coverage for this test case. We are testing all operators in Integration tests, If the engine throw UnsupportedOperationException, we just skip the test result.

PR #717, opened 2021-03-02
Use detach c api instead of stop_gradient

stu1130:
## Description ##
detach & stop_gradient are doing the same thing (correct me if I am wrong). The former one is for dynamic graph while the later is for static graph

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/NDArray.java
+++ b/api/src/main/java/ai/djl/ndarray/NDArray.java
@@ -162,16 +162,18 @@ public interface NDArray extends NDResource {
      * ai.djl.training.GradientCollector#backward(NDArray)} can compute the gradient with respect to
      * it.
      */
-    void attachGradient();
+    default void requiresGradient() {
+        requiresGradient(true);
+    }
 
     /**
-     * Attaches a gradient {@code NDArray} with the given {@link SparseFormat} to this {@code
-     * NDArray} and marks it so {@link ai.djl.training.GradientCollector#backward(NDArray)} can
-     * compute the gradient with respect to it.
+     * Attaches a gradient {@code NDArray} to this {@code NDArray} and marks it so {@link
+     * ai.djl.training.GradientCollector#backward(NDArray)} can compute the gradient with respect to
+     * it.
      *
-     * @param sparseFormat the {@link SparseFormat} for the gradient
+     * @param requiresGrad if {@code NDArray} requires gradient or not
      */
-    void attachGradient(SparseFormat sparseFormat);
+    void requiresGradient(boolean requiresGrad);
 
     /**
      * Returns the gradient {@code NDArray} attached to this {@code NDArray}.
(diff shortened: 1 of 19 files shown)


stu1130:
also should we use another set of name for this kind of gradient manipulation? attachGradient & detachGradient seems confuses the user with NDManager attach & detach

stu1130:
the other way is to change NDManager attach & detach. Since most people are coming from Python side, they will be used to using detach and they will see it’s available to use but it will do a completely different thing by Marcos

zachgk:
detach and stopGradient should be different things. If attachGradient says that you should track gradients for an NDArray, detachGradient says you shouldn't track it. So, they mutate the NDArray.

The stopGradient is to make a copy of the NDArray. But, copies usually will still propagate gradients back to the original while stopGradient does not. It can then be used for magnifying gradients or other gradient manipulations when used instead of or in conjunction with the original gradient propagating array.

stu1130:
In PyTorch, requires_grad=False have the limitation on leaf variables where you have to use detach(). https://discuss.pytorch.org/t/understanding-of-requires-grad-false/39765 so It looks like they are the same?

zachgk:
Yeah, so the relationship seems to be:

| DJL | PyTorch | MXNet |
| ---- | -----| ---- |
| attachGradient | require_grad(True)| attach_grad('write')|
| detachGradient | require_grad(False)|attach_grad(null)|
| stopGradient| detach |stopGrad and maybe also ndarray.detach|

stu1130:
> Yeah, so the relationship seems to be:
> 
> DJL	PyTorch	MXNet
> attachGradient	require_grad(True)	attach_grad('write')
> detachGradient	require_grad(False)	attach_grad(null)
> stopGradient	detach	stopGrad and maybe also ndarray.detach

If that is the case, I propose we should remove the case 2 and keep only attachGradient & detachGradient. Does that sound good?

zachgk:
Case 2 is the same as freezing parameters so I think we still want it. I like requiresGrad(Boolean) for case 1 and 2. It expresses the user goal (this NDArray requires the gradients so that it can be optimized).

I am not as much a fan of detach because it "detaches" from the computation graph. But, that is a weird comment and it is hard to understand what the implications of it are. Stop gradient which stops gradients from propagating is a bit more straightforward. It is also closer to the similar operation of scaleGradient which magnifies particular gradients more than back-propagation normally would.

stu1130:
> Case 2 is the same as freezing parameters so I think we still want it. I like requiresGrad(Boolean) for case 1 and 2. It expresses the user goal (this NDArray requires the gradients so that it can be optimized).
> 
> I am not as much a fan of detach because it "detaches" from the computation graph. But, that is a weird comment and it is hard to understand what the implications of it are. Stop gradient which stops gradients from propagating is a bit more straightforward. It is also closer to the similar operation of scaleGradient which magnifies particular gradients more than back-propagation normally would.

DJL Parameter already have an attribute called requiresGrad to cover freezing params so we can still get rid of case 2. But in general, I am also fine with your proposal. @frankfliu @lanking520 @goswamig @aksrajvanshi what do you guys think?

stu1130:
I have updated the PR to add requiresGrad(boolean)

zachgk (approved):

Issue #633, opened 2021-02-09
Model Downlading


anfee1:
## Description

This feature should allow a user to download a model using the ModelView UI or a command line terminal. This will give the user a zip package containing the json file, the version of the model the users looking for, the Readme file and all other misc files related to the model.

This will include a download button on the Model View and any other interface implementations that would be needed for this feature.

PR #719, opened 2021-03-02
Model Dowload Buttons

anfee1:
## Description ##
This gives the functionality of providing download URIs that are taken from the model, converts them to URLs, and then feeds it back to the model view in the form of a clickable button.

Interesting edge cases to note here:
A way needs to be found to clear the model links on the front end so when you switch from model to model the unused links will disappear. More will be explained in an issue that I will link. 

**Special Thanks** to @ebamberg for the help/insight. #694 

Fixes:#633


central/src/main/java/ai/djl/serving/central/handler/ModelDownloadHandler.java
```diff
@@ -0,0 +1,118 @@
...
+                        } else {
+                            throw new BadRequestException("modelName is mandatory.");
+                        }
+
+                    } catch (IOException | ModelNotFoundException ex) {
+                        throw new IllegalArgumentException(ex.getMessage(), ex);
+                    }
+                })
+                .exceptionally((ex) -> Collections.emptyMap())
+                .thenAccept(linksMap -> jsonResponse.sendAsJson(ctx, request, linksMap));
```
ebamberg, line 105: Does this means every web-client=browser need to have access to the url through firewall,proxy and eventuelly to the filesystem where central stores uploads ?
Cause this just returns a link and not downloads the files.

I just thinking about future improvements and just want top drop a few devOp thoughts...

1) every client need access to the url, in bigger enterprises with restricted network-policies it would be beneficially to grant access for the central server through all firewalls/proxies needed instead for every client. Same for serving instances running on cloud-nodes and need access
2) Authentication , if you plan to add role-based authentication and authorization in future it would be easier to allow central to check privileges and manages the download.  
3) upload artifacts using central, which is planned as I understood. when download runs through central there would be no need to grant access to file-system and central web-UI
4) using of central as Mirror in front of real-repositories. 
5) using central for caching of remote-model-repositories

just a few ideas, might be wrong if you want to scale out based on network-filesystem instead of scaling out central.

The ideas are just from a DevOp perspective

central/src/main/java/ai/djl/serving/central/handler/ModelDownloadHandler.java
```diff
@@ -0,0 +1,118 @@
...
+                            return ModelLink.linkFinder(modelName);
+                        } else {
+                            throw new BadRequestException("modelName is mandatory.");
+                        }
+
+                    } catch (IOException | ModelNotFoundException ex) {
+                        throw new IllegalArgumentException(ex.getMessage(), ex);
+                    }
+                })
+                .exceptionally((ex) -> Collections.emptyMap())
```
ebamberg, line 104: no logging ?

central/src/main/webapp/components/ModelView.jsx
```diff
@@ -250,6 +252,9 @@ export default function ModelView(props) {
 							: <DynForm data={noSynset}/>
 						}
 					</TabPanel>
+					<TabPanel value={index} index={6} className={classes.tabpanel}>
+                        <ModelDownloadButtons modelName={props.model.name}/>
```
ebamberg, line 256: Maybe it would be better to add a Menu with actions which are shown all the time instead of adding a tab with download action.
If no more interactions for models are planned this is fine. but I am just wondering in case more actions to be included in future, if we want to have 1 tab per action. 
ffrom a usability perspective it is usally a better approach to show all possible actions on a central place. 

As I said just a thought in case that more interactions with the model are planned.

anfee1: I like this idea actually. I wanted to change what the UI looked like, but thought it would be messy to add a change like that to what should be a simple PR. Also, I didn't want to delay a PR while the actual functionality hasn't been checked.

anfee1: I also am not completely sure of the end goal for this project, so it could be very much the plan to add more actionable items and if that's the case I think that you surely had a better solution with the action button.

central/src/main/java/ai/djl/serving/central/handler/ModelDownloadHandler.java
anfee1: Thank you I must have missed this.

central/src/main/webapp/components/ModelView.jsx
ebamberg: speaking of future plans for this sub-project, I was also thinking if it might be better to do the download through the central server instead of just sending the link to the web-client.

At the moment every web-client/browser needs access through firewalls/proxies to download the model from the original URL as well as to central.
Instead of  that it would be possible that just the central-server is configured to access all necessary network-resources.
The client could just call an URL of the central-server and the server would stream the model-download to the client 
 I hope you know what I mean :-) ?

This would open the door to some future improvements like:
1) Authentication/Authorisation to download managed by central
2) central could be used as a Mirror or for Caching of remote models - similar to how Nexus can be configured when using maven. Nexus is used often this way for caching or to control which artifacts are used inside of a company.
3) uploading models with central without granting access for all web-clients to whatever storage central is using. so different storage strategies would be possible.
4)  single point to configure firewall access. Would also be nice for cloud based services using models.

just a few ideas from a DevOp perspective 

But as I said depends on future plans for this project and can easily be added later on.

zachgk: I am having trouble thinking of another action that would go in a separate button. Maybe we make a more generic "Get Model" tab. Then, it could include the download button and some other actions about getting the model like a sample criteria snippet.

@ebamberg I think what you are saying about proxying the download through the model server makes sense (at least as an option). We can probably put that in a later PR, though. Do you want to open an issue about it?

central/src/main/webapp/components/DownloadButtons.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+const useFetch = (modelName) => {
+	const [data, setData] = useState([]);
+
+	useEffect(() => {
+		async function fetchData() {
+			axios.get("http://"+window.location.host+"/serving/models?modelName="+modelName)
+				.then(function(response) {
+					let appdata = Object.keys(response.data).map(function(key) {
+						console.log(key)
```
zachgk, line 17: Try not to leave in debugging console.log statements when you submit the PR.

central/src/main/webapp/components/DownloadButtons.jsx
```diff
@@ -0,0 +1,53 @@
...
+							link: response.data[key]
+						};
+					});
+					setData(appdata);
+					console.log(appdata)
+				})
+				.catch(function(error) {
+					console.log(error);
+				})
+				.then(function() {
```
zachgk, line 29: This empty .then(...) call doesn't do anything and we can safely delete it

zachgk (changes requested):

central/src/main/java/ai/djl/serving/central/classes/ModelLink.java
```diff
@@ -0,0 +1,74 @@
...
+                                            artifact.getFiles().entrySet()) {
+                                        URI fileUri = URI.create(entry.getValue().getUri());
+                                        URI baseUri = artifact.getMetadata().getRepositoryUri();
+                                        if (!fileUri.isAbsolute()) {
+                                            fileUri = base.resolve(baseUri).resolve(fileUri);
+                                        }
+                                        try {
+                                            links.put(entry.getKey(), fileUri);
+                                        } catch (Exception e) {
+                                            logger.info(String.valueOf(e));
```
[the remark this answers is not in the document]
anfee1: This was used for debugging.

central/src/main/java/ai/djl/serving/central/http/BadRequestException.java
```diff
@@ -0,0 +1,40 @@
...
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package ai.djl.serving.central.http;
+
+/** Thrown when a bad HTTP request is received. */
+public class BadRequestException extends IllegalArgumentException {
```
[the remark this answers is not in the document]
anfee1: Well it was to say if somehow there is an error where a request did not have all of the parameters that are expected in the request.

central/src/main/webapp/components/ModelView.jsx
ebamberg: @zachgk yes sure I can open a issue about that. That sounds sensible. Can I do that on the project tab?

for other actions: I can think of "deploy model" (see serving, deploy to 1 or more remote serving instances) , "upload model", "create docker" , "remove model",  "invalidate model" to flag a model as "don't use anymore for new  projects" or so. just a few ideas, maybe some more fancy actions like "convert model" to convert a model from keras to TFLite for example.

central/src/main/java/ai/djl/serving/central/http/BadRequestException.java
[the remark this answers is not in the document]
ebamberg: "serving" uses exactly the same exception for the same purpose I guess. so this would be constistent to "serving", especially when we move central as plugin into "serving" 
+1 from me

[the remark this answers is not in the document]
zachgk: Yeah, if serving has the same one then we should keep it to prioritize consistency. This refactor could be revisited later once serving and central are merged

central/src/main/webapp/components/ModelView.jsx
zachgk: Github requires write permissions to add it on the project tab. If you go ahead and create a normal issue, I will add it to the project tab for you (feel free to add other issues too).

zachgk (approved):

central/src/main/webapp/components/ModelView.jsx
ebamberg: > Github requires write permissions to add it on the project tab. If you go ahead and create a normal issue, I will add it to the project tab for you (feel free to add other issues too).

already done.
I have created a issue for this: https://github.com/awslabs/djl/issues/735

PR #724, opened 2021-03-04
update onnxruntime along with string tensor

lanking520:
## Description ##

This change introduced the String tensor Support for ONNXRuntime along with the 1.7.0 version bump.


frankfliu (approved):

onnxruntime/onnxruntime-engine/src/main/java/ai/djl/onnxruntime/engine/OrtNDManager.java
```diff
@@ -63,6 +63,26 @@ public OrtNDArray create(Buffer data, Shape shape, DataType dataType) {
         }
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public NDArray create(String data) {
+        return create(new String[] {data});
```
zachgk, line 69: Should have shape here. This converts from [] to [1]

PR #742, opened 2021-03-12
remove exclusion for pytorch and tf from jacoco

lanking520:
## Description ##

Add PyTorch and TensorFlow back to Jacoco report

### DIFF
--- a/tools/gradle/jacoco.gradle
+++ b/tools/gradle/jacoco.gradle
@@ -14,7 +14,7 @@ task jacocoMergeTestData(type: JacocoMerge) {
     }
 }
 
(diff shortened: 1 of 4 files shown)


tools/gradle/jacoco.gradle
```diff
@@ -14,7 +14,7 @@ task jacocoMergeTestData(type: JacocoMerge) {
     }
 }
 
-def exclusions = [":examples", ":integration", ":pytorch:pytorch-engine", ":tensorflow:tensorflow-engine"]
```
zachgk, line 17: It may be worth adding the PtCoverageTest as well

zachgk (approved):

PR #749, opened 2021-03-16
Use setter for requiresGrad()

frankfliu:
## Description ##

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/NDArray.java
+++ b/api/src/main/java/ai/djl/ndarray/NDArray.java
@@ -161,19 +161,10 @@ public interface NDArray extends NDResource {
      * Attaches a gradient {@code NDArray} to this {@code NDArray} and marks it so {@link
      * ai.djl.training.GradientCollector#backward(NDArray)} can compute the gradient with respect to
      * it.
-     */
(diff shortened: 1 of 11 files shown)


stu1130 (approved):

zachgk (approved):

api/src/main/java/ai/djl/ndarray/NDArray.java
```diff
@@ -157,23 +157,14 @@ default boolean isScalar() {
...
-
     /**
      * Attaches a gradient {@code NDArray} to this {@code NDArray} and marks it so {@link
      * ai.djl.training.GradientCollector#backward(NDArray)} can compute the gradient with respect to
      * it.
      *
      * @param requiresGrad if {@code NDArray} requires gradient or not
      */
-    void requiresGradient(boolean requiresGrad);
+    void setRequiresGradient(boolean requiresGrad);
```
stu1130, line 167: do we still want a default setRequirersGradient()? do that sound like Java-friend API?

frankfliu: It's a bit confusing here, even we what to keep it, it should be requireGradient()

PR #757, opened 2021-03-18
Adding build.gradle and changes to nightly publish

aksrajvanshi:
Added extra statement to add djl-serving tar file to s3 bucket. Made changes to the serving build.gradle file.

*Major review needed for the cloudfront line - Line 90* since I could not test it.

### DIFF
--- a/.github/workflows/nightly_publish.yml
+++ b/.github/workflows/nightly_publish.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Copy files to S3 with the AWS CLI
           if: ${{ github.event.inputs.mode == 'snapshot' }}
           run: |
-            aws s3 cp serving/build/distributions/serving-0.11.0-SNAPSHOT.tar s3://djl-ai-akshay/publish/djl-serving-${{ github.event.inputs.mode }}/serving-0.11.0-${{ github.event.inputs.mode }}.tar
+            aws s3 cp serving/build/distributions/serving-0.11.0-SNAPSHOT.tar s3://djl-ai-akshay/publish/djl-serve-${{ github.event.inputs.mode }}/serving-0.11.0-${{ github.event.inputs.mode }}.tar
 
   test-pytorch:
     if: github.repository == 'awslabs/djl'
--- a/.github/workflows/nightly_publish.yml
+++ b/.github/workflows/nightly_publish.yml
@@ -86,7 +86,8 @@ jobs:
       - name: Copy files to S3 with the AWS CLI
           if: ${{ github.event.inputs.mode == 'snapshot' }}
           run: |
-            aws s3 cp serving/build/distributions/serving-0.11.0-SNAPSHOT.tar s3://djl-ai-akshay/publish/djl-servin-${{ github.event.inputs.mode }}/serving-0.11.0-${{ github.event.inputs.mode }}.tar
+            aws s3 cp serving/build/distributions/serving-0.11.0-SNAPSHOT.tar s3://djl-ai/publish/djl-serving-${{ github.event.inputs.mode }}/serving-0.11.0-${{ github.event.inputs.mode }}.tar \
+            aws cloudfront create-invalidation --distribution-id E371VB8JQ6NRVY --paths "djl-serving-${{ github.event.inputs.mode }}"
 
   test-pytorch:
     if: github.repository == 'awslabs/djl'
--- a/.github/workflows/nightly_publish.yml
+++ b/.github/workflows/nightly_publish.yml
@@ -167,5 +167,6 @@ jobs:
       - name: Copy files to S3 with the AWS CLI
         if: ${{ github.event.inputs.mode == 'staging' }}
         run: |
+          ./gradlew serving:dT
           aws s3 cp serving/build/distributions/serving-*.tar s3://djl-ai/publish/djl-serving/
           aws cloudfront create-invalidation --distribution-id E371VB8JQ6NRVY --paths "/djl-serving/*"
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly_publish.yml
```diff
@@ -77,6 +77,17 @@ jobs:
         with:
           name: mxnet-model-zoo
           path: mxnet/mxnet-model-zoo/build/reports
+      - name: Configure AWS Credentials
```
zachgk, line 80: @frankfliu Do we want to be publishing the updates to serving nightly and not as part of releases?

lanking520: I am thinking we should make it as a different job or even different workflow. The nightly publish is only tied to Maven uploads for now

.github/workflows/nightly_publish.yml
```diff
@@ -77,6 +77,17 @@ jobs:
...
+          uses: aws-actions/configure-aws-credentials@v1
+          with:
+            aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+            aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+            aws-region: us-west-2
+      - name: Copy files to S3 with the AWS CLI
+          if: ${{ github.event.inputs.mode == 'snapshot' }}
+          run: |
+            aws s3 cp serving/build/distributions/serving-0.11.0-SNAPSHOT.tar s3://djl-ai/publish/djl-serving-${{ github.event.inputs.mode }}/serving-0.11.0-${{ github.event.inputs.mode }}.tar \
```
lanking520, line 89: shouldn't we remove `-SNAPSHOT` if `inputs.mode` is 'release'?

.github/workflows/nightly_publish.yml
frankfliu: This should be part of release, instead of checking 'snapshot', it should be 'staging'

.github/workflows/nightly_publish.yml
aksrajvanshi: so when it gets copied to s3, it'll take the name of the input mode. I can add a new condition for release or staging as per requirement.

.github/workflows/nightly_publish.yml
aksrajvanshi: Ok. I can change the input mode to staging.

.github/workflows/nightly_publish.yml
```diff
@@ -77,6 +77,17 @@ jobs:
...
+      - name: Configure AWS Credentials
+          uses: aws-actions/configure-aws-credentials@v1
+          with:
+            aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+            aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+            aws-region: us-west-2
+      - name: Copy files to S3 with the AWS CLI
+          if: ${{ github.event.inputs.mode == 'staging' }}
+          run: |
+            aws s3 cp serving/build/distributions/serving-0.11.0-SNAPSHOT.tar s3://djl-ai/publish/djl-serving-${{ github.event.inputs.mode }}/serving-0.11.0-${{ github.event.inputs.mode }}.tar \
```
frankfliu, line 89: Do we need gzip the tar file?

frankfliu (approved):

.github/workflows/nightly_publish.yml
aksrajvanshi: No, we need tar file only. That'll do it nicely for the brew package

.github/workflows/nightly_publish.yml
```diff
@@ -167,5 +167,6 @@ jobs:
       - name: Copy files to S3 with the AWS CLI
         if: ${{ github.event.inputs.mode == 'staging' }}
         run: |
+          ./gradlew serving:dT
```
frankfliu, line 170:
```suggestion
          ./gradlew serving:dT -Pstaging
```

frankfliu (approved):

.github/workflows/nightly_publish.yml
```diff
@@ -142,18 +142,31 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
       - name: Publish to snapshot repository
+        if: ${{ github.event.inputs.mode == "" || github.event.inputs.mode == "snapshot" }}
```
lanking520, line 145: There are only two possibilities for the input modes right? snapshot or staging. Why we need such complicated mechanism? By default snapshot is set.

lanking520:
```
- name: Publish to snapshot/staging repository
  run: |
    ./gradlew publish -P${{ github.event.inputs.mode }}
    cd bom
    ./gradlew publish -P${{ github.event.inputs.mode }}
```

aksrajvanshi: @lanking520 If we have the inputs.mode as empty string, the publish will be empty. I think we'll need an if condition for an empty string and another one for the staging.

lanking520: We had a default value set on the top as:

```
default: 'snapshot'
```
Unless the person trigger the job with empty string purposefully this condition will be met. Another issue is if the user mispelled `staging` or `snapshot` what will happened? It does not throw any errors and just skip these publish stages which is not expected. If you really would like to handle the cases for a String with value other than staging/snapshot, you should do ${{ github.event.inputs.mode != "staging" && github.event.inputs.mode != "snapshot" }}

PR #758, opened 2021-03-19
Fix LSTM nightly build failure

stu1130:
We do the double embedding (one in the model for training, the other in the Translator for prediction).
The PR removes the embedding part in Translator and just convert the token to indices without duplicate embedding


frankfliu:
Why it works earlier?

frankfliu (approved):

PR #774, opened 2021-03-22
Load config.properties relative to model server home

frankfliu:
## Description ##

### DIFF
--- a/serving/src/main/java/ai/djl/serving/Arguments.java
+++ b/serving/src/main/java/ai/djl/serving/Arguments.java
@@ -12,6 +12,7 @@
  */
 package ai.djl.serving;
 
+import ai.djl.serving.util.ConfigManager;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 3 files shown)


aksrajvanshi (approved):

serving/src/main/java/ai/djl/serving/Arguments.java
```diff
@@ -82,13 +83,19 @@ public Path getConfigFile() {
...
             }
             return file;
         }
-        Path file = Paths.get("config.properties");
+        Path modelServerHome = Paths.get(ConfigManager.getModelServerHome());
+        Path file = modelServerHome.resolve("conf/config.properties");
+        if (Files.isRegularFile(file)) {
+            return file;
+        }
+        file = modelServerHome.resolve("conf/config.properties");
```
goswamig, line 98: what is difference between line `94` and `98` ?

aksrajvanshi: correct. @frankfliu I think the line 94 and 98 are repeated.

PR #775, opened 2021-03-22
fix paddleOCR size issue

lanking520:
## Description ##

The word recognition OCR model from PaddlePaddle require height to be fixed 32 or an error will be thrown.

### DIFF
--- a/paddlepaddle/paddlepaddle-model-zoo/src/main/java/ai/djl/paddlepaddle/zoo/cv/wordrecognition/PpWordRecognitionTranslator.java
+++ b/paddlepaddle/paddlepaddle-model-zoo/src/main/java/ai/djl/paddlepaddle/zoo/cv/wordrecognition/PpWordRecognitionTranslator.java
@@ -64,7 +64,7 @@ public class PpWordRecognitionTranslator implements Translator<Image, String> {
     @Override
     public NDList processInput(TranslatorContext ctx, Image input) {
         NDArray img = input.toNDArray(ctx.getNDManager());
-        int[] hw = resize32(input.getHeight(), input.getWidth());
+        int[] hw = resize32(input.getWidth());
         img = NDImageUtils.resize(img, hw[1], hw[0]);
         img = NDImageUtils.toTensor(img).sub(0.5f).div(0.5f);
         img = img.expandDims(0);
@@ -77,14 +77,9 @@ public class PpWordRecognitionTranslator implements Translator<Image, String> {
         return null;
     }
 
-    private int[] resize32(double h, double w) {
-        double min = Math.min(h, w);
-        if (min < 32) {
-            h = 32.0 / min * h;
-            w = 32.0 / min * w;
-        }
-        int h32 = (int) h / 32;
-        int w32 = (int) w / 32;
-        return new int[] {h32 * 32, w32 * 32};
+    private int[] resize32(double w) {
+        // Paddle doesn't rely on aspect ratio
+        int width = ((int) Math.max(32, w)) / 32 * 32;
+        return new int[] {32, width};
     }
 }


paddlepaddle/paddlepaddle-model-zoo/src/main/java/ai/djl/paddlepaddle/zoo/cv/wordrecognition/PpWordRecognitionTranslator.java
```diff
@@ -84,7 +84,8 @@ public Batchifier getBatchifier() {
             w = 32.0 / min * w;
         }
         int h32 = (int) h / 32;
-        int w32 = (int) w / 32;
-        return new int[] {h32 * 32, w32 * 32};
+        int w32 = Math.max((int) w / 32 / h32, 1);
+        // height must be 32 (fixed)
+        return new int[] {32, w32 * 32};
```
stu1130, line 89: if height is fixed to 32 shouldn't the algorithm be just w = 32 / h * w ?

lanking520: We still need to make sure width are base 32 number and has proper aspect ratio

frankfliu: Does this work:

int ratio = Math.floorDiv(w, h);

lanking520: No, this doesn't apply here.

lanking520: floorDiv cannot be used on floating number

frankfliu: As Jake pointed out, if you height is always 32, the width should simply be 32 * ratio,
no need to test if size is less then 32 or not. the whole function can be:
```
int ratio =  Math.max((int) w / h, 1);
return new int[] {32, 32 * ratio};
```

lanking520: I tried this way, the last inference result is always wrong. Will try this implementAction again.

paddlepaddle/paddlepaddle-model-zoo/src/main/java/ai/djl/paddlepaddle/zoo/cv/wordrecognition/PpWordRecognitionTranslator.java
```diff
@@ -77,14 +77,8 @@ public Batchifier getBatchifier() {
...
-        double min = Math.min(h, w);
-        if (min < 32) {
-            h = 32.0 / min * h;
-            w = 32.0 / min * w;
-        }
-        int h32 = (int) h / 32;
-        int w32 = (int) w / 32;
-        return new int[] {h32 * 32, w32 * 32};
+    private int[] resize32(double w) {
+        int width = ((int) Math.max(32, w)) / 32 * 32;
```
frankfliu, line 81: This is not correct. This basically keep the original width and removed the remainder.

lanking520: No, it trimmed the width to a 32 based number. The apect ratio didn't help on the inference result

lanking520: Yes

frankfliu (approved):

PR #777, opened 2021-03-23
Fix config.properties lookup

frankfliu:
## Description ##


zachgk (approved):

aksrajvanshi (commented):
So the reviewers on the homebrew side suggest we write the logs to #{testpath}/log and #{var}/log. 
Screen Shot 2021-03-23 at 11 28 23 AM


Moreover, the curl ping is getting refused too. 
Screen Shot 2021-03-23 at 11 34 03 AM



Also, I'm planning to assert the log file since curl seems to be causing an issue with their build system. I saw the same thing being done with tomcat too.

PR #797, opened 2021-03-26
Plugin Manager for Serving component

ebamberg:
## Description ##

This adds a way to load plug-ins into serving component. Final idea is to allow for instance "central" to run as plugin into serving or to have a sharable StaticFileRequestHandler in serving as well as in central.

This is an early PR of this functionality before the PR is getting to huge. Just to simplify Code reviews and discussion by splitting the work up into different parts. 

This PR covers, restructuring the gradle project into serving-plugins-servings-api and serving-core.
Implementing  a PluginManager that allows to load plugins (using Java SPI to detect service implementations ) from a folder and register RequestHandler to the management-port (for now).

TODO for the next parts:
- UnitTest coverage for the new Components
- - convert central to be a plug-in
- inject other Managers like ModelManager into to plugins (calling setters in the plugins objects if there are any)
- PluginLifecycle (=State Engine) "initializing, active" which allows to injects dependent plugins into plugins using calls if setters. (a plugin only goes into active state after all dependencies are full-filled) - maybe a lightweight solution without lifecycle is also possible  when just have 2 lists- list of unitialized plugins ,list of fully initialized plugins  


frankfliu:
@ebamberg 
This PR is a bit hard to merge:
1. Our CI depends on serving folder structure, we need update the Github Actions and AWS CodeBuilder build spec
2. The artificateId of serving is : ai.djl.serving, rename it serving-core requires quite a lot of build script change, and it also impact brew installation scripts

I suggest to only push serving one level down and not rename to serving-core. From folder structure point of view, this is not the best, but isn't too bad.

Let me create a PR to just push serving one level down, and you can rebase on that change.

ebamberg:
> @ebamberg
> This PR is a bit hard to merge:
> 
> 1. Our CI depends on serving folder structure, we need update the Github Actions and AWS CodeBuilder build spec
> 2. The artificateId of serving is : ai.djl.serving, rename it serving-core requires quite a lot of build script change, and it also impact brew installation scripts
> 
> I suggest to only push serving one level down and not rename to serving-core. From folder structure point of view, this is not the best, but isn't too bad.
> 
> Let me create a PR to just push serving one level down, and you can rebase on that change.

ok I see the problem. yeah sure let us push serving down one level. as you said not perfect but will work.

frankfliu:
@ebamberg 
I created a PR to push serving and central one level down: https://github.com/awslabs/djl/pull/798

ebamberg:
> @ebamberg
> I created a PR to push serving and central one level down: #798

I have seen the PR i#798 is merged, I will tidy up and test mine PR then based on this.

ebamberg:
@frankfliu I did some manual fixes and smoke testing after your rebase/project restructuring. 
See commit 05d0492873445c9a59a1c798dcb6faf23251b946
Should be ready for code review again

frankfliu:
@ebamberg sorry for the delay.

I made some changes on top your PR:
1. Some cosmetic changes such as javadoc and variable name 
2. Addressed spotBugs issues
3. I moved serving-api back to server. I agree it's much cleaner to have a separated package, but we have to maintain another maven package. Our serving.jar itself is tiny, it should be fine to directly depend on serving. Let's keep it together for time being, and separate them in future if needed.

Let me know if you have any concern. I will merge this PR once you review my changes.

ebamberg:
> @ebamberg sorry for the delay.
> 
> I made some changes on top your PR:
> 
> 1. Some cosmetic changes such as javadoc and variable name
> 2. Addressed spotBugs issues
> 3. I moved serving-api back to server. I agree it's much cleaner to have a separated package, but we have to maintain another maven package. Our serving.jar itself is tiny, it should be fine to directly depend on serving. Let's keep it together for time being, and separate them in future if needed.
> 
> Let me know if you have any concern. I will merge this PR once you review my changes.

The changes are ok with me. yes please merge it.

My only small concern - but which is no show stopper for me - is about moving back the api project and let plugins depend on the whole serving project. In this way we cannot control the functionality used by plugins. Plugins can use every class of serving. So changes on serving-core-class can break plugins and we cannot control versioning - in meaning of we cannot guarantee that new plugins run on older serving-instance and vice-versa cause we don't have a restricted shared public api (=interfaces).

But I don't expect a wave of 3rd party plugins for now :-) so I am OK to keep it together for now and do the change in future as you said, Just wanted to have the idea and concern documented so we can pick it up later from here if required.

btw. I already started to change "central" to run as a plugin. I will raise another PR for this as soon as it is finished and stable tested.

frankfliu (approved):

frankfliu:
@ebamberg 
I completely agree with you that we should have a serving-api package, but I just want to keep it lightweight for the time being.

PR #800, opened 2021-03-30
Improve TensorFlow new architecture refactoring

stu1130:
1. move the EagerSession to engine level
2. add operators back
3. use cached input metatdata so we don't do session feed and fetch on every inference but only on the first one
4. TfOpExecutor requires users to specify the # of outputs (99% op have only 1 ouput and the rest should be easy to infer the # output from the operator & parameters itself) for better memory usage & potential performance gain

### DIFF
--- a/tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
+++ b/tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
@@ -12,7 +12,6 @@
  */
 package ai.djl.tensorflow.engine;
 
-
 import ai.djl.Device;
 import ai.djl.engine.Engine;
 import ai.djl.ndarray.BaseNDManager;
@@ -28,20 +27,14 @@ import java.nio.DoubleBuffer;
 import java.nio.FloatBuffer;
 import java.nio.IntBuffer;
 import java.nio.LongBuffer;
-import java.util.concurrent.atomic.AtomicReference;
-import org.tensorflow.internal.c_api.TFE_Context;
 import org.tensorflow.internal.c_api.TFE_TensorHandle;
 
 public class TfNDManager extends BaseNDManager {
 
     static final TfNDManager SYSTEM_MANAGER = new SystemManager();
 
-    private AtomicReference<TFE_Context> eagerSessionHandle;
-    private static Integer seed;
-
     private TfNDManager(NDManager parent, Device device) {
         super(parent, device);
-        eagerSessionHandle = new AtomicReference<>(JavacppUtils.createEagerSession(true, 0, null));
     }
 
     static TfNDManager getSystemManager() {
@@ -52,14 +45,6 @@ public class TfNDManager extends BaseNDManager {
     @Override
     public ByteBuffer allocateDirect(int capacity) {
         return ByteBuffer.allocateDirect(capacity).order(ByteOrder.nativeOrder());
-    }
-
-    public static void setRandomSeed(Integer seed) {
-        TfNDManager.seed = seed;
-    }
-
-    public static Integer getRandomSeed() {
-        return seed;
     }
 
     /** {@inheritDoc} */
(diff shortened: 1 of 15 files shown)


tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
```diff
@@ -233,7 +218,7 @@ public NDArray randomNormal(float loc, float scale, Shape shape, DataType dataTy
         NDArray axes = create(shape.getShape());
         TfOpExecutor opBuilder =
                 opExecutor("RandomStandardNormal").addInput(axes).addParam("dtype", dataType);
-        if (seed != null) {
+        if (Engine.getEngine("TensorFlow").getSeed() != null) {
             opBuilder.addParam("seed", 1234);
```
frankfliu, line 222: This looks like a bug, we should use the see instead of 1234

tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
```diff
@@ -233,7 +218,7 @@ public NDArray randomNormal(float loc, float scale, Shape shape, DataType dataTy
         NDArray axes = create(shape.getShape());
         TfOpExecutor opBuilder =
                 opExecutor("RandomStandardNormal").addInput(axes).addParam("dtype", dataType);
-        if (seed != null) {
+        if (Engine.getEngine("TensorFlow").getSeed() != null) {
```
frankfliu, line 221: if (getEngine().getSeed() != null)

tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
```diff
@@ -212,7 +197,7 @@ public NDArray randomUniform(float low, float high, Shape shape, DataType dataTy
         NDArray axes = create(shape.getShape());
         TfOpExecutor opBuilder =
                 opExecutor("RandomUniform").addInput(axes).addParam("dtype", dataType);
-        if (seed != null) {
+        if (Engine.getEngine("TensorFlow").getSeed() != null) {
             opBuilder.addParam("seed", 1234);
```
frankfliu, line 201: hard coded 1234, seems a bug

tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
```diff
@@ -212,9 +197,9 @@ public NDArray randomUniform(float low, float high, Shape shape, DataType dataTy
         NDArray axes = create(shape.getShape());
         TfOpExecutor opBuilder =
                 opExecutor("RandomUniform").addInput(axes).addParam("dtype", dataType);
-        if (seed != null) {
-            opBuilder.addParam("seed", 1234);
-            opBuilder.addParam("seed2", 1234);
+        if (getEngine().getSeed() != null) {
```
frankfliu, line 200: 1. getSeed() is called multiple times, define a variable instead
2. Should the two seed the same?, should we do something seed + 1?

tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
```diff
@@ -254,17 +239,8 @@ public TfNDManager newSubManager(Device device) {
...
-    /** {@inheritDoc} */
-    @Override
-    public void close() {
-        super.close();
-        TFE_Context handle = eagerSessionHandle.getAndSet(null);
-        if (handle != null && !handle.isNull()) {
-            handle.close();
-        }
+        return new TfOpExecutor(
+                this, ((TfEngine) Engine.getEngine("TensorFlow")).getEagerSession(), operation);
```
frankfliu, line 243: getEngine()

tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
stu1130: I checked tf source code and change it

frankfliu (approved):

PR #805, opened 2021-03-31
add finding BlockFactory feature in model loading

lanking520:
## Description ##

Support BlockFactory searching and loading in the current Model directory. This method should work on:

```
factory.class
my_fancy_jar.jar
```

We will scan through this and figure out if there is a class implements BlockFactory.

### DIFF
--- a/api/src/main/java/ai/djl/util/Utils.java
+++ b/api/src/main/java/ai/djl/util/Utils.java
@@ -15,39 +15,26 @@ package ai.djl.util;
 import ai.djl.ndarray.NDArray;
 import ai.djl.nn.Parameter;
 import java.io.ByteArrayOutputStream;
-import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.lang.reflect.Constructor;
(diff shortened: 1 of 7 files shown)


api/src/main/java/ai/djl/util/Utils.java
```diff
@@ -349,4 +362,118 @@ public static Path getCacheDir() {
...
+     * @param <T> the Template T for the output Class
+     * @return the Class implementation
+     */
+    public static <T> T findImplementation(Path path, String className) {
+        try {
+            Path classesDir = path.resolve("classes");
+            // we only consider .class files and skip .java files
+            List<Path> jarFiles =
+                    Files.list(path)
+                            .filter(p -> p.toString().endsWith(".jar"))
```
goswamig, line 382: it it possible we have some other type of file as well ?

lanking520: We only scan on the .class files. These mostly only live in Java jars.

frankfliu (approved):

PR #806, opened 2021-04-01
Fix TensorFlow memory leak & inplace flaky test

stu1130:
## Description ##

### DIFF
--- a/tensorflow/tensorflow-api/build.gradle
+++ b/tensorflow/tensorflow-api/build.gradle
@@ -5,9 +5,10 @@ configurations {
 }
 
 dependencies {
-    api "org.bytedeco:javacpp:1.5.4"
+    // javacpp bug fix https://github.com/bytedeco/javacpp/commit/7f27899578dfa18e22738a3dd49701e1806b464a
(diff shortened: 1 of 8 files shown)


tensorflow/tensorflow-api/build.gradle
```diff
@@ -5,7 +5,8 @@ configurations {
 }
 
 dependencies {
-    api "org.bytedeco:javacpp:1.5.4"
+    // javacpp bug fix https://github.com/bytedeco/javacpp/commit/7f27899578dfa18e22738a3dd49701e1806b464a
+    api "org.bytedeco:javacpp:1.5.6-SNAPSHOT"
```
lanking520, line 9: When will this package becomes non-snapshot?

lanking520 (approved):

tensorflow/tensorflow-api/build.gradle
stu1130: I asked in the issue https://github.com/awslabs/djl/issues/690 but Samuel didn't answer it. will follow up with him

lanking520 (approved):

PR #810, opened 2021-04-02
Reduce the Conv testing during normal run

zachgk:
Currently, the convolution testing tests with a range of inputs across a number
of variables. This increases the test duration a bit much for the standard unit
test suite.

Instead, I reduce down to only two tests for unit testing. During nightly, the
full combination of ranges will be used.

### DIFF
--- a/api/src/test/java/ai/djl/nn/convolutional/Conv1dTest.java
+++ b/api/src/test/java/ai/djl/nn/convolutional/Conv1dTest.java
@@ -40,7 +40,16 @@ public class Conv1dTest extends OutputShapeTest {
                         * strideWidthRange.size()
                         * dilationWidthRange.size();
 
-        IntStream.range(0, (int) rows)
+        IntStream streamToTest;
+        if (Boolean.getBoolean("nightly")) {
+            // During nightly testing, test all rows
+            streamToTest = IntStream.range(0, (int) rows);
+        } else {
+            // During unit testing, only test the first and last
+            streamToTest = IntStream.of(0, (int) rows - 1);
+        }
+
+        streamToTest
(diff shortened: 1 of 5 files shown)


api/src/test/java/ai/djl/nn/convolutional/Conv1dTest.java
```diff
@@ -40,7 +40,16 @@ public void testOutputShapes() {
                         * strideWidthRange.size()
                         * dilationWidthRange.size();
 
-        IntStream.range(0, (int) rows)
+        IntStream streamToTest;
+        if (Boolean.getBoolean("nightly")) {
+            // During nightly testing, test all rows
+            streamToTest = IntStream.range(0, (int) rows);
+        } else {
```
goswamig, line 47: What are the chances that unit will pass and nightly will fail ? 
Do we have alarm on nightly tests?

goswamig (approved):
Thank you for doing this. I have two questions otherwise LGTM.

api/src/test/java/ai/djl/nn/convolutional/Conv1dTest.java
zachgk: I would say the chances should be pretty low. Most of the complexity is on the engine side, so we just need to call it appropriately. It has been a standard practice of ours to move the slower tests to nightly to keep the unit test suite fast.

We don't have alarms on nightly, but we probably should. However, that is a problem on its own and I think should be handled separately from this. We already have many nightly only or nightly expanded tests, so this one doesn't change it much.

aksrajvanshi (approved):

PR #814, opened 2021-04-03
Face detection examples: retinaface model, lightweight model.

mymagicpower:
## Description ##

Add two face detection examples: retinaface model, lightweight model.


examples/src/main/java/ai/djl/examples/inference/face/model/Landmark.java
```diff
@@ -0,0 +1,29 @@
...
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package ai.djl.examples.inference.face.model;
+
+import ai.djl.modality.cv.output.Point;
+import java.util.List;
+
+public class Landmark {
```
lanking520, line 19: You can place this class under https://github.com/awslabs/djl/tree/master/api/src/main/java/ai/djl/modality/cv/output.  You may also need to follow other classes in the same directory to have annotation on on public classes and methods

examples/src/main/java/ai/djl/examples/inference/face/model/FaceDetectedObjects.java
```diff
@@ -0,0 +1,36 @@
...
+ * and limitations under the License.
+ */
+
+package ai.djl.examples.inference.face.model;
+
+import ai.djl.modality.cv.output.BoundingBox;
+import ai.djl.modality.cv.output.DetectedObjects;
+import java.util.List;
+
+public class FaceDetectedObjects extends DetectedObjects {
```
lanking520, line 20: If you believe this method is also general applies to other Face detection applications, I would suggest put it under: https://github.com/awslabs/djl/tree/master/api/src/main/java/ai/djl/modality/cv/output. Same applies to the annotation for public classes and methods

examples/src/main/java/ai/djl/examples/inference/face/FaceDetectionTranslator.java
```diff
@@ -0,0 +1,210 @@
...
+        this.scales = scales;
+        this.steps = steps;
+    }
+
+    @Override
+    public NDList processInput(TranslatorContext ctx, Image input) {
+        width = input.getWidth();
+        height = input.getHeight();
+        NDArray array = input.toNDArray(ctx.getNDManager(), Image.Flag.COLOR);
+        array = array.transpose(2, 0, 1).flip(0); // HWC -> CHW RGB -> BGR
```
lanking520, line 67: Do you think if we can replace most of the preprocessing with toTensor?

ToTensor in DJL: https://github.com/awslabs/djl/blob/master/api/src/main/java/ai/djl/ndarray/internal/NDArrayEx.java#L421

Maybe also Normailize method defined in DJL: https://github.com/awslabs/djl/blob/master/api/src/main/java/ai/djl/ndarray/internal/NDArrayEx.java#L411

We had built in Transform function that can apply these transformation: https://javadoc.io/doc/ai.djl/api/latest/ai/djl/modality/cv/transform/package-summary.html

Lastly, is RGB -> BGR really necessary here? In most of the cases the input image is following RGB format

lanking520 (commented):
Thanks a lot for your contribution. In order to pass CI testing, you may needs to add some comments on the public functions you defined. Please do the following command before push:

```
./gradlew fJ
```

This will automatically fix the unused imports and reformat the code.

frankfliu:
@mymagicpower 
This looks great. I made some changes on top your PR. 
The boundingbox seems a bit off, I'm not sure it's caused by my change or not. But I think we can merge this and figure it out later.

frankfliu:
> @mymagicpower
> This looks great. I made some changes on top your PR.
> The boundingbox seems a bit off, I'm not sure it's caused by my change or not. But I think we can merge this and figure it out later.

I found the bug and fixed. Now it good to go.

frankfliu (approved):

lanking520 (approved):

PR #825, opened 2021-04-05
[WIP]Model Uploads

anfee1:
## Description ##

This allows a user to upload one file and have the file name sent to the ModelUpload server. This currently does not have the functionality to test the file type. This functionality will be added in future commits. This will not be sending the files to an actual data center. This is not currently ready to be merged.

Screen Shot 2021-04-05 at 12 30 04 PM
Screen Shot 2021-04-05 at 12 29 56 PM
Screen Shot 2021-04-05 at 12 28 57 PM

### DIFF
--- a/serving/central/src/main/webapp/components/UploadFab.jsx
+++ b/serving/central/src/main/webapp/components/UploadFab.jsx
@@ -42,10 +42,12 @@ export default function UploadFab() {
         console.log(event.target.files[0])
     }
 
-    const handleClose = (event) => {
-           setShow(false);
+    const handleUpload = (event) => {
+       setShow(false);
+       if (selectedFile){
            alert("Your file is being uploaded!")
-           axios.get("http://"+window.location.host+"/uploading/models?modelName="+selectedFile.name)
+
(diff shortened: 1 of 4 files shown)


serving/central/src/main/webapp/components/UploadFab.jsx
```diff
@@ -47,7 +47,7 @@ export default function UploadFab() {
        if (selectedFile){
            alert("Your file is being uploaded!")
 
-           axios.post("http://"+window.location.host+"/uploading/models?modelName="+selectedFile.name)
+           axios.get("http://"+window.location.host+"/uploading/models?modelName="+selectedFile.name)
```
zachgk, line 50: Post is right. The convention with HTTP methods is to use post to create things (https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods). You won't be able to use the query parameters though (no ?modelName=...). See the post example on https://github.com/axios/axios

anfee1: This change was meant to be apart of another commit. But, you are correct.

PR #832, opened 2021-04-08
Central as plugin

ebamberg:
## Description ##

"Central" can be used as Plugin in serving by copying the central jar into plugin-folder of serving.

"Central" is still available as standalone application.

"serving" accepts classes implementing JsonSerializer as plugins. 

to be continued.....


frankfliu:
I updated your PR.

The JsonSerializer should be be registered int JsonUtils, it has global impact. The customization of json serialization should only be localized in the plugin itself. I removed MetaDataSerializer for now, It's not clear to me why we need this class.

I created a JsonSerializable class, this will work if the class only has one serialization form. If we need to support multiple serialization form, we need have a registration per plugin.

frankfliu (approved):

api/src/main/java/ai/djl/modality/Classifications.java
```diff
@@ -130,6 +138,12 @@ public Classifications(List<String> classNames, NDArray probabilities) {
         return null;
     }
 
+    /** {@inheritDoc} */
```
ebamberg, line 141: This is a violation against separation of concern (SoC) principle.

if there are will be other forms of serialisation should every class implements its own way of toJson, ToSoap, toProtoBuf, toJMS ? 
I would say a class should not know about how it can be serialized

ebamberg:
> I updated your PR.
> 
> The JsonSerializer should be be registered int JsonUtils, it has global impact. The customization of json serialization should only be localized in the plugin itself. I removed MetaDataSerializer for now, It's not clear to me why we need this class.
> 
> I created a JsonSerializable class, this will work if the class only has one serialization form. If we need to support multiple serialization form, we need have a registration per plugin.

@frankfliu  Did you test the code after the change? I am sure this breaks the UI and you are not able to view any Model-Info. I expect a javascript error in ModelView.jspx cause no metadata object is defined. (MetaDataSerializer  is needed for this)

The new JsonSerializable-interface violates separation of concerns principle. A class should not know about every way it can get serialized.
Let's think about if HTTP-Request-with-Json-response is the only way that serving will support in support!? 
Serialisation is a cross cutting concern, in my experience to ignore this will cause maintainability problems in future.
also adds unnecessary dependency of classes like Classifications to GSON.

You made a valid point when 2 plugins need to serialize the same class in 2 different ways, but I think the JsonSerializable interface doesn't solve this problem neither. 
Original PR has a global effects on the particular instance of serving when a plugin is installed, but with the plugin-mechanism you are going to configure serving for your specific purpose anyway. IMO a smaller problem then the SoC-problem at this time.

Open to discuss

frankfliu:
@ebamberg 
I didn't test UI part, I thought the default Json serialization for Metadata is compatible with MetaDataSerializer.

The problem of using JsonUtils.GSON_BUILDER is it's global, If you changed this builder at one part of the code, it impact the whole system. The original code broke basicdataset testing.

And JsonUtils is a utility class, it should not have the knowledge of individual class. Adding Classification and DetectedObjects in it a really a hack.

JsonSerializable should be just a default for the class, just like Comparable to a Class, Ideally, djl-serving should be able register JsonSerializer at plugin level.

frankfliu:
@ebamberg 
I added MetaDataSerializer back and now UI is working.

I have some concern about serialize transient field. The Artifact and Metadata has circular reference, with transient field, json serialization will cause a dead loop, and that's why we need a customized MetaDataSerializer. I think the right solution would be create a proper ModelMetadata object instead of using Artifact.

ebamberg:
> @ebamberg
> I added MetaDataSerializer back and now UI is working.
> 
> I have some concern about serialize transient field. The Artifact and Metadata has circular reference, with transient field, json serialization will cause a dead loop, and that's why we need a customized MetaDataSerializer. I think the right solution would be create a proper ModelMetadata object instead of using Artifact.

@frankfliu yes fully agreed. I have the same concern. the circular refernce

ebamberg:
> @ebamberg
> I added MetaDataSerializer back and now UI is working.
> 
> I have some concern about serialize transient field. The Artifact and Metadata has circular reference, with transient field, json serialization will cause a dead loop, and that's why we need a customized MetaDataSerializer. I think the right solution would be create a proper ModelMetadata object instead of using Artifact.

@frankfliu yes fully agreed. I have the same concern. Out of a glut feeling I would also prophecy the circular reference will cause more issues in future. When I look in the code this reference seems only to exist so the MetaData-object can be re-used to deserialize metadata in the repository class, but not to get used as an object but to access some primitive data-fields. Am I right, or do I miss something ?  agreed that a proper ModelMetaData object is the proper solution. I guess you are talking about swap the current impl with a proper one in the core. Another way for the specific problem could to be add a ModelMataData-DTO to transfer from central-backend to frontend. Both will work. 1 option probably would avoid future issues due a better design, 2nd option would be a quick win for central. So both are fine in my opinion.  btw why do we have all the other transient fields in Artifacts, like Version and so on. I somehow could see why they have to be transient except to avoid decoupling of serializer. Just curious to understand this design.

frankfliu:
@ebamberg 
When I say ModelMetadata object, I'm actually mean ModelMataData-DTO.
The metadata was original used. as a top level object: https://github.com/awslabs/djl/blob/master/model-zoo/src/test/resources/mlrepo/model/cv/image_classification/ai/djl/zoo/mlp/metadata.json

The transient Metadata field in Artifact class is just a hack to make it easy for code, not meant to be serialized.

We can merge this PR and improve it later.

PR #840, opened 2021-04-09
Skip flaky ModelServerTest on windows

zachgk:
See https://github.com/awslabs/djl/pull/836/checks?check_run_id=2300913821 as an example of an instance of the test failing on an unrelated PR.


frankfliu:
We don't need this any more

zachgk:
Closed in favor of #842

PR #841, opened 2021-04-09
[XGBoost] add XGBoost inference support

lanking520:
## Description ##

Add support for XGBoost model. Now only supports for Mac and Linux.

### DIFF
--- a/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
+++ b/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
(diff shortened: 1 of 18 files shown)


ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void close() {
+        if (handle != null && handle.get() != 0L) {
+            long pointer = handle.getAndSet(0L);
+            JniUtils.deleteDMatrix(pointer);
+        }
+        if (data != null) {
```
frankfliu, line 154: You don't need to check null

ml/xgboost/src/test/java/ai/djl/ml/xgboost/XgbModelTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+public class XgbModelTest {
+
+    @BeforeClass
+    public void downloadXGBoostModel() throws IOException {
+        Path modelDir = Paths.get("build/model");
+        DownloadUtils.download(
+                "https://resources.djl.ai/test-models/xgboost/regression.json",
```
frankfliu, line 39: how big is the json file? better check in to test/resources folder.

frankfliu (approved):

ml/xgboost/src/test/java/ai/djl/ml/xgboost/XgbModelTest.java
lanking520: Around 114kb

ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
lanking520: Then we essentially not collecting garbage here?

PR #845, opened 2021-04-10
Defer engine initialization

frankfliu:
## Description ##

In current implementation, all the engines are initialized at the beginning. This behavior isn't ideal when multiple engine jar files are included in the project. Some project may not need all the engines at runtime. This change provide flexibility that allows a project include all the jars in the class and only load required engine when needed.

### DIFF
--- a/mxnet/mxnet-engine/src/main/java/ai/djl/mxnet/engine/MxEngine.java
+++ b/mxnet/mxnet-engine/src/main/java/ai/djl/mxnet/engine/MxEngine.java
@@ -40,6 +40,7 @@ import java.nio.file.Paths;
 public final class MxEngine extends Engine {
 
     public static final String ENGINE_NAME = "MXNet";
+    static final int RANK = 1;
     private static final String MXNET_EXTRA_LIBRARY_VERBOSE = "MXNET_EXTRA_LIBRARY_VERBOSE";
 
     /** Constructs an MXNet Engine. */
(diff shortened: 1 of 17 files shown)


mxnet/mxnet-engine/src/main/java/ai/djl/mxnet/engine/MxEngine.java
```diff
@@ -40,6 +40,7 @@
 public final class MxEngine extends Engine {
 
     public static final String ENGINE_NAME = "MXNet";
+    static final int RANK = 1;
```
goswamig, line 43: wow :)

pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtEngine.java
```diff
@@ -37,6 +37,7 @@
     private static final Logger logger = LoggerFactory.getLogger(PtEngine.class);
 
     public static final String ENGINE_NAME = "PyTorch";
+    static final int RANK = 2;
```
goswamig, line 40: Is this something we decide? Why do we have to provide this real rank? 
User can specify their choice of engine and that becomes RANK 1 right ?

goswamig (approved):
LGTM

pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtEngine.java
frankfliu: The rank here is just a way to provide a deterministic way for the default engine.
1. User can always use -Dai.djl.default_engine to choose the default engine.
2. All the engine are available as long as they are included in class path

PR #849, opened 2021-04-12
Adding batch dot product operator

markbookk:
## Description ##

Adding batch dot product operator to MXNet

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/NDArray.java
+++ b/api/src/main/java/ai/djl/ndarray/NDArray.java
@@ -4630,7 +4630,7 @@ public interface NDArray extends NDResource {
      * Batchwise product of this {@code NDArray} and the other {@code NDArray}.
      *
      * <ul>
-     *   <li>batch_dot is used to compute dot product of x and y when x and y are data in batch,
+     *   <li>batchDot is used to compute dot product of x and y when x and y are data in batch,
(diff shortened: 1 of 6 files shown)


api/src/main/java/ai/djl/ndarray/NDArray.java
```diff
@@ -4625,4 +4625,30 @@ default NDArray oneHot(int depth) {
      *     href=https://d2l.djl.ai/chapter_linear-networks/softmax-regression.html#classification-problems>Classification-problems</a>
      */
     NDArray oneHot(int depth, float onValue, float offValue, DataType dataType);
+
+    /**
+     * Batchwise product of this {@code NDArray} and the other {@code NDArray}.
+     *
+     * <ul>
+     *   <li> batch_dot is used to compute dot product of x and y when x and y are data in batch, namely N-D (N >= 3) arrays in shape of (B0, …, B_i, :, :).
```
zachgk, line 4633: You got this from MXNet and there should be a next sentence so include that information as well. Also, feel free to edit this description for grammar, clarity, comprehendability, etc. Don't feel like you are beholden to their exact words. You can even write the description on your own if you prefer

markbookk: Cool thanks. Yeah for now I will just stick with this documentation but I will be sure to change it for any future additions if I consider it necessary.

frankfliu (approved):

PR #851, opened 2021-04-12
BertCodeDataset uses Vocabulary and clean Vocabulary

zachgk:
This refactor makes the BertCodeDataset use the standard vocabulary class
instead of it's custom Dictionary.

In addition, there are several improvements to the Vocabulary as well. The
SimpleVocabulary was renamed to DefaultVocabulary as it is not that simple. The
FtVocabulary was removed as the default can be used instead.

The pruning of the DefaultVocabulary was also improved. It adds the option to
limit the vocabulary to a fixed number of elements (by frequency) which is a
common usage in addition to requiring a minFrequency. This specifically was used
in the BertCodeDataset. It changes the default of the SimpleVocabulary as well
to not have a default minFrequency of 10 but to require an explicit one. I
believe this makes more sense as 10 is a very arbitrary number and users won't
expect their vocabularies to be randomly pruned.


extensions/fasttext/src/test/java/ai/djl/fasttext/CookingStackExchangeTest.java
```diff
@@ -90,7 +92,8 @@ public void testWord2Vec() throws IOException, MalformedModelException, ModelNot
 
             FtModel ftModel = (FtModel) model.getWrappedModel();
             FtWord2VecWordEmbedding fasttextWord2VecWordEmbedding =
-                    new FtWord2VecWordEmbedding(ftModel, new FtVocabulary());
+                    new FtWord2VecWordEmbedding(
+                            ftModel, new DefaultVocabulary(Arrays.asList("bread")));
```
frankfliu, line 96: Collections.singletonList("bread")

examples/src/main/java/ai/djl/examples/training/TrainBertOnCode.java
```diff
@@ -73,10 +73,12 @@ public static TrainingResult runExample(String[] args) throws IOException, Trans
...
     }
 
-    private static Model createBertPretrainingModel(int dictionarySize) {
+    private static Model createBertPretrainingModel(long vocabularySize) {
         Block block =
                 new BertPretrainingBlock(
-                        BertBlock.builder().micro().setTokenDictionarySize(dictionarySize));
+                        BertBlock.builder()
+                                .micro()
+                                .setTokenDictionarySize(Math.toIntExact(vocabularySize)));
```
frankfliu, line 81: Should we also change the .setTokenDictionarySize() function name?

examples/src/main/java/ai/djl/examples/training/util/BertCodeDataset.java
```diff
@@ -84,12 +81,11 @@ public void prepare(Progress progress) throws IOException, TranslateException {
...
         parsedFiles = files.stream().map(BertCodeDataset::parseFile).collect(Collectors.toList());
         // determine dictionary
-        Map<String, Long> countedTokens = countTokens(parsedFiles);
-        dictionary = buildDictionary(countedTokens, 35000);
+        vocabulary = buildVocabulary();
     }
 
-    public int getDictionarySize() {
-        return dictionary.tokens.size();
+    public long getDictionarySize() {
```
frankfliu, line 87: refactor function name?

examples/src/main/java/ai/djl/examples/training/TrainBertOnCode.java
zachgk: I'll do this in a later PR

examples/src/main/java/ai/djl/examples/training/util/BertCodeDataset.java
zachgk: I'll do this in a later PR

frankfliu (approved):

PR #855, opened 2021-04-13
[XGBoost] add option to Model

lanking520:
## Description ##

Add option to XGBoost model. User can define treelimit and Mode during definition

### DIFF
--- a/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbModel.java
+++ b/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbModel.java
@@ -62,7 +62,7 @@ public class XgbModel extends BaseModel {
                         .setMode(
                                 XgbSymbolBlock.Mode.valueOf(
                                         ((String) options.get("Mode"))
-                                                .toUpperCase(Locale.ENGLISH)));
+                                                .toUpperCase(Locale.ROOT)));
             }
             if (options.containsKey("TreeLimit")) {
                 ((XgbSymbolBlock) block)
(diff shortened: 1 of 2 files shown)


ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbSymbolBlock.java
```diff
@@ -109,6 +111,28 @@ public Long getHandle() {
             this.value = value;
         }
 
+        /**
+         * Gets the mode from name.
+         *
+         * @param name the name of the mode
+         * @return the corresponding name
+         */
+        public static Mode getMode(String name) {
```
frankfliu, line 120: We don't need this:
```
Model model = Model.valueOf("DEFAULT");
```

ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbModel.java
```diff
@@ -54,6 +55,20 @@ public void load(Path modelPath, String prefix, Map<String, ?> options) throws I
...
         }
         block = JniUtils.loadModel((XgbNDManager) manager, modelFile.toAbsolutePath().toString());
+        // set extra options
+        if (options != null) {
+            if (options.containsKey("Mode")) {
+                ((XgbSymbolBlock) block)
+                        .setMode(
+                                XgbSymbolBlock.Mode.valueOf(
+                                        ((String) options.get("Mode"))
+                                                .toUpperCase(Locale.ENGLISH)));
```
frankfliu, line 65: 1. Not sure if toUpperCase is needed. I prefer let user set it correctly.
2. User Locale.ROOT instead

goswamig (approved):

frankfliu (approved):

PR #874, opened 2021-04-19
Serving fix issues after pr on upstream branch

ebamberg:
## Description ##

no plugins were loaded cause plugin-folder was null. broken after merge last changes with PR, worked before.
fixed this issue.


frankfliu (approved):
Thanks for fixing this issue.

PR #883, opened 2021-04-21
Fix TensorFlow tile error

stu1130:
## Description ##

### DIFF
--- a/tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDArray.java
+++ b/tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDArray.java
@@ -43,7 +43,7 @@ public class TfNDArray extends NativeResource<TFE_TensorHandle> implements NDArr
     private Shape shape;
     private Device device;
     private TfNDManager manager;
-    private String name = "";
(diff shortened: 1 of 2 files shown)


tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDArray.java
```diff
@@ -1322,6 +1322,12 @@ public NDArray isNaN() {
     /** {@inheritDoc} */
     @Override
     public NDArray tile(long repeats) {
+        // tf tile doesn't support scalar
+        if (isScalar()) {
+            try (NDArray temp = reshape(1)) {
```
frankfliu, line 1327: Why tile did not fail and affect next operation? Looks like a TF bug?

frankfliu (approved):

lanking520 (approved):

tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDArray.java
stu1130: We never test scalar with tile op

PR #888, opened 2021-04-22
[tensorflow] create an EagerSession per NDManager

frankfliu:
## Description ##


tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
```diff
@@ -248,8 +255,22 @@ public TfNDManager newSubManager(Device device) {
...
+    @Override
+    public void close() {
+        super.close();
+        TFE_Context eagerSession = eagerSessionHandle.getAndSet(null);
+        if (eagerSession != null && !eagerSession.isNull()) {
+            eagerSession.close();
+        }
+    }
+
+    TfOpExecutor opExecutor(String operation) {
```
stu1130, line 268: maybe make the function public? The user can use it to call TF operator not in DJL

stu1130 (approved):
Checked the C api, there is no way to reset the eager session. So creating a new EagerSession handle is the only solution for now.

PR #904, opened 2021-04-27
Added support for setGraphExecutorOptimize with torchscript models.

hodovo:
## Description ##

This PR adds support for torch::jit::setGraphExecutorOptimize which allows the user to prevent model "warmup" periods while torchscript optimizes the model on GPU.

Users can disable the torchscript optimization with the following code:

```java
JniUtils.setGraphExecutorOptimize(false);
```

Since this feature is enabled by default in torchscript, it will only disable optimization when the the method above is called. Since JNI maintains an individual environment per thread, the above method must be called for each thread that is using the model which optimization should be disabled for.  

This change is backwards compatible and does not alter the usage of any existing code.

### DIFF
--- a/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtSymbolBlock.java
+++ b/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtSymbolBlock.java
@@ -114,6 +114,11 @@ public class PtSymbolBlock extends AbstractSymbolBlock implements AutoCloseable 
         if (first) {
             synchronized (PtSymbolBlock.class) {
                 if (first) {
+                    String optimizeProperty = System.getProperty("ai.djl.pytorch.setGraphExecutorOptimize");
+                    if (optimizeProperty != null && !optimizeProperty.isEmpty()) {
+                        boolean optimize = Boolean.parseBoolean(optimizeProperty);
+                        JniUtils.setGraphExecutorOptimize(optimize);
+                    }
                     inputDescriptions = new PairList<>();
                     outputDescriptions = new PairList<>();
                     for (NDArray array : inputs) {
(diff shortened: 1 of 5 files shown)


hodovo:
@frankfliu I created it as a draft pull request as this is my first pull request. Please have a look and you can mark it ready for review.

hodovo:
After doing some testing, there is currently a bug with multi-threaded inference where it will have the same multi-second delay after a few inferences. It does not happen single threaded.

I have fixed it by running `PyTorchLibrary.LIB.setGraphExecutorOptimize(false);` before every time `PyTorchLibrary.LIB.moduleForward` is called. This indicates that when using multiple threads, the value of the setGraphExecutorOptimize is being reset in the JNI code or it is not being respected.

I propose that we add another parameter to `PyTorchLibrary.LIB.moduleForward` which takes in whether setGraphExecutorOptimize is true or false and sets it each time.

I am going to continue investigating to see if I can figure out why the value is not being respected when set in the PtSymbolicBlock on the first inference and only when using more than one thread.

hodovo:
I was able to resolve this by calling `JniUtils.setGraphExecutorOptimize(false);` in each thread I planned to use. I do not have much experience with JNI, but it seems that because each thread provides its own environment to each JNI call, this causes the reset.

I think the best course of action would to be add a parameter to `PyTorchLibrary.LIB.moduleForward` as I mentioned above. Let me know what you think.

hodovo:
It might be a good idea to just allow the developer to utilize the JNI function how they want rather than focusing it one way or another. This flexibility will be needed in some situations, like if you want to load models on one thread with optimization and another set of models without optimization. With this global approach, it is much less flexible and it doesn't enforce any changes.

I just reverted my changes to PtSymbolBlock as I think flexibility is the way to go.

frankfliu (approved):

PR #919, opened 2021-04-30
[tensorflow] Fix tensorflow-api dependency

frankfliu:
## Description ##


stu1130 (commented):
I think we should document why customer need javacpp 1.5.6 on their production

stu1130 (approved):

PR #924, opened 2021-05-02
Added initial support for float16.

hodovo:
## Description ##

This PR adds initial support for float16 via two main features.

1. A new class `ToHalf` allows the user to add it at the end of their existing pipeline to convert the `NDArray` into a float16 format. For an example,

```java
public class YoloV5Pipeline extends Pipeline {

    public YoloV5Pipeline(int size, DataType dataType) {
        super(new Resize(size), new ToTensor());
        if (dataType == DataType.FLOAT16) {
            add(new ToHalf());
        }
    }
}
```

2. `NDArray.toFloatArray` has been modified to accept the float16 datatype and parse it accordingly. This automatically implements float16 in all existing float32 workloads without additional code changes. Since Java only supports float32, no changes are required on the users side either.

Please let me know if anything can be improved or if I am missing something. Since this is a change to `NDArray`, I would appreciate a thorough review and consideration of the side effects. Thank you for taking the time to review this PR :)


hodovo:
It seems I am not allowed to use the `short` variable type. Please let me know how we should move forward :)

api/src/main/java/ai/djl/modality/cv/transform/ToHalf.java
```diff
@@ -0,0 +1,30 @@
...
+ * and limitations under the License.
+ */
+package ai.djl.modality.cv.transform;
+
+import ai.djl.ndarray.NDArray;
+import ai.djl.ndarray.types.DataType;
+import ai.djl.translate.Transform;
+
+/** A {@link Transform} that converts the input to float16 if needed. */
+public class ToHalf implements Transform {
```
frankfliu, line 20: I don't think this is necessary, user can use Lambda function as Transform easily:
```
addTransform(a -> a.toType(DataType.FALOAT16, false);
```

api/src/main/java/ai/djl/util/Float16Utils.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
frankfliu, line 2:
```suggestion
 * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

api/src/test/java/ai/djl/util/Float16UtilsTest.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
frankfliu, line 2:
```suggestion
 * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

api/src/main/java/ai/djl/util/Float16Utils.java
```diff
@@ -0,0 +1,121 @@
...
+        return ret;
+    }
+
+    /**
+     * Converts an array of float32 values into a byte buffer of float16 values.
+     *
+     * @param floats an array of float32 values.
+     * @return a byte buffer with float16 values represented as shorts (2 bytes each).
+     */
+    public static ByteBuffer toByteBuffer(float[] floats) {
```
frankfliu, line 55: You might want to take NDManager as parameter:
```
ByteBuffer buffer = manager.allocateDirect(floats.length * 2);
```
This can avoid one memory copy to native code.

api/src/main/java/ai/djl/modality/cv/transform/ToHalf.java
hodovo: I think for simplicity it might be good to have this since the same thing could be said about ToTensor. I can remove it if everyone agrees that it is not needed though.

api/src/main/java/ai/djl/util/Float16Utils.java
```diff
@@ -0,0 +1,124 @@
...
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package ai.djl.util;
+
+import ai.djl.ndarray.NDManager;
+
```
frankfliu, line 16:
```suggestion
```

api/src/main/java/ai/djl/modality/cv/transform/ToHalf.java
frankfliu: 1. We added ToTensor to solve discoverability issue, toType() seems obvious to me
2. It's arguable if we add ToHalf, why we don't add ToUint8, ToFloat64 etc.

hodovo: Both points make sense, and I definitely agree with #2. The main reason I didn't prefer the lambda is due to the check if it should actually convert the type as well as whether it should be a copy or not. I will go ahead and remove it and users can implement their own via lambda or class.

api/src/main/java/ai/djl/util/Float16Utils.java
hodovo: Fixed.

hodovo:
@frankfliu Since I am utilizing short for its size (2 bytes), I don't think int is a suitable replacement. The checks will continue to fail as it does not want me using short. I will wait for your guidance on this. Thank you for the help :)

frankfliu:
> @frankfliu Since I am utilizing short for its size (2 bytes), I don't think int is a suitable replacement. The checks will continue to fail as it does not want me using short. I will wait for your guidance on this. Thank you for the help :)

The PMD doesn't like usage of short, what we need to do is just Suppress the warning:

@SuppressWarnings("PMD.AvoidUsingShortType")
public final class Float16Utils {

api/src/main/java/ai/djl/util/Float16Utils.java
```diff
@@ -0,0 +1,123 @@
...
+ * and limitations under the License.
+ */
+package ai.djl.util;
+
+import ai.djl.ndarray.NDManager;
+import java.nio.ByteBuffer;
+import java.nio.ShortBuffer;
+
+/** {@code Float16Utils} is a set of utilities for working with float16. */
+public final class Float16Utils {
```
frankfliu, line 20:
```suggestion
@SuppressWarnings("PMD.AvoidUsingShortType")
public final class Float16Utils {
```

frankfliu (approved):

PR #957, opened 2021-05-14
Fix NPE issue in benchmark

frankfliu:
## Description ##

### DIFF
--- a/examples/src/main/java/ai/djl/examples/inference/benchmark/util/Arguments.java
+++ b/examples/src/main/java/ai/djl/examples/inference/benchmark/util/Arguments.java
@@ -258,14 +258,14 @@ public class Arguments {
     }
 
     public Class<?> getInputClass() {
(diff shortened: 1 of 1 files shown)


examples/src/main/java/ai/djl/examples/inference/benchmark/util/Arguments.java
```diff
@@ -258,14 +258,14 @@ public String getOutputDir() {
     }
 
     public Class<?> getInputClass() {
-        if (inputShapes == null) {
+        if (inputShapes.isEmpty()) {
```
stu1130, line 261: how about inputShapes == null || inputShapes.isEmpty()

frankfliu: inputShapes is PairList and initialized in constructor.

stu1130 (approved):

PR #983, opened 2021-05-26
[pytorch] Load extran shared library for pytorch engine

frankfliu:
## Description ##

### DIFF
--- a/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtEngine.java
+++ b/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtEngine.java
@@ -22,6 +22,10 @@ import ai.djl.pytorch.jni.JniUtils;
 import ai.djl.pytorch.jni.LibUtils;
 import ai.djl.training.GradientCollector;
 import ai.djl.util.RandomUtils;
+import java.io.FileNotFoundException;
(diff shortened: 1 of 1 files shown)


pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtEngine.java
```diff
@@ -53,6 +57,21 @@ static Engine newInstance() {
             }
             logger.info("Number of inter-op threads is " + JniUtils.getNumInteropThreads());
             logger.info("Number of intra-op threads is " + JniUtils.getNumThreads());
+
+            String paths = System.getenv("PYTORCH_EXTRA_LIBRARY_PATH");
```
lanking520, line 61: This has nothing to do with PyTorch, can user just simply do System.load() on the so files?

frankfliu: 1. PyTorch custom ops shared library usually depends on libtorch.so, and they installed on different path. This requires libtorch.so file is loaded first.

2. For DJL serving, there is no code can be executed before engine is loaded. A environment variable can allow user load custom ops easily.

stu1130 (approved):

lanking520 (approved):

PR #986, opened 2021-05-28
Min max scaler util

ebamberg:
## Description ##

This is MinMaxScaler implementation similar to sklearn.preprocessing.MinMaxScaler.
wanted to share this with you, just in-case that such tooling is from interest to add to the djl repo.

I needed such function in some time-series-forecast-demo and I haven't seen a similar implementation in the repo.  So mainly matches functionality I needed.


ebamberg:
added a function to detach to MinMaxScaler from NDManager

frankfliu (approved):

PR #1005, opened 2021-06-09
[tensoflow] Add truncated-normal operation

AzizZayed:
## Description ##

Add support for the [Truncated Normal ](https://www.tensorflow.org/api_docs/python/tf/random/truncated_normal) operation to TensorFlow only. An UnsupportedOperationException is thrown for all other engines.

### DIFF
--- a/tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
+++ b/tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
@@ -244,6 +244,31 @@ public class TfNDManager extends BaseNDManager {
 
     /** {@inheritDoc} */
     @Override
+    public NDArray truncatedNormal(float loc, float scale, Shape shape, DataType dataType) {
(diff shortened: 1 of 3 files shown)


tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
```diff
@@ -20,15 +20,9 @@
...
-import java.nio.Buffer;
-import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
-import java.nio.DoubleBuffer;
-import java.nio.FloatBuffer;
-import java.nio.IntBuffer;
-import java.nio.LongBuffer;
 import org.tensorflow.internal.c_api.TFE_Context;
 import org.tensorflow.internal.c_api.TFE_TensorHandle;
+import java.nio.*;
```
frankfliu, line 25: We don't use *, please update your IntelliJ settings.

frankfliu (approved):

PR #1018, opened 2021-06-13
[api] add loadModel() to Critiera

frankfliu:
Change-Id: Ie5c9f59ec283c780c323b2a99c440f416369c5fb

## Description ##


zachgk (approved):
We should also move the listModels as well

PR #1029, opened 2021-06-16
Update code to use criteria.loadModel()

frankfliu:
Change-Id: I38ef601ab211ac32587e1db27ee0d8f88cf3c7d8

## Description ##


zachgk (approved):
We may also want to do a search in docs as well to replace

lanking520 (approved):

PR #1035, opened 2021-06-17
add paddle gpu in engine

lanking520:
## Description ##

Paddle GPU 10.1


paddlepaddle/paddlepaddle-engine/src/main/java/ai/djl/paddlepaddle/engine/PpEngine.java
```diff
@@ -106,6 +106,8 @@ public NDManager newBaseManager() {
     @Override
     public NDManager newBaseManager(Device device) {
         if (getAlternativeEngine() != null) {
+            // use CPU as a default to achieve best performance
+            device = Device.cpu();
```
frankfliu, line 110: return alternativeEngine.newBaseManager(Device.cpu());

frankfliu (approved):

PR #1045, opened 2021-06-22
Refactor BlockFactory interface

frankfliu:
Change-Id: I7e9aa60f541c00852c548332338ee0fc914ee92f

## Description ##


api/src/main/java/ai/djl/nn/BlockFactory.java
```diff
@@ -27,8 +30,11 @@
...
      *
-     * @param manager the manager to assign to block
+     * @param model the model of the block
+     * @param modelPath the directory of the model location
+     * @param arguments the block creation arguments
      * @return the uninitialized block
+     * @throws IOException if IO operation fails during creating block
      */
-    Block newBlock(NDManager manager);
+    Block newBlock(Model model, Path modelPath, Map<String, ?> arguments) throws IOException;
```
zachgk, line 39: Is this right? I thought the plan was to use BlockFactory to deprecate the `Map<String, ?> arguments` and replace it with serializing the BlockFactory?

frankfliu: The purpose of BlockFactory is to create the Block then you can load the parameters into the Block.
The problem is the BlockFactory may need extra information:
1. Some files in the model directory
2. some model specific arguments. If we hard-code those arguments, which will make the BlockFactory implementation not re-useable.

zachgk (approved):

Issue #937, opened 2021-05-04
Remove final from EasyTrain so that we can extend it


hmf:
## Description

I am working on data with labels but do not use those labels for learning and inference - unsupervised learning with auto-encoders. I do need the labels later for validation though, so creating a `DataSet` that sets the `labels` to be the same as the `data` and loosing the label information, won't solve my problem. To get this working, I changed  [EasyTrain.java](https://github.com/deepjavalibrary/djl/blob/05ad89af6a6c8e184b1222cc47ec3a3513b762ad/api/src/main/java/ai/djl/training/EasyTrain.java) of version 0.9.0, to set the `label`member equal to the `data` member. However, `EasyTrain` has been updated, and will in the future most probably be updated again. It is much easier for users to simply extend the class and change the required behavior. In my case just change the [trainSplit](https://github.com/deepjavalibrary/djl/blob/05ad89af6a6c8e184b1222cc47ec3a3513b762ad/api/src/main/java/ai/djl/training/EasyTrain.java#L116) and [validateSplit](https://github.com/deepjavalibrary/djl/blob/05ad89af6a6c8e184b1222cc47ec3a3513b762ad/api/src/main/java/ai/djl/training/EasyTrain.java#L170) methods. 

**Will this change the current api? How?**

Removing the `final` qualifier won't change the API, so most users won't even be aware of the change. 

**Who will benefit from this enhancement?**

Anyone that has to adapt the `EasyTrain` class to do something similar to the already existing class.

zachgk:
This seems reasonable. Would you like to open a PR?

You will probably have to add a "@SuppressWarnings" annotation to the EasyTrain as well because one of our checkers thinks utility classes should be final.

hmf:
On my to-do list. 
Thanks.

frankfliu:
@hmf
EasyTrain is a utility class. It doesn't make sense to have a class inherit to it. Even you have a subclass and you changed `trainSplit`, it's a completely independent class.

If you want to customize some of the behavior, the right way to do is add a new parameter to the training options, and EasyTrain will do different thing based on the option.

hmf:
@frankfliu How would one parametrize the requirements I described above? I don't think I can come up with a set of parameters that are generic enough to be useful for other users.  Also I only have this use case that seems very specific. Do you have any other functionality in mind that could/should be parametrized.  

TIA

PR #1047, opened 2021-06-24
Making utility class EasyTrain non-final. Requires suppressing "Final…

hmf:
…Class" style check. See issue #937.

## Description ##

This change was discussed in issue #937

I am working on data with labels but do not use those labels for learning and inference - unsupervised learning with auto-encoders. I do need the labels later for validation though, so creating a `DataSet` that sets the labels to be the same as the data and loosing the label information, won't solve my problem. To get this working, I changed `EasyTrain.java` of version 0.9.0, to set the labelmember equal to the data member. However, `EasyTrain` has been updated, and will in the future most probably be updated again. It is much easier for users to simply extend the class and change the required behavior. In my case just change the `trainSplit` and `validateSplit` methods.

Will this change the current api? How?
This change is a backward compatible. Removing the final qualifier won't change the API, so most users won't even be aware of the change.

Who will benefit from this enhancement?
Interesting edge cases to note here: sometimes we need to adapt training to use different type of labeling (in my case unsupervised learning).  Anyone that has to adapt the `EasyTrain` class to do something similar to the already existing class.

### DIFF
--- a/api/src/main/java/ai/djl/training/EasyTrain.java
+++ b/api/src/main/java/ai/djl/training/EasyTrain.java
@@ -27,7 +27,8 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 
 /** Helper for easy training of a whole model, a trainining batch, or a validation batch. */
-public final class EasyTrain {
+@SuppressWarnings("FinalClass")
+public class EasyTrain {
 
     private EasyTrain() {}
 


frankfliu:
@hmf 
I think your use case is valid, we need a way to customize trainSplit and validSplit.

But I don't think your solution can solve your problem. All the methods in EasyTrain is static, If you inherit the EasyTrain.trainSplit in child class, you won't be able to override the behavior.

Here is my suggestion:
1. Create a functioninterface in EasyTrain:
```
    @FunctionalInterface
    public interface SplitFunction {

        boolean apply(Trainer trainer, GradientCollector collector, BatchData batchData, Batch split);
    }
```

2. Add an overload:

```
    public static void trainBatch(Trainer trainer, Batch batch, SplitFunction function) {
```

3. Now you can implement your trainSplit and pass it to EasyTrain

hmf:
@frankfliu My apologies for not replying sooner. I was trying to see if I could analyze and respond, however I am currently overwhelmed with other work. I need time to think this over properly.

PR #1048, opened 2021-06-25
Change application from GAN to image-generation

AzizZayed:
## Description ##

Change the name of the application to show a problem such as Image Generation rather than a DL technique.

### DIFF
--- a/api/src/main/java/ai/djl/Application.java
+++ b/api/src/main/java/ai/djl/Application.java
@@ -64,6 +64,8 @@ public class Application {
                 return CV.ACTION_RECOGNITION;
             case "cv/word_recognition":
                 return CV.WORD_RECOGNITION;
+            case "cv/image_generation":
+                return CV.IMAGE_GENERATION;
             case "nlp":
                 return NLP.ANY;
             case "nlp/question_answer":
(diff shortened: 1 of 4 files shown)


zachgk (approved):

frankfliu:
This file need to be updated as well: https://github.com/deepjavalibrary/djl/blob/master/pytorch/pytorch-model-zoo/src/test/resources/mlrepo/model/cv/gan/ai/djl/pytorch/biggan-deep/metadata.json

It's location should also be moved to: pytorch/pytorch-model-zoo/src/test/resources/mlrepo/model/cv/image-generation/ai/djl/pytorch/biggan-deep/metadata.json

api/src/main/java/ai/djl/Application.java
```diff
@@ -183,7 +183,7 @@ public int hashCode() {
         Application WORD_RECOGNITION = new Application("cv/word_recognition");
 
         /** An application that accepts a seed and returns generated images. */
-        Application GAN = new Application("cv/gan");
+        Application IMAGE_GENERATION = new Application("cv/image_generation");
```
frankfliu, line 186: We also need implement creation of IMAGE_GENERATION instance:
```
    public static Application of(String path) {
    }

```

frankfliu (changes requested):

frankfliu (approved):

PR #1053, opened 2021-06-25
Use pytorch 1.8.1 for 0.12.0 release

frankfliu:
Change-Id: I9a2a62d2e3ce7f1c0845d70e14a6a0eee91b4bb1

## Description ##


lanking520:
That's not solve the problems. We should try to keep the versions.

lanking520 (approved):

PR #1060, opened 2021-06-29
make SpProcessor public

lanking520:
## Description ##

Per customer demand, make SpProcessor public to achieve better performance


frankfliu:
We should make SpTokenizer.getProcessor() public as well.

I think creation of SpProcessor should always through SpTokenizer.
We can keep newInstance() and loadModel package protected.

frankfliu (approved):

Issue #1056, opened 2021-06-27
UniversalSentenceEncoder embeddings don't change regardless of input


markhng525:
## Description
When I try to test out the UniversalSentenceEncoder based upon the example class `./examples/src/main/java/ai/djl/examples/inference/UniversalSentenceEncoder.java`

Using a different set of inputs, the _encoded_  _embeddings_ remain unchanged `float[][] embeddings1 = UniversalSentenceEncoder.predict(inputs1)`.

### Expected Behavior
I expect the `embeddings` and `embeddings1` values to be different

## How to Reproduce?
```
    public static void main(String[] args) throws IOException, ModelException, TranslateException {
        List<String> inputs = new ArrayList<>();
        inputs.add("The quick brown fox jumps over the lazy dog.");
        inputs.add("I am a sentence for which I would like to get its embedding");

        float[][] embeddings = UniversalSentenceEncoder.predict(inputs);

        List<String> inputs1 = new ArrayList<>();
        inputs1.add("There is a stray dog over there.");
        inputs1.add("I should see a new type of encoded output");


        float[][] embeddings1 = UniversalSentenceEncoder.predict(inputs1);

        if (embeddings == null | embeddings1 == null) {
            logger.info("This example only works for TensorFlow Engine");
        } else {
            for (int i = 0; i < inputs.size(); i++) {
                logger.info(
                        "Embedding for: " + inputs.get(i) + "\n" + Arrays.toString(embeddings[i]));
                logger.info(
                        "Embedding for: " + inputs1.get(i) + "\n" + Arrays.toString(embeddings1[i]));
            }
        }
    }
```

### Steps to reproduce
1. Simply run `UniversalSentenceEncoder#main` in IDE

## What have you tried to solve it?

1. Tried loading `universal-sentence-encoder` model from local filesystem downloaded from google
2. Created 2 different `model` instances in case `Predictor<String[], float[][]> predictor = model.newPredictor()` was caching something underneath the hood.

## Environment Information

Please run the command `./gradlew debugEnv` from the root directory of DJL (if necessary, clone DJL first). It will output information about your system, environment, and installation that can help us debug your issue. Paste the output of the command below:

```
--------- Environment Variables ---------
[...]

PR #1066, opened 2021-06-30
[tensorflow] Fix bug in string tensor implementation

frankfliu:
Change-Id: Ib15ec1568a96f8984e1c7c124b2d7e10a2d9bd71

Fixes: #1056

## Description ##


tensorflow/tensorflow-engine/src/main/java/ai/djl/tensorflow/engine/TfNDManager.java
```diff
@@ -101,8 +103,8 @@ public TfNDArray create(Buffer data, Shape shape, DataType dataType) {
     /** {@inheritDoc} */
     @Override
     public NDArray create(String data) {
-        TFE_TensorHandle handle = JavacppUtils.createStringTensor(data);
-        return new TfNDArray(this, handle);
+        Pair<TF_Tensor, TFE_TensorHandle> pair = JavacppUtils.createStringTensor(data);
```
stu1130, line 106: can you add comment why we need to maintain two pointer

stu1130 (approved):

PR #1135, opened 2021-07-31
Add buffer to file IO

frankfliu:
Change-Id: I5e270e2ed6d8878bdff624b7a8bb829b8bf1e671

## Description ##


lanking520 (approved):
Why bufferedIO is better than all other InputStream?

PR #1142, opened 2021-08-04
Change NDArray.toString() output

frankfliu:
Change-Id: I4c0b29e56fd6ed214656cf2681337e595ecda4e3

## Description ##

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/NDArray.java
+++ b/api/src/main/java/ai/djl/ndarray/NDArray.java
@@ -4400,6 +4400,16 @@ public interface NDArray extends NDResource {
     /**
      * Runs the debug string representation of this {@code NDArray}.
      *
+     * @return the debug string representation of this {@code NDArray}
+     */
+    default String toDebugString() {
+        int maxSize = Integer.getInteger("ai.djl.nd_display_limit", 0);
+        return NDFormat.format(this, maxSize, 10, 10, 20);
+    }
+
+    /**
+     * Runs the debug string representation of this {@code NDArray}.
+     *
      * @param maxSize the maximum elements to print out
      * @param maxDepth the maximum depth to print out
      * @param maxRows the maximum rows to print out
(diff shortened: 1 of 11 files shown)


api/src/main/java/ai/djl/ndarray/internal/NDFormat.java
```diff
@@ -48,6 +49,15 @@ public static String format(
             format = new BooleanFormat();
         } else if (dataType.isInteger()) {
             format = new IntFormat(array);
+        } else if (dataType == DataType.STRING) {
+            return "ND: "
+                    + array.getShape()
+                    + ' '
+                    + array.getDevice()
+                    + ' '
+                    + array.getDevice()
```
zachgk, line 58:
```suggestion
                    + array.getDataType()
```

api/src/main/java/ai/djl/ndarray/NDArray.java
```diff
@@ -4397,6 +4397,16 @@ default NDArray countNonzero(int axis) {
      */
     NDArrayEx getNDArrayInternal();
 
+    /**
+     * Runs the debug string representation of this {@code NDArray}.
+     *
+     * @return the debug string representation of this {@code NDArray}
+     */
+    default String toDebugString() {
+        int maxSize = Integer.getInteger("ai.djl.nd_display_limit", 0);
```
zachgk, line 4406: This makes the default maxSize 0, which means it won't show any NDArrays. We should probably have a larger default to allow some to be displayed. Maybe 100 like MxNDarray before?

frankfliu: Currently, toString() is called by default in debugger, retrieving the data makes the debugger very slow. The default should be 0. User can manually configure debugger to use toDebugString() when then expend the data view, this will be on-demand and won't impact performance.

We display the content mostly in Jupyter notebook, I'm consider add the property in the IJava kernel to workaround this issue.

zachgk: It would also affect println statements as well. Here is another option. What if we added something like [this check](https://www.rgagnon.com/javadetails/java-detect-if-running-in-debug-mode.html) to see if the JVM is running in debug mode?

frankfliu: Now we won't print ndarray content in debugger

zachgk (approved):

PR #1144, opened 2021-08-04
[djl-bench] Run benchmark on multiple GPUs

frankfliu:
Change-Id: Ie27e090699695526df42fefd01db2aba78dc3f73

## Description ##

### DIFF
--- a/extensions/benchmark/src/main/java/ai/djl/benchmark/MultithreadedBenchmark.java
+++ b/extensions/benchmark/src/main/java/ai/djl/benchmark/MultithreadedBenchmark.java
@@ -12,7 +12,9 @@
  */
 package ai.djl.benchmark;
 
+import ai.djl.Device;
 import ai.djl.ModelException;
+import ai.djl.engine.Engine;
 import ai.djl.inference.Predictor;
 import ai.djl.metric.Metrics;
 import ai.djl.repository.zoo.ZooModel;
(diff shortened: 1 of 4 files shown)


extensions/benchmark/src/main/java/ai/djl/benchmark/MultithreadedBenchmark.java
```diff
@@ -41,16 +43,22 @@
...
+        int numOfThreads = arguments.getThreads() * devices.length;
         int delay = arguments.getDelay();
         AtomicInteger counter = new AtomicInteger(iteration);
         logger.info("Multithreading inference with {} threads.", numOfThreads);
 
+        List<ZooModel<Void, float[]>> models = new ArrayList<>(devices.length);
         List<PredictorCallable> callables = new ArrayList<>(numOfThreads);
-        for (int i = 0; i < numOfThreads; ++i) {
-            callables.add(new PredictorCallable(model, metrics, counter, i, i == 0));
+        for (Device device : devices) {
```
zachgk, line 55: This means that the number of threads is really the number of threads per GPU. Is this the behavior users would expect? Should we document this somewhere or change some names to make it clearer?

frankfliu: number of threads per GPU is confusing, I change it to total thread and added error checking,
I also changed number of iteration in the same way.

zachgk (approved):

PR #1162, opened 2021-08-11
let paddle hold the creation buffer

lanking520:
## Description ##

Avoid random crashes during the bytebuffer early catch


paddlepaddle/paddlepaddle-engine/src/main/java/ai/djl/paddlepaddle/engine/PpNDArray.java
```diff
@@ -26,6 +26,7 @@
 public class PpNDArray extends NativeResource<Long> implements NDArrayAdapter {
 
     private PpNDManager manager;
+    private ByteBuffer data;
```
frankfliu, line 29: Add a comment here to avoid this been removed

frankfliu (approved):

PR #1203, opened 2021-09-05
[api] Refactor Predictor API

frankfliu:
1. Add default implemenation to getBatchifier()
2. pass TranslatorContext to prepare()
3. Add getBlock() and getPredictorManager() to TranslatorContext

Change-Id: I1e23fcae60d92905597d5e1f8dd583e58bc79eeb

## Description ##

### DIFF
--- a/api/src/main/java/ai/djl/inference/Predictor.java
+++ b/api/src/main/java/ai/djl/inference/Predictor.java
@@ -13,7 +13,6 @@
 package ai.djl.inference;
 
 import ai.djl.Model;
-import ai.djl.engine.EngineException;
 import ai.djl.metric.Metrics;
 import ai.djl.ndarray.LazyNDArray;
 import ai.djl.ndarray.NDArray;
@@ -135,7 +134,7 @@ public class Predictor<I, O> implements AutoCloseable {
         long begin = System.nanoTime();
         try (PredictorContext context = new PredictorContext()) {
             if (!prepared) {
-                translator.prepare(manager, model);
+                translator.prepare(context);
                 prepared = true;
             }
             Batchifier batchifier = translator.getBatchifier();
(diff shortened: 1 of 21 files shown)


api/src/main/java/ai/djl/translate/Translator.java
```diff
@@ -75,14 +72,16 @@
      *
      * @return the {@link Batchifier}
      */
-    Batchifier getBatchifier();
+    default Batchifier getBatchifier() {
+        return null;
```
zachgk, line 76: I'm not sure I agree with this default because it makes it confusing whether to translate Record -> Record or Batch -> Batch. So far, we have mainly focused on Record -> Record while treating the batch one as a special case while this default changes that. This would be especially confusing for training users.

What if we make the default translator use StackBatchifier for training. Then, add a new interface like `BatchTranslator extends Translator` with a `null` default that can explain how it affects translation and the pre/post-processing in comparison to regular usages of the translator.

api/src/main/java/ai/djl/inference/Predictor.java
```diff
@@ -284,6 +281,18 @@ public NDManager getNDManager() {
             return ctxManager;
         }
 
+        /** {@inheritDoc} */
+        @Override
+        public NDManager getPredictorManager() {
```
zachgk, line 286: Maybe just `getManager`?

zachgk (changes requested):

api/src/main/java/ai/djl/inference/Predictor.java
zachgk: Nevermind. This is on context rather than predictor so the clarification is helpful.

zachgk (approved):

PR #1213, opened 2021-09-10
Refactor alternative engine API

frankfliu:
1. make NDArrayAdapter base class
2. Fallback to alternative engine when needed
3. Allow NDArray auto convert between engines
4. Prepare for ops across different engine

Change-Id: Ia1499169c432d7a76e2df550a266b61240a55609

## Description ##

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
+++ b/api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
@@ -18,6 +18,7 @@ import ai.djl.ndarray.types.DataType;
 import ai.djl.ndarray.types.Shape;
 import ai.djl.ndarray.types.SparseFormat;
 import java.nio.Buffer;
+import java.util.Arrays;
 
 /**
  * A base implementation of the {@link NDArray} that does nothing. This can be used for overriding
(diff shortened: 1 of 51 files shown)


lanking520 (approved):

lanking520:
This change may introduce some latency increment from hidden conversion. Saying if I import two engines at the same time in my packages:

- MXNet
- PaddlePaddle

If I would like to use PaddlePaddle alone without touching other packages, it may by default create NDArray in MXnet and convert to Paddle.

frankfliu:
> This change may introduce some latency increment from hidden conversion. Saying if I import two engines at the same time in my packages:
> 
> * MXNet
> * PaddlePaddle
> 
> If I would like to use PaddlePaddle alone without touching other packages, it may by default create NDArray in MXnet and convert to Paddle.

With this change, it will create PaddlePaddle NDArray, not alternative engine, alternative engine is kick in only when the operation that the engine is not supported.

api/src/main/java/ai/djl/ndarray/NDManager.java
```diff
@@ -161,6 +161,14 @@ static NDManager from(NDResource resource) {
      */
     ByteBuffer allocateDirect(int capacity);
 
+    /**
+     * Converts an external engine's {@link NDArray} to current engine.
```
zachgk, line 165: Does this convert or create a new one? Both description and adopt seem to imply changing.

Maybe rename to `matching(array)`.

api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
```diff
@@ -27,908 +28,1029 @@
...
+    public NDArray sequenceMask(NDArray sequenceLength) {
+        return sequenceMask(sequenceLength, 0);
     }
 
     /** {@inheritDoc} */
     @Override
-    default NDArray zerosLike() {
-        throw new UnsupportedOperationException(UNSUPPORTED_MSG);
+    public NDArray zerosLike() {
+        return getManager().zeros(getShape(), getDataType(), getDevice());
```
zachgk, line 205: Some of these are fully generic and don't rely on anything adapter specific. We may want to move the definitions to NDArray instead. This could be in another PR if you prefer.

api/src/main/java/ai/djl/ndarray/NDManager.java
frankfliu: How about `from(NDArray)` ?

api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
frankfliu: I didn't move to NDArray is because I'm not sure if we should remove native implementation for MXNet and PyTorch

api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
```diff
@@ -27,908 +28,1029 @@
...
+    public NDArray onesLike() {
+        return getManager().ones(getShape(), getDataType(), getDevice());
     }
 
     /** {@inheritDoc} */
     @Override
-    default boolean contentEquals(Number number) {
-        throw new UnsupportedOperationException(UNSUPPORTED_MSG);
+    public boolean contentEquals(Number number) {
+        return number.equals(toArray()[0]);
```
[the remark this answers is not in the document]
frankfliu: Will fix it in next PR.

api/src/main/java/ai/djl/ndarray/NDManager.java
zachgk: Yeah, `from` is better. Go with that

api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
zachgk: We don't need to remove the native implementations. The ones in NDArray are just defaults and can be overridden for better performance

frankfliu: I don't think there is performance gain, the only thing it might saved is getShape() and. getDataType() call.

PR #1261, opened 2021-10-01
Update README.md

delgadofarid:
## Description ##

Documentation change:
pytorch-native-cpu isn't working for CentOS with glibc = 1.17, but pytorch-native-cpu-precxx11 is; therefore I think the docs should be more clear making the supported glibc version inclusive (>=1.17 rather than >1.17).


frankfliu (approved):

frankfliu:
@delgadofarid Thanks for your contribution.

PR #1273, opened 2021-10-06
Update Artifact.toString()

frankfliu:
Change-Id: I048ef91bbaee200d7d94970555e3d9c97d8efd52

## Description ##


api/src/main/java/ai/djl/repository/Artifact.java
```diff
@@ -319,13 +319,14 @@ public String toString() {
         StringBuilder sb = new StringBuilder(100);
         if (metadata != null) {
             sb.append(metadata.getGroupId())
-                    .append(':')
+                    .append('/')
                     .append(metadata.getArtifactId())
-                    .append(':');
```
zachgk, line 324: Doesn't there need to be a trailing '/' here? Otherwise, there wouldn't be a separating slash if you had both artifactId and version

PR #1331, opened 2021-10-30
[opencv] Add opencv extension

frankfliu:
Change-Id: I738397d316ae128dd66f6ff5de8d52bfc56927e2

## Description ##


extensions/opencv/src/main/java/ai/djl/opencv/OpenCVImageFactory.java
```diff
@@ -0,0 +1,106 @@
...
+        Mat img = Imgcodecs.imread(path.toAbsolutePath().toString());
+        if (img.empty()) {
+            throw new IOException("Read image failed: " + path);
+        }
+        return new OpenCVImage(img);
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public Image fromUrl(URL url) throws IOException {
```
zachgk, line 56: Can we make this a default on the ImageFactory?

extensions/opencv/src/main/java/ai/djl/opencv/OpenCVImageFactory.java
```diff
@@ -0,0 +1,106 @@
...
+    public Image fromNDArray(NDArray array) {
+        Shape shape = array.getShape();
+        if (shape.dimension() == 4) {
+            throw new UnsupportedOperationException("Batch is not supported");
+        } else if (shape.get(0) == 1 || shape.get(2) == 1) {
+            throw new UnsupportedOperationException("Grayscale image is not supported");
+        } else if (array.getDataType() != DataType.UINT8 && array.getDataType() != DataType.INT8) {
+            throw new IllegalArgumentException("Datatype should be INT8 or UINT8");
+        }
+        int width;
```
zachgk, line 91: Simpler to set the height and width after the transpose. Then, you don't even need an else case.

zachgk (changes requested):
Can you also add a README for the opencv extension and add that README to the mkdocs?

extensions/opencv/src/main/java/ai/djl/opencv/OpenCVImageFactory.java
frankfliu: OpenCV dependency is about 85M, I think this should be optional. You can include this extension only if they care about the image reading performance.

zachgk: I mean just the `ImageFactory.fromUrl` as it can be implemented in terms of InputStream

frankfliu: done

zachgk (approved):

saudet:
Why haven't you considered the [JavaCPP Presets for OpenCV](https://github.com/bytedeco/javacpp-presets/tree/master/opencv)? Could you please explain why you are trying so hard to avoid using anything related to JavaCPP? What's wrong with JavaCPP? I would more than happy to fix anything that prevents you from using JavaCPP, if only you could express those needs! Please do speak up!

frankfliu:
@saudet 

I want to make it clear that I'm not against JavaCPP at all. We do considered about JavaCPP, and we are still interested in using JavaCPP with PyTorch (Due to resource constraint, we paused this initiative. I happy to schedule a meeting with you if you can lead this initiative).

I actually started with JavaCPP Presets for OpenCV. Everything works perfect except the default dependency size is huge (480M). I didn't figure out an easy way to reduce the size (e.g. I failed exclude openblas)

I'm actually not happy with `org.openpnp:opencv` size (80M) either, but I don't have time to create a build pipeline and only compile needed feature myself.

saudet:
If you look only at the binaries of the presets for Linux, Mac, and Windows (x86_64), those are also about 80 MB: https://repo1.maven.org/maven2/org/bytedeco/opencv/4.5.3-1.5.6/
You don't need to depend on all of platforms in your application! Also, the average user is not going to be interested by binaries other than their development platform anyway, so you can have it download just that by default.

As for the dependency on OpenBLAS, it does accelerate OpenCV and PyTorch quite a bit, but neither OpenCV nor PyTorch distribute binaries with BLAS enabled, which I find weird. Are users of DJL also not concerned with performance? If you can say for sure that your users are unconcerned with performance, please let me know, and we'll see what we can do.

Thank you for clarifying your needs!

saudet:
BTW, when PyTorch is used, an alternative to OpenCV could be TorchVision. When all symbols are properly stripped from the libraries, the archives are only 1~2 MB per platform: https://pypi.org/project/torchvision/#files

frankfliu:
@saudet 
torchvision depends on python, which blocks us from integrating with torchvision, see: #1394

saudet:
Right, it does come with a few operations in C++, but looking more closely, it doesn't look too useful. It actually looks mostly like a wrapper around Pillow... In any case, we can enable building the Python libraries as part of the build for the JavaCPP Presets for PyTorch, and that would get us that missing library. BTW, I was thinking about this, and I think DJL could use the JavaCPP Presets for PyTorch as is, but simply rebundle it as necessary, similarly to how you're already doing it for TensorFlow for Java. Although I don't make releases frequently because it needs to some testing, I plan to continue updating the snapshots regularly, so you could pick those binaries and test those independently on your own, before making frequent releases of DJL. How does that sound?

PR #1399, opened 2021-12-02
add sentencepiece load from bytes method

lanking520:
## Description ##

Add SentencePiece load from byte array method.

### DIFF
--- a/extensions/sentencepiece/src/main/java/ai/djl/sentencepiece/SpTokenizer.java
+++ b/extensions/sentencepiece/src/main/java/ai/djl/sentencepiece/SpTokenizer.java
@@ -50,6 +50,16 @@ public class SpTokenizer implements Tokenizer, AutoCloseable {
         loadModel(modelPath, prefix);
     }
 
+    /**
+     * Creates a SentencePiece Tokenizer from byte array.
+     *
+     * @param serializedModel the serialized model
+     */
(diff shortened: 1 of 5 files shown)


extensions/sentencepiece/src/main/java/ai/djl/sentencepiece/SpTokenizer.java
```diff
@@ -50,6 +50,16 @@ public SpTokenizer(Path modelPath, String prefix) throws IOException {
         loadModel(modelPath, prefix);
     }
 
+    /**
+     * Create a SentencePiece Tokenizer from byte array.
```
frankfliu, line 54:
```suggestion
     * Creates a SentencePiece Tokenizer from byte array.
```

extensions/sentencepiece/src/main/java/ai/djl/sentencepiece/SpTokenizer.java
```diff
@@ -96,6 +106,10 @@ private void loadModel(Path modelPath, String prefix) throws IOException {
         processor.loadModel(modelFilePath);
     }
 
+    private void loadModel(byte[] serializedModel) {
```
frankfliu, line 109: This method is unnecessary

frankfliu (approved):

PR #1406, opened 2021-12-08
[huggingface] Adds huggingface tokenizer extension

frankfliu:
Change-Id: I1af56de1e841dbcf00c947d2feefb7800699831d

## Description ##


extensions/tokenizers/rust/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "djl"
+version = "0.1.0"
+authors = ["Frank Liu <frankfliu2000@gmail.com>"]
+edition = "2021"
+
+[dependencies]
+jni = "0.19.0"
+tokenizers = { path = "../tokenizers/tokenizers", version = "*" }
```
nguyenvietyen, line 9: I'd fix the version here, unless you want to run this in a CI and spot breaking changes.

extensions/tokenizers/rust/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+
+[lib]
+crate_type = ["cdylib"]
```
nguyenvietyen, line 12: Have you tried building a staticlib with JNI before? It would be easier to reuse.
(we gave it 1h, but didn't succeed on our side with JNA)

extensions/tokenizers/rust/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+
+#[no_mangle]
+pub extern "system" fn Java_ai_djl_huggingface_tokenizers_jni_TokenizersLibrary_createTokenizer(
+    env: JNIEnv,
+    _: JObject,
+    input: JString,
+) -> jlong {
+    let identifier: String = env
+        .get_string(input)
+        .expect("Couldn't get java string!")
```
nguyenvietyen, line 34: AFAIK, `.expect` will trigger a Rust panic and default crash the VM. Alternatively, the code could check for the erroneous case and use `env.throw` for a safer erroneous result handling.

extensions/tokenizers/rust/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+    for i in word_ids {
+        if let Some(word_id) = i {
+            long_ids.push(*word_id as jlong)
+        } else {
+            long_ids.push(-1)
+        }
+    }
+
+    let array: jlongArray = env.new_long_array(len).unwrap();
+    env.set_long_array_region(array, 0, &long_ids).unwrap();
```
nguyenvietyen, line 240: I don't know JNI well: does this mean that the VM becomes responsible for GCing the jlongArray? (which would be memory safe)

extensions/tokenizers/rust/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+    env: JNIEnv,
+    _: JObject,
+    handle: jlong,
+) -> jlongArray {
+    let encoding = cast_handle::<Encoding>(handle);
+    let word_ids = encoding.get_word_ids();
+    let len = word_ids.len() as jsize;
+    let mut long_ids: Vec<jlong> = Vec::new();
+    for i in word_ids {
+        if let Some(word_id) = i {
```
nguyenvietyen, line 232: I believe pattern matching on Some/None is more Rust-idiomatic here.

extensions/tokenizers/rust/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+    for i in special_token_masks {
+        long_ids.push(*i as jlong)
+    }
+
+    let array: jlongArray = env.new_long_array(len).unwrap();
+    env.set_long_array_region(array, 0, &long_ids).unwrap();
+    array
+}
+
+fn to_handle<T: 'static>(val: T) -> jlong {
```
nguyenvietyen, line 302: I'm unsure if using the `static` lifetime is recommend here. Have you tried your own lifetime like `a`?
(note my understanding of lifetimes is more theoretical than actually practised...)

extensions/tokenizers/rust/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+    let handle = Box::into_raw(Box::new(val)) as jlong;
+    handle
+}
+
+fn cast_handle<T>(handle: jlong) -> &'static mut T {
+    assert_ne!(handle, 0, "Invalid handle value");
```
nguyenvietyen, line 308: I think exposing `env` would be great here, so you can throw a JVM exception on this.

extensions/tokenizers/rust/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+
+    let ptr = handle as *mut T;
+    unsafe { &mut *ptr }
```
nguyenvietyen, line 311: Due to the way you pass the references, I do not think making it `mut` is necessary here. But maybe I overlooked something along the way.

extensions/tokenizers/rust/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+    array
+}
+
+#[no_mangle]
+pub extern "system" fn Java_ai_djl_huggingface_tokenizers_jni_TokenizersLibrary_getSpecialTokenMask(
+    env: JNIEnv,
+    _: JObject,
+    handle: jlong,
+) -> jlongArray {
+    let encoding = cast_handle::<Encoding>(handle);
```
nguyenvietyen, line 289: The pattern of transforming a Rust slice into a jlongArray kinda repeats itself here several times. It could be more DRY by making a more generic function transforming the Rust slice into a jLongArray, and accepting as input a reference to the Rust slice (e.g. special token mask).

nguyenvietyen (commented):
It looks super cool! 

I'm very curious how the performance is with this implementation. It definitively has less overhead.

extensions/tokenizers/rust/Cargo.toml
frankfliu: The version of tokenizers is defined in `gradle.perperties` file, we always clone a specific version of tokenizer to build the project.

extensions/tokenizers/rust/Cargo.toml
frankfliu: This module is purely used for JNI, I not sure where else will reused the staticlib. We can make it staticlib if a use case comes up.

extensions/tokenizers/rust/src/lib.rs
frankfliu: Maybe we should throw a OutOfMemoryError here. I will try to improve it in a separate PR.

extensions/tokenizers/rust/src/lib.rs
frankfliu: Yes, JVM will collect the return long[]

extensions/tokenizers/rust/src/lib.rs
frankfliu: I think this should be static, the object should be alive until java side explicitly close it. rust side doesn't know how long it should live.

extensions/tokenizers/rust/src/lib.rs
frankfliu: Agree, will address this issue in a separate PR.

extensions/tokenizers/rust/src/lib.rs
frankfliu: I actually don't really understand the code, I copied from: https://github.com/exonum/exonum-java-binding/blob/57aa684f45d8501bba7bf1c1b25eed257d2b94e7/exonum-java-binding/core/rust/src/utils/handle.rs#L83-L90

lanking520 (approved):

zachgk (approved):

PR #1436, opened 2021-12-29
Support for Tuples as input values for PyTorch models

zakhio:
This PR adds support for Tuples as input values for PyTorch models.

## Description ##

It is a simple PR which combines input values to Tuples if NDArray's name have '\w+()` syntax . Also there is a test.


frankfliu (approved):

frankfliu:
@zakhio 
Thanks for your contribution, this looks great.

PR #1478, opened 2022-01-20
[api] Fixes topK items for DetectedObjects and make it configurable to Classifications

frankfliu:
Change-Id: I4ffd29cff5205f32ee18afa5269d70491634936e

## Description ##


zachgk (approved):
I think we can do this for now, but I am not convinced it is the right long-term approach. It is part of a larger problem that I suspect we will deal with for all ServingTranslators in how the internal Java types are serialized and deserialized to Input/Output. I think we may need to add some ability to configure it and some ability to specify a Java function for full flexibility.

PR #1543, opened 2022-03-24
[pytorch] Allows multiple native jars to package into fat jar

frankfliu:
Fixes #1422

Change-Id: I477d85725e84f30a7a79733b16d0224abe83f3ce

## Description ##

### DIFF
--- a/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
+++ b/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
@@ -290,10 +290,24 @@ public final class LibUtils {
                 return new LibTorch(dir.toAbsolutePath(), platform, flavor);
             }
 
+            Matcher m = VERSION_PATTERN.matcher(version);
+            if (!m.matches()) {
+                throw new IllegalArgumentException("Unexpected version: " + version);
+            }
+            String[] versions = m.group(1).split("\\.");
+            int minorVersion = Integer.parseInt(versions[1]);
+            int buildVersion = Integer.parseInt(versions[2]);
+            String pathPrefix;
+            if (minorVersion > 10 || (minorVersion == 10 && buildVersion == 2)) {
+                pathPrefix = "pytorch/" + flavor + '/' + classifier;
+            } else {
+                pathPrefix = "native/lib";
+            }
+
             Files.createDirectories(cacheDir);
             tmp = Files.createTempDirectory(cacheDir, "tmp");
             for (String file : platform.getLibraries()) {
-                String libPath = "native/lib/" + file;
+                String libPath = pathPrefix + '/' + file;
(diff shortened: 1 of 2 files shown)


frankfliu:
@maziyarpanahi

With this PR, you should be able to package multiple pytorch native library ( > 1.10.2) in a single jar

zachgk (approved):

DevinTDHa:
Hi @frankfliu,

Any information on when `pytorch-native-cpu:1.10.2` would be published on maven?

Thanks!

frankfliu:
It should be release together with DJL 0.17.0, we usually release new version in 45-60 days.

DevinTDHa:
@frankfliu I've got a question regarding this PR:

We are using sbt to package the libraries for a fat jar, so it will include `pytorch-native-cpu` with the classifiers  `osx-x86_64` `linux-x86_64`  `win-x86_64`. However as I understand, the corresponding `native/lib/pytorch.properties` file is the same for all 3 platforms and some merging strategy needs to be used. How would you suggest dealing with this issue? Is there some way to distinguish the different platforms in the properties file?

lanking520:
> @frankfliu I've got a question regarding this PR:
> 
> We are using sbt to package the libraries for a fat jar, so it will include `pytorch-native-cpu` with the classifiers `osx-x86_64` `linux-x86_64` `win-x86_64`. However as I understand, the corresponding `native/lib/pytorch.properties` file is the same for all 3 platforms and some merging strategy needs to be used. How would you suggest dealing with this issue? Is there some way to distinguish the different platforms in the properties file?

From design point of view, you should not bundle three of the classifier (for Mac OS, Linux, Windows) into a single jar. Each OS should have their own jars. Fat jar is used mostly on the Spark related works (mostly Linux based). DJL is using the specific OS to determine which native jar to load

maziyarpanahi:
Hi @lanking520 

I think the question is regarding this discussion: https://github.com/deepjavalibrary/djl/discussions/1422 and more precisely this conclusion that something needs to happen to allow multiple CPU operating systems to be packaged all together: https://github.com/deepjavalibrary/djl/discussions/1422#discussioncomment-1860599

The problem with Apache Spark is that `spark-shell` and `spark-submit` don't support `classifier` in SBT. So it entirely discards that part and just randomly downloads the first one that it sees (which is always for the wrong operating system)

To go around this issue in Apache Spark, we package all 3 operating systems on the CPU into 1 fat jar. Obviously, the size of the package and the final Fat JAR will be larger than it should be, but it allows us to use that dependency that has a classifier on all operating systems, all machines, and all possible situations (spark-shell, spark-submit, etc.) - we do this with `tensorflow-java` as well since they use `classifier` too.

frankfliu:
@DevinTDHa 

Currently we don't have a good way to avoid the `pytorch.properties` conflict for fatjar.
I create a project that allow you load class from `.jar` file in a fatjar: https://github.com/frankfliu/junkyard/tree/master/jarjar

Please take a look and see if that works for you.

PR #1544, opened 2022-03-24
Support freeze parameters

### LINKED ISSUE #392: Freeze a block for training
## Question
Is there a way to freeze a block for transfer learning ? something similar to Keras **trainable = False** ?

### LINKED ISSUE #1494: Freezing some parameters during backpropagation
Hi,

I would like to know if it is possible to freeze some parameters whose ndarray indexes I want to provide between epochs.
Is is more for experimentation. Based on that, I might want to approach a branch&bound algorithm to optimize the loss value.

Any ideas welcome!

zachgk:
This adds a method to support freezing and unfreezing a parameter for transfer
learning. There is also a helper to (un)freeze all paremters in a block, but
without filtering. For more advanced use cases of (un)freezing part of the
parameters in a block, it should be implemented using block.getChildren(),
block.getDirectParameters() and/or block.getParameters().

fixes #392
fixes #1494

### DIFF
--- a/api/src/main/java/ai/djl/nn/Block.java
+++ b/api/src/main/java/ai/djl/nn/Block.java
@@ -272,9 +272,9 @@ public interface Block {
      * Freezes or unfreezes all parameters inside the block for training.
      *
      * @param freeze true if the parameter should be frozen
-     * @see Parameter#freeze(Boolean)
+     * @see Parameter#freeze(boolean)
      */
-    default void freezeParameters(Boolean freeze) {
+    default void freezeParameters(boolean freeze) {
         for (Parameter parameter : getParameters().values()) {
             parameter.freeze(freeze);
         }
--- a/api/src/main/java/ai/djl/nn/Block.java
+++ b/api/src/main/java/ai/djl/nn/Block.java
@@ -269,6 +269,18 @@ public interface Block {
             throws IOException, MalformedModelException;
 
     /**
+     * Freezes or unfreezes all parameters inside the block for training.
+     *
+     * @param freeze true if the parameter should be frozen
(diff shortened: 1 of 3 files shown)


api/src/main/java/ai/djl/nn/Parameter.java
```diff
@@ -133,6 +133,22 @@ public boolean requiresGradient() {
...
+     *
+     * <p>Sometimes during training, especially during transfer learning, it is typical to train
+     * only part of the model. For this, the freeze can be used to prevent certain parts from being
+     * trained.
+     *
+     * <p>This modifies the {@link #requiresGradient()} of the parameter.
+     *
+     * @param freeze true if the parameter should be frozen ({@code freeze == !requiresGradient()})
+     */
+    public void freeze(Boolean freeze) {
```
frankfliu, line 147:
```suggestion
    public void freeze(boolean freeze) {
```

api/src/main/java/ai/djl/nn/Block.java
```diff
@@ -268,6 +268,18 @@ default NDList forward(
     void loadParameters(NDManager manager, DataInputStream is)
             throws IOException, MalformedModelException;
 
+    /**
+     * Freezes or unfreezes all parameters inside the block for training.
+     *
+     * @param freeze true if the parameter should be frozen
+     * @see Parameter#freeze(Boolean)
+     */
+    default void freezeParameters(Boolean freeze) {
```
frankfliu, line 277:
```suggestion
    default void freezeParameters(boolean freeze) {
```

frankfliu (approved):

PR #1555, opened 2022-04-07
Separate AbstractSymbolBlock from AbstractBlock

zachgk:
Right now, the AbstractSymbolBlock inherits from AbstractBlock. However, the
AbstractBlock parameters system and getParameters implies that it is always
possible to get the parameters for a block. While this should be true for a
correctly written DJL block, it is not always true for symbol blocks.

So, this change separates them to ensure that trying to get parameters when it
is not possible returns the exception reflecting that the operation is
unsupported. The shared functionality between AbstractSymbolBlock and
AbstractBlock was moved to a common base class, AbstractBaseBlock.

### DIFF
--- a/api/src/main/java/ai/djl/nn/AbstractBlock.java
+++ b/api/src/main/java/ai/djl/nn/AbstractBlock.java
@@ -12,25 +12,16 @@
  */
 package ai.djl.nn;
 
-import ai.djl.MalformedModelException;
 import ai.djl.ndarray.NDList;
 import ai.djl.ndarray.NDManager;
 import ai.djl.ndarray.types.DataType;
 import ai.djl.ndarray.types.Shape;
 import ai.djl.training.ParameterStore;
-import ai.djl.training.initializer.Initializer;
 import ai.djl.util.Pair;
 import ai.djl.util.PairList;
-import java.io.DataInputStream;
 import java.io.DataOutputStream;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.LinkedHashMap;
-import java.util.List;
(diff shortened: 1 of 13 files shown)


engines/mxnet/mxnet-engine/src/main/java/ai/djl/mxnet/engine/MxSymbolBlock.java
```diff
@@ -156,6 +160,11 @@ public PairList<String, Shape> describeInput() {
         return inputDescriptions;
     }
 
+    @Override
```
frankfliu, line 163:
```suggestion
    /** {@inheritDoc} */
    @Override
```

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtSymbolBlock.java
```diff
@@ -146,6 +151,42 @@ public PairList<String, Shape> describeInput() {
         return inputDescriptions;
     }
 
+    @Override
```
frankfliu, line 154:
```suggestion
    /** {@inheritDoc} */
    @Override
```

api/src/main/java/ai/djl/nn/AbstractBlock.java
```diff
@@ -100,7 +79,7 @@ public abstract class AbstractBlock implements Block {
 
     /** Constructs a new {@code AbstractBlock} instance. */
     public AbstractBlock() {
-        this((byte) 1);
+        super();
```
frankfliu, line 82:
```suggestion
```

frankfliu (approved):

PR #1569, opened 2022-04-13
allow empty batch data

patins1:
## Description ##

This fix handles the edge case of flattening a batch of zero size. It is a good  quality metric for a software function to be able to handle zero data input even if the calling function could detect this case and create a fake expected output (that was my work-around before). 

Before this fix, this exception would have been thrown:

Caused by: ai.djl.engine.EngineException: cannot reshape tensor of 0 elements into shape [0, -1] because the unspecified dimension size -1 can be any value and is ambiguous
	at ai.djl.pytorch.jni.PyTorchLibrary.torchReshape(Native Method)

### DIFF
--- a/api/src/main/java/ai/djl/nn/Blocks.java
+++ b/api/src/main/java/ai/djl/nn/Blocks.java
@@ -29,6 +29,10 @@ public final class Blocks {
      */
     public static NDArray batchFlatten(NDArray array) {
         long batch = array.size(0);
+		if (batch == 0) {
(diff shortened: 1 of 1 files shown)


zachgk (approved):

zachgk:
The change looks good @patins1, but it seems like you are failing the formatter check. Can you run `./gradlew formatJava` (see [docs](http://docs.djl.ai/docs/development/development_guideline.html#coding-conventions))?

PR #1575, opened 2022-04-16
[examples] Fix ImageClassification invalid probability

### LINKED ISSUE #1574: [examples] ImageClassification gives invalid probability.
## Description
As discussion #1566, `examples.inference.ImageClassification` gives probability >0 or <1.

Set `.optApplySoftmax(true)` to set a Softmax classification layer.

dandansamax:
## Description ##

- Set a softmax classification layer in `examples.inference.ImageClassification`.
- Write a test for Class `ImageClassification`
- close #1574


examples/src/test/java/ai/djl/examples/inference/ImageClassificationTest.java
```diff
@@ -0,0 +1,37 @@
+/*
```
frankfliu, line 1: We don't need this test, we have the same test in TrainMnistTest.java

dandansamax: OK, I miss that. I will delete it.

frankfliu: Please add check the value should be less than 1 in TrainMnistTest.java

dandansamax: I'm a little confused about the if statement `Boolean.getBoolean("nightly")`. When does it become true?

frankfliu: We have continuous and nightly build, see:
https://github.com/deepjavalibrary/djl/blob/master/.github/workflows/nightly_publish.yml#L37

frankfliu (approved):

Issue #46, opened 2020-04-08
Part of Speech Tagging Dataset


zachgk:
## Description
This is a task to add at least one part of speech tagging dataset. These datasets help provide an example of an NLP token classification task, as well as having some use for training multi-purpose NLP models. A good example might be one from [Universal Dependencies](https://universaldependencies.org).

AKAGIwyf:
Hi, I wonder if there are any other websites that also include this dataset, since the Penn Tree bank dataset in the Linguistic Data Consortium costs $1700.

zachgk:
That's a good point @AKAGIwyf. I changed it to use a different POS dataset which should be freely available

AKAGIwyf:
> That's a good point @AKAGIwyf. I changed it to use a different POS dataset which should be freely available

We've found a version of Penn Treebank which is free on github but without POS tags as Torchtext, it had been pre-processed and I've written the code for it. I wonder if you need this kind of dataset whether or not

zachgk:
@AKAGIwyf  If you want to add it, more datasets are always good. It provides options for users for which one they want to train. The main goal for this issue was to add at least one POS dataset

zachgk:
@AKAGIwyf, were you working on this besides the Penn Treebank? @LanAtGitHub is interested in working on this, but I don't want to give it to a second person if you have already started

zachgk:
Just wanted to make sure. It's assigned @LanAtGitHub

siddvenk:
The above PR added a POS dataset.

PR #1580, opened 2022-04-18
[basicdataset] Add PennTreebank dataset

### LINKED ISSUE #1579: Add Penn Tree Bank Dataset
## Description
Add the Penn Treebank dataset to the basic datasets for language modeling. The dataset is pre-processed.

AKAGIwyf:
## Description ##
Add a version of Penn Treebank which is free on github but without POS tags as Torchtext, it had been pre-processed.

close #1579.


basicdataset/src/test/resources/mlrepo/dataset/nlp/ai/djl/basicdataset/penntreebank/metadata.json
```diff
@@ -0,0 +1,40 @@
...
+    }
+  },
+  "artifacts": [
+    {
+      "version": "1.0",
+      "snapshot": false,
+      "name": "penntreebank",
+      "files": {
+        "train":{
+          "uri" : "https://raw.githubusercontent.com/wojzaremba/lstm/master/data/ptb.train.txt",
```
zachgk, line 23: Is this right? I looked at the data, but it seems like just pre-processed text. I was expecting to see some kind of labels like https://www.ling.upenn.edu/courses/Fall_2003/ling001/penn_treebank_pos.html on all of the words

AKAGIwyf: Yes, this issue is not for part of speech tagging dataset. I'd already mentioned about this pre-processed dataset in issue #46, we are also working on adding a new POS dataset for issue #46. But in this new issue I just want to add this pre-processed dataset as WikiText2 which is just text data for language modeling.

basicdataset/src/test/resources/mlrepo/dataset/nlp/ai/djl/basicdataset/penntreebank/metadata.json
```diff
@@ -0,0 +1,40 @@
+{
+  "metadataVersion": "0.2",
+  "resourceType": "dataset",
+  "application": "nlp",
+  "groupId": "ai.djl.basicdataset",
+  "artifactId": "penntreebank",
+  "name": "penntreebank",
+  "description": "The Penn Treebank (PTB) project selected 2,499 stories from a three year Wall Street Journal (WSJ) collection of 98,732 stories for syntactic annotation.",
+  "website": "https://blog.salesforceairesearch.com/the-wikitext-long-term-dependency-language-modeling-dataset/",
```
zachgk, line 9: This website is for wikitext. I recommend using the github repo you get the data from as the site, rather than the original site for the penn treebank

basicdataset/src/test/resources/mlrepo/dataset/nlp/ai/djl/basicdataset/penntreebank/metadata.json
```diff
@@ -0,0 +1,40 @@
...
+    "license": {
+      "name": "LDC User Agreement for Non-Members",
+      "url": "https://catalog.ldc.upenn.edu/license/ldc-non-members-agreement.pdf"
+    }
+  },
+  "artifacts": [
+    {
+      "version": "1.0",
+      "snapshot": false,
+      "name": "penntreebank",
```
zachgk, line 20: Maybe give some addition to the name like "penntreebank-unlabeled-processed"

basicdataset/src/main/java/ai/djl/basicdataset/nlp/PennTreebank.java
```diff
@@ -0,0 +1,193 @@
...
+import ai.djl.translate.TranslateException;
+import ai.djl.util.Progress;
+import java.io.IOException;
+import java.nio.file.Path;
+
+/**
+ * The Penn Treebank (PTB) project selected 2,499 stories from a three year Wall Street Journal
+ * (WSJ) collection of 98,732 stories for syntactic annotation.
+ */
+public class PennTreebank implements RawDataset<Path> {
```
zachgk, line 33: Maybe rename this to PennTreebankText as it is not the full penn treebank due to lacking labels. We don't want to have users come to this later and be confused as to what it is. In that case, it is better to use the name to help clarify what it is, and also probably use documentation to help add additional details to the explanation.

Can you also make this implement TextDataset instead of (or in addition to) RawDataset? The RawDataset was mostly intended as a hack for FastText, which doesn't support training through the standard DJL API. Here, we do want normal DJL models to work with it. The dataset you are working with is one that features unlabeled data organized in sentences where each sentence is on a separate line. So, I recommend having the dataset treat each line as a record in the dataset.

There are different things you do with these kinds of datasets. You could use it to construct word embeddings like [word2vec](https://d2l.djl.ai/chapter_natural-language-processing-pretraining/word2vec.html). Another common use case would be [NLP Pretraining](https://d2l.djl.ai/chapter_natural-language-processing-pretraining/index.html) like you would use for BERT-style transformer models. The goal with this dataset is to make it easy for these use cases.

You will also run into the problem where the DJL [`Record`](https://javadoc.io/doc/ai.djl/api/latest/ai/djl/training/dataset/Record.html) will expect some kind of label for the dataset, but this dataset does not have labels. In that situation, you can use either an empty NDList or null for the label.

zachgk (changes requested):

basicdataset/src/main/java/ai/djl/basicdataset/nlp/PennTreebank.java
AKAGIwyf: OK, thanks, it helps a lot. I’ll change my codes.

basicdataset/src/test/java/ai/djl/basicdataset/PennTreebankTextTest.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.IOException;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class PennTreebankTextTest {
+
+    private static final int EMBEDDING_SIZE = 15;
+
+    @Test
+    public void testPennTreebankTextTrainLocal() throws IOException, TranslateException {
```
zachgk, line 31: I notice you have three tests which are exactly the same besides the usage. Rather than copying and pasting, it is better to loop through the different usages. The technically correct way is to use TestNg parameterized tests, but it is probably easier to just have a `for (Usage usage : new Usage[]{Usage.TRAIN, Usage.VALIDATION, Usage.TEST})` loop.

zachgk (commented):
Thanks for the changes. I made a small change to fix the license. If you haven't done much with licenses, the Apache2 license is typically used for code but there are different licenses used for other kinds of content (books, movies, documents, papers, datasets, artwork, etc). It is important to get the license right and follow the terms because the license describes what you are legally allowed to do with the content.

Besides that, the metadata is uploaded, so you can not modify your tests to use the DJL central repository instead of the local repository.

zachgk (approved):

PR #1583, opened 2022-04-19
[pytorch] Upgrade to pytorch 1.11.0

frankfliu:
## Description ##


engines/pytorch/pytorch-native/src/main/patch/cuda.cmake
```diff
@@ -0,0 +1,509 @@
+# ---[ cuda
```
zachgk, line 1: Can you add a comment at the top indicating the source of the file (including the commit)?

zachgk (approved):

engines/pytorch/pytorch-native/src/main/patch/cuda.cmake
frankfliu: done

lanking520 (approved):

PR #1595, opened 2022-04-22
[basicdataset] Add a Universal Dependencies Corpus for English

LanAtGitHub:
## Description ##
- Add a Gold Standard Universal Dependencies Corpus for English, built over the source material of the English Web Treebank LDC2012T13. 
- cloesed #46


lanking520:
thank you for your contribution, we will take a look

basicdataset/src/test/resources/mlrepo/dataset/nlp/ai/djl/basicdataset/universal-dependencies-en/metadata.json
```diff
@@ -0,0 +1,29 @@
+{
+  "metadataVersion": "0.2",
+  "resourceType": "dataset",
+  "application": "nlp",
+  "groupId": "ai.djl.basicdataset",
+  "artifactId": "universal-dependencies-en",
+  "name": "universal-dependencies-en",
+  "description": "A Gold Standard Universal Dependencies Corpus for English, built over the source material of the English Web Treebank LDC2012T13.",
+  "website": "https://universaldependencies.org/",
```
zachgk, line 9: It is unclear what the website and source of this corpus are. In general, consider that universal dependencies is a collection of POS datasets. Then, there are really two ways to go about building this into DJL:

1. Add one metadata.json file "ai.djl.basicdataset:universal-dependencies" for all of universal dependencies. Then, within it there can be many different artifacts which have a "name" indicating which universal dependencies dataset it is. In this case, the names should correspond to the names on the website, something like "universal-dependencies-English-GUM", "universal-dependencies-English-Atis", or "universal-dependencies-French-GSD". You can then give a "properties" to the individual artifacts as well to search through the datasets or arguments to include information necessary for loading the dataset like where to find the data in the files.
2. Add one metadata file for each of the datasets in universal dependencies. Then, the website, license, name, description, etc should be for that dataset specifically instead of universal dependencies.

Reading your metadata, it seems like you are between the two options a bit. I think either one could be a clear way of handling it, but you would want to pick one and stick with it. If it were me, I think in the current formulation of metadata it would have to be option 2 because different datasets may have different licenses and the metadata only accepts licenses at the metadata level, not the artifact level.

When approaching something like this, it is also good to think about how things will develop in the future as well. One goal is that there might be more Universal Dependencies datasets added too. So, how would they be implemented? Is it possible for them to share a single dataset class? Maybe they might only be able to share an abstract UniveralDependencies class, but each dataset would need it's own class to extend it for non-shared code? Or, it could be that they are so different that trying to share code wouldn't be worth it. For universal dependencies, which do you think it is?

LanAtGitHub: Hello @zachgk, thanks for your reviews and I already know how to build the proper formulation of metadata. However, there is another problem confusing me. I review my current dataset class which extends `TextDataset` class and think it is unable to accurately extract data in Universal Dependencies dataset since it follows a complex format called CoNLL-U. Maybe I should let the dataset class implement `RawDataset`? If I do so, sharing a single dataset class for all Universal Dependencies datasets seems obvious and appropriate.

LanAtGitHub: Hello @zachgk, thanks for your reviews and I already know how to build the proper formulation of metadata. However, I still feel confused for the future development. The official Universal-Dependencies-English-EWT has many fields for different tasks. But currently for pos tagging, I just use a modified version of it which only contains two fields - tokens and corresponding universal pos tags). If I try to deal with complete Universal-Dependencies-English-EWT, maybe I should let the dataset class implement `RawDataset` instead of extending `TextDataset`. If I do so, sharing a single dataset class for all Universal Dependencies datasets seems obvious and appropriate. Or for different task, the way of loading Universal-Dependencies dataset should be different, sharing code is not a good choice in this case.

frankfliu (approved):

zachgk (approved):

PR #1599, opened 2022-04-24
[basicdataset] Fixed out of bound limit

WHALEEYE:
## Description ##

In Stanford Question Answering Dataset the limit is specially handled, and there exists a severe bug that if the limit is too large (or the limit is set as default) it will cause an `IndexOutOfBoundException`.

This PR fixed this bug and changed the test cases into a more strict one so that this bug will be detected.

### DIFF
--- a/basicdataset/src/main/java/ai/djl/basicdataset/nlp/StanfordQuestionAnsweringDataset.java
+++ b/basicdataset/src/main/java/ai/djl/basicdataset/nlp/StanfordQuestionAnsweringDataset.java
@@ -238,18 +238,16 @@ public class StanfordQuestionAnsweringDataset extends TextDataset implements Raw
      * @return the last index of the answer in {@code TargetTextData} that needs to be preprocessed
      */
     private int getLastAnswerIndex(int questionInfoIndex) {
-        boolean endAtHead = false;
-        QuestionInfo questionInfo = questionInfoList.get(questionInfoIndex);
-        while (questionInfo.answerIndexList.isEmpty()) {
-            if (questionInfoIndex == 0) {
-                endAtHead = true;
-                break;
+        // Go backwards through the questionInfoList until it finds one with an answer
+        for (; questionInfoIndex >= 0; questionInfoIndex--) {
+            QuestionInfo questionInfo = questionInfoList.get(questionInfoIndex);
+            if (!questionInfo.answerIndexList.isEmpty()) {
+                return questionInfo.answerIndexList.get(questionInfo.answerIndexList.size() - 1);
             }
-            questionInfo = questionInfoList.get(--questionInfoIndex);
         }
-        return endAtHead
-                ? 0
-                : questionInfo.answerIndexList.get(questionInfo.answerIndexList.size() - 1);
+
+        // Could not find a QuestionInfo with an answer
+        return 0;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


basicdataset/src/main/java/ai/djl/basicdataset/nlp/StanfordQuestionAnsweringDataset.java
```diff
@@ -241,13 +241,19 @@ public Object getData() throws IOException {
...
     protected void preprocess(List<String> newTextData, boolean source) throws EmbeddingException {
         TextData textData = source ? sourceTextData : targetTextData;
-        QuestionInfo questionInfo = questionInfoList.get(Math.toIntExact(this.limit) - 1);
-        int lastIndex =
-                source
-                        ? questionInfo.questionIndex
-                        : questionInfo.answerIndexList.get(questionInfo.answerIndexList.size() - 1);
-        textData.preprocess(
-                manager, newTextData.subList(0, Math.min(lastIndex + 1, newTextData.size())));
+        int limit;
```
frankfliu, line 244: A few comments:
1. avoid use `limit` as local variable, it's confusing because it's clash with `this.limit`
2. to avoid ArrayOutOfBoundException, the following should just work:

```
        int index = Math.min(Math.toIntExact(limit), questionInfoList.size()) -1;
        QuestionInfo questionInfo = questionInfoList.get(index);
```

WHALEEYE: OK, I'll modify this.

frankfliu (approved):

basicdataset/src/main/java/ai/djl/basicdataset/nlp/StanfordQuestionAnsweringDataset.java
```diff
@@ -228,6 +228,30 @@ public Object getData() throws IOException {
...
+            if (questionInfoIndex == 0) {
+                endAtHead = true;
+                break;
+            }
+            questionInfo = questionInfoList.get(--questionInfoIndex);
+        }
+        return endAtHead
+                ? 0
+                : questionInfo.answerIndexList.get(questionInfo.answerIndexList.size() - 1);
+    }
```
zachgk, line 253:
```suggestion
    private int getLastAnswerIndex(int questionInfoIndex) {
        // Go backwards through the questionInfoList until it finds one with an answer
        for (; questionInfoIndex >= 0; questionInfoIndex--) {
            QuestionInfo questionInfo = questionInfoList.get(questionInfoIndex);
            if (!questionInfo.answerIndexList.isEmpty()) {
                return questionInfo.answerIndexList.get(questionInfo.answerIndexList.size() - 1);
            }
        }

        // Could not find a QuestionInfo with an answer
        return 0;
    }
```

I found this function a little difficult to understand, so I wanted to suggest a rewrite that might seem a bit clearer. One thing to pay attention to when writing code which is easy to read is to match what people expect. When we write loops that go through a list, it is almost always a for loop. So, that can fit expectations a bit more than using a while loop, even if the backwards for loop with an existing variable is a bit less common type of for loop. It also simplifies control flow a bit using a `return` to exit from your loop early rather than a `break` and then a conditional after the `break`.

zachgk (approved):

PR #1605, opened 2022-04-26
[MXNet] making boolean as atomic boolean

lanking520:
## Description ##

Boolean field may have issue during the concurrent inference mode on huge traffic. Saying if two thread in the same synchronized block and both cached the boolean value, the updates may not take effect. This may cause a rear case that CachedOp is both initializing and inferencing. Making it atomic to avoid such caching thing from happening.

### DIFF
--- a/engines/mxnet/mxnet-engine/src/main/java/ai/djl/mxnet/engine/MxSymbolBlock.java
+++ b/engines/mxnet/mxnet-engine/src/main/java/ai/djl/mxnet/engine/MxSymbolBlock.java
@@ -36,6 +36,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


frankfliu (approved):

engines/mxnet/mxnet-engine/src/main/java/ai/djl/mxnet/engine/MxSymbolBlock.java
```diff
@@ -184,9 +185,9 @@ protected NDList forwardInternal(
             NDList inputs,
             boolean training,
             PairList<String, Object> params) {
-        if (first) {
+        if (first.get()) {
             synchronized (MxSymbolBlock.class) {
-                if (first) {
+                if (first.get()) {
```
zachgk, line 190: Should this be `first.getAndSet(...)`?

frankfliu: Let close this PR since this doesn't really work.

PR #1608, opened 2022-04-28
[ONNXRuntime] fix naming

lanking520:
## Description ##

Fixed the bug on the name where user has placed name inside.

### DIFF
--- a/engines/onnxruntime/onnxruntime-engine/src/main/java/ai/djl/onnxruntime/engine/OrtSymbolBlock.java
+++ b/engines/onnxruntime/onnxruntime-engine/src/main/java/ai/djl/onnxruntime/engine/OrtSymbolBlock.java
@@ -88,8 +88,16 @@ public class OrtSymbolBlock extends AbstractSymbolBlock implements AutoCloseable
             // If input data has name
             if (inputs.get(0).getName() != null) {
                 for (NDArray input : inputs) {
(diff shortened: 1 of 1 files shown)


engines/onnxruntime/onnxruntime-engine/src/main/java/ai/djl/onnxruntime/engine/OrtSymbolBlock.java
```diff
@@ -85,10 +85,18 @@ protected NDList forwardInternal(
         Map<String, OnnxTensor> container = new ConcurrentHashMap<>();
         // forward
         try (OrtNDManager sub = (OrtNDManager) manager.newSubManager()) {
-            // feed data in to match names
-            for (int i = 0; i < inputNames.size(); ++i) {
-                OrtNDArray ortNDArray = sub.from(inputs.get(i));
-                container.put(inputNames.get(i), ortNDArray.getTensor());
+            // If input data has name
+            if (inputs.get(0).getName() != null) {
```
frankfliu, line 89: We should not assume all input has name

lanking520: User should either provide all names to the NDArray or provides no name

frankfliu (approved):

Issue #248, opened 2020-10-25
Way  for proper Integer array indexing?


HuaMuLanChina:
## Question
In numpy:
```
>>> array = np.ones((5,2))
>>> sel = np.array([1,2,4])
>>> array[sel].shape
(3, 2)
```
what i've tried:
```
NDArray array = manager.ones(new Shape(5,2));
long[] d = {1,2,4};
NDArray sel = manager.create(d);
array.get(sel)//error
```

lanking520:
There is a way, try to do following the javadoc: https://javadoc.io/doc/ai.djl/api/latest/ai/djl/ndarray/index/NDIndex.html

```java
NDIndex index = new NDIndex().addIndices(0) // pick first dim
                                                      .addPickDim(sel);
array.get(sel);
```

HuaMuLanChina:
```
import ai.djl.ndarray.NDArray;
import ai.djl.ndarray.NDManager;
import ai.djl.ndarray.types.Shape;
import ai.djl.ndarray.index.NDIndex;

NDManager manager = NDManager.newBaseManager();
NDArray array = manager.arange(10).reshape(-1,2);
long[] d = {1,2,4};
NDArray sel = manager.create(d);
NDIndex index = new NDIndex().addIndices(0).addPickDim(sel);
array.get(index);
```
can't pass block runner.

```
jdk.jshell.EvalException: get() currently supports all, fixed, and slices indices
	at ai.djl.ndarray.index.NDArrayIndexer.get(NDArrayIndexer.java:75)
	at ai.djl.ndarray.NDArray.get(NDArray.java:498)
	at .(#10:1)
```

zachgk:
We don't support this right now.

For clarification, there are two types of indexing with numeric indices: [pick](https://mxnet.apache.org/versions/1.7.0/api/python/docs/api/symbol/symbol.html#mxnet.symbol.pick) and [gather](https://mxnet.apache.org/versions/1.7.0/api/python/docs/api/symbol/symbol.html#mxnet.symbol.gather_nd). We have support for pick, but need to add gather to support this.

ManuLasker:
@HuaMuLanChina Hello, what you can do and it functions for me with the pytorch engine is:
```
NDManager manager = NDManager.newBaseManager();
NDArray array = manager.arange(10).reshape(-1,2);
long[] d = {1,2,4};
NDArray sel = manager.create(d).repeat(2).reshape(-1, 2);
[...]

PR #1622, opened 2022-05-03
Support gather of pytorch

KexinFeng:
## Description ##

This originiates from issue https://github.com/deepjavalibrary/djl/issues/248. This is regarding support of advanced indexing.

As part of this support, this PR implements `gather`. It's the same as the gather in
https://pytorch.org/docs/1.11/generated/torch.gather.html#torch.gather 
https://numpy.org/doc/stable/reference/generated/numpy.take_along_axis.html
It currently supports pytorch engine only. MXNet and TensorFlow have different defination of gather.

The design is the following:
`gather(arr, idx, axis)` returns the following based on the conventions [1](https://pytorch.org/docs/1.11/generated/torch.gather.html#torch.gather) [2](https://numpy.org/doc/stable/reference/generated/numpy.take_along_axis.html
):
```
out_{ijk} = arr_{idx_{ijk}, j, k} if axis=0
              or arr_{i, idx_{ijk}, k} if axis=1
              or arr_{i, j, idx_{ijk}}  if axis=2
```
But `gather` is used to index across axis, including the preceding axis. This *conflicts* with the linear storage of `List<NDIndexElement>` inside `NDIndex`. So `gather` is treated parallel to `NDIndex` when fed into `get`.

Similar functions in numpy / pytorch / mxnet can also be implemented: e.g. `torch.take`, and advanced indexing like `arr[:, [0 ,2]]` and `arr[idx_r, idx_c]`


api/src/main/java/ai/djl/ndarray/index/NDIndexFullGather.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
frankfliu, line 2:
```suggestion
 * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

api/src/main/java/ai/djl/ndarray/index/NDIndexFullGather.java
```diff
@@ -0,0 +1,51 @@
...
+import ai.djl.ndarray.NDArray;
+
+/** A simplified representation of a gather-based {@link NDIndex}-like class. */
+public final class NDIndexFullGather {
+
+    private NDArray indices;
+    private int axis;
+
+    /**
+     * Constructs a new {@link NDIndexFullGather}.
```
frankfliu, line 24:
```suggestion
     * Constructs a new {@code NDIndexFullGather} instance.
```

KexinFeng: Got it. Thanks!

api/src/main/java/ai/djl/ndarray/index/NDIndexFullGather.java
KexinFeng: Got it. Thanks!

integration/src/main/java/ai/djl/integration/tests/ndarray/NDIndexTest.java
```diff
@@ -52,6 +54,23 @@ public void testPick() {
         }
     }
 
+    @Test
+    public void testGather() {
+        // Currently in windows gradle cannot find all the engines.
+        // In the dependency, changing runtimeOnly to api however will remedy the problem.
+        // TODO: remove this when gradle problem is fixed.
+        TestRequirements.notWindows();
+        Engine engine = Engine.getEngine("PyTorch");
```
zachgk, line 63: The integration module is supposed to be for "engine agnostic tests". So, you shouldn't need to specify "PyTorch" here. For the other engines, it will throw a `UnsupportedOperationException` instead of running successfully. But, our test runner is designed to treat `UnsupportedOperationException` as a special test result, not a failure. It's similar to how a testng `SkipException` counts as a skipped test result rather than a test failure.

In the future, we should be able to implement gather for the other engines and then it will just run the test because it is already there.

zachgk (approved):

integration/src/main/java/ai/djl/integration/tests/ndarray/NDIndexTest.java
```diff
@@ -52,6 +53,22 @@ public void testPick() {
         }
     }
 
+    @Test
+    public void testGather() {
+        // Currently in windows gradle cannot find all the engines.
```
frankfliu, line 58: Is this comment still valid?

zachgk: Probably. It seems like something is wrong with `runtimeOnly` dependencies on windows gradle.

KexinFeng: Yes after updating gradle, on GitHub the test stills fails for the same reason

PR #1627, opened 2022-05-06
Support of take from pytorch

KexinFeng:
## Description ##

This is a follow-up on PR https://github.com/deepjavalibrary/djl/pull/1622. The `take` feature will enable more advanced indexing.

It's the same as `take` in [numpy ](https://numpy.org/doc/stable/reference/generated/numpy.take.html#numpy.take) and [ pytorch](https://pytorch.org/docs/1.11/generated/torch.take.html?highlight=take#torch.take).

Defination of `NDArray take(NDArray index)`:
Returns a partial pointed by the index array, according to row-major-ordering linear indexing.
The shape of the output is decided by that of `index`.

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/NDArray.java
+++ b/api/src/main/java/ai/djl/ndarray/NDArray.java
@@ -514,6 +514,21 @@ public interface NDArray extends NDResource, BytesSupplier {
     /**
      * Returns a partial {@code NDArray}.
      *
+     * @param index the boolean or int {@code NDArray} that indicates what to get
+     * @return the partial {@code NDArray}
+     */
+    default NDArray get(NDArray index) {
+        DataType indexType = index.getDataType();
+        if (indexType == DataType.BOOLEAN) {
+            return get(new NDIndex().addBooleanIndex(index));
+        } else {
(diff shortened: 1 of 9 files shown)


lanking520 (approved):

api/src/main/java/ai/djl/ndarray/NDArray.java
```diff
@@ -511,6 +511,20 @@ default NDArray get(NDIndex index) {
...
+     * Returns a partial {@code NDArray}.
+     *
+     * @param index the boolean or int {@code NDArray} that indicates what to get
+     * @return the partial {@code NDArray}
+     */
+    default NDArray get(NDArray index) {
+        if (index.getDataType() == DataType.BOOLEAN) {
+            return get(new NDIndex().addBooleanIndex(index));
+        } else {
+            return take(index);
```
zachgk, line 524: Probably in the next PR, change this to `return get(new NDIndex().addTakeIndex(index))` or combine both branches into a single path `return get(new NDIndex().addArrayIndex(index))` and handle it on the NDIndexing side.

zachgk (approved):

PR #1631, opened 2022-05-07
new LinearCollection block type

patins1:
This change enables the execution of multiple linear transformations using only one block as opposed to using multiple Linear blocks in a more complex setup.
I measured a speed-up of at least 10 for my scenario as the linear transformations are handled within one ndarray.


lanking520:
@patins1 Thanks for your contribution, can you try to rerun the test?

patins1:
> @patins1 Thanks for your contribution, can you try to rerun the test?

Updated my fork branch with latest master, and the test run green now :)

Issue #1563, opened 2022-04-12
how using djl to call yolov5 ?


javadoge:
I have some problems when using djl to call yolov5, it can't be recognized normally

my code :

`

        int imageSize = 640;
        Pipeline pipeline = new Pipeline();
        pipeline.add(new Resize(imageSize));
        pipeline.add(new ToTensor()); 

        Translator<Image, DetectedObjects> translator =  YoloV5Translator
                .builder()
                .setPipeline(pipeline)
                .optSynset(Arrays.asList("person"))
                .optThreshold(0.8f)
                .build();
        
        Criteria<Image, DetectedObjects> criteria = Criteria.builder()
                .setTypes(Image.class, DetectedObjects.class)
                .optModelUrls("build/yolov5/")
                .optModelName("best.torchscript.pt")
                .optTranslator(translator)
                .optProgress(new ProgressBar())
                .optEngine("PyTorch")
                .build();

`

![detected-DSC00020_0](https://user-images.githubusercontent.com/5305836/162902286-5a533ccc-ee0f-4b86-b431-ea418cdce27a.png)

but， when  i use 

 `  python detect.py --source xxx.png --weights ./best.torchscript.pt `  

it can work.

what problem when using djl to call yolov5, need help , thanks ~

frankfliu:
@javadoge 

Each model may have sightly different post processing. The default DJL YoloV5Translator matches MXNet gluonCV's implementation. You might need to make some changes to match your model.

Can you shared your model and python script?

javadoge:
> @javadoge
> 
> Each model may have sightly different post processing. The default DJL YoloV5Translator matches MXNet gluonCV's implementation. You might need to make some changes to match your model.
> 
> Can you shared your model and python script?

thanks!

python script : python detect.py --source 444.png --weights ./best.torchscript

result :

![1](https://user-images.githubusercontent.com/5305836/163078192-f1cbfa57-7eda-4821-9d1b-089c14ce5321.jpg)
![2](https://user-images.githubusercontent.com/5305836/163078214-f26440f5-5998-4801-9700-aee4be29cb96.png)


model: 

 
[Uploading best.zip…]()

lanking520:
@javadoge  I got your problem solved. Root cause is the synset. Yolo has 80 classes, where in your input, you only put a person class. I just make a workaround with 80 copies.

## Code that can make this work

```
public class Yolo {


[...]

PR #1651, opened 2022-05-13
add width/height conversion to ObjectDetection

lanking520:
## Description ##

Provide a new function for Width/Height conversion for Bounding box. 

https://github.com/deepjavalibrary/djl/issues/1563


api/src/main/java/ai/djl/modality/cv/translator/ObjectDetectionTranslator.java
```diff
@@ -60,6 +62,7 @@ public void prepare(TranslatorContext ctx) throws Exception {
         protected float threshold = 0.2f;
         protected double imageWidth;
         protected double imageHeight;
+        protected boolean applyRatio = false;
```
frankfliu, line 65:
```suggestion
        protected boolean applyRatio;
```

frankfliu (approved):

PR #1652, opened 2022-05-14
Create QuantileLoss.java

gforman44:
QuantileLoss for training toward a particular quantile, e.g. esimating close to the P75.

## Description ##

Sometimes you want to train your estimator for a particular quantile, e.g. the 75th percentile (P75).   This provides [weighted] QuantileLoss, e.g. 

[https://aws.amazon.com/blogs/machine-learning/measuring-forecast-model-accuracy-to-optimize-your-business-objectives-with-amazon-forecast/](https://aws.amazon.com/blogs/machine-learning/measuring-forecast-model-accuracy-to-optimize-your-business-objectives-with-amazon-forecast/)


api/src/main/java/ai/djl/training/loss/QuantileLoss.java
```diff
@@ -0,0 +1,43 @@
...
+    public NDArray evaluate(NDList labels, NDList predictions) {
+        NDArray pred = predictions.singletonOrThrow();
+        NDArray target = labels.singletonOrThrow().reshape(pred.getShape());
+        NDArray loss = pred.sub(target).mul(target.lte(pred).toType(DataType.FLOAT32, false).sub(quantile)).abs().mul(2);
+        if (weighted) {
+            loss = loss.div(target.sum());
+        } else {
+            loss = loss.div(target.size());
+        }
+        return loss;
```
zachgk, line 41: Can you add a test to https://github.com/deepjavalibrary/djl/blob/master/integration/src/main/java/ai/djl/integration/tests/training/LossTest.java or https://github.com/deepjavalibrary/djl/blob/master/integration/src/main/java/ai/djl/integration/tests/training/EvaluatorTest.java?

zachgk (dismissed):
Thanks for adding this @gforman44! I added a few comments

gforman44:
It is a Loss, not just an Evaluator.

I put  this reference in the java doc for further explanation.
https://bibinmjose.github.io/2021/03/08/errorblog.html

Yes, it deserves a test case.  Harder with non-deterministic learning.

Haven’t had time.

George

zachgk:
I'm not sure what you mean by it's harder with non-deterministic learning. We don't really require a full learning test case. The way we currently test the losses is just a unit test. Essentially, it is just passing it a sample input and then verify that with a hard-coded expected result number. The purpose of this is mostly to make sure that the loss runs and it doesn't change accidentally. You can find examples of what I am referring to [with the other loss tests](https://github.com/deepjavalibrary/djl/blob/master/integration/src/main/java/ai/djl/integration/tests/training/LossTest.java).

gforman44:
One more thing before the framework gets too ossified:

consider making Loss an *interface* that is a sub-interface of Evaluator
*interface*. Then any Loss can be used as an additional Evaluation.  And a
base class for Evaluator can be used for either easily.  My base class maps
to double[] so that subclasses can have several accumulators per test
situation, rather than your current Loss base class which maps to Double.

zachgk:
That seems like a reasonable suggestion (making Loss and Evaluator interfaces for more flexibility). With something like QuantileLoss, I can see how you may want to track multiple QuantileLosses. Or, with something like TopK accuracy you could have multiple values of k.

If you have time, do you want to try contributing those changes yourself?

KexinFeng (approved):

frankfliu (approved):

gforman44:
It is a Loss, not just an Evaluator.**

I felt some of the questions were already addressed in the java doc, such
as when you want  to train for the P90 of the distribution.

I had put  this reference in the java doc for further explanation:
https://bibinmjose.github.io/2021/03/08/errorblog.html
Would you like more?
https://docs.aws.amazon.com/forecast/latest/dg/API_WeightedQuantileLoss.html
Here’s one comparing different metrics including Quantile Loss:
https://docs.aws.amazon.com/forecast/latest/dg/metrics.html



Yes, you are right,  it deserves a test case.  Harder with
non-deterministic learning.

Haven’t had time.

George


** I  think Loss should be a subclass of Evaluator.

PR #1653, opened 2022-05-14
Create Coverage.java

gforman44:
## Description ##

For regression problems, you'd like to know how often you're over-estimating the target.   Sometimes you want to overestimate, say, 75% of the time, e.g. to make sure you have enough milk in stock to satisfy your customers.

- Note that this code could be generalized to reshape the predictions to match the labels.  It re-uses the AbstractAccuracy class, which requires an axis number in its constructor, but this is ignored here, and generally for regression problems that don't use 1-hot encoding.

### DIFF
--- a/api/src/main/java/ai/djl/training/evaluator/Coverage.java
+++ b/api/src/main/java/ai/djl/training/evaluator/Coverage.java
@@ -4,15 +4,17 @@ import ai.djl.ndarray.NDArray;
 import ai.djl.ndarray.NDList;
 import ai.djl.util.Pair;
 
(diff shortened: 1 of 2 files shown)


zachgk (dismissed):
This looks really good. Can you add a test for it to https://github.com/deepjavalibrary/djl/blob/master/integration/src/main/java/ai/djl/integration/tests/training/EvaluatorTest.java to make sure that it works now and isn't broken by accident later

frankfliu (approved):

KexinFeng (approved):

PR #1656, opened 2022-05-14
[pytorch] Add macOS M1 chip support.

frankfliu:
Change-Id: I5fb7f2bc1ba07626fb8dcc45e1daf77d4b31765b

## Description ##


lanking520:
@frankfliu where do you find the binary and have you include this into our build pipeline?

frankfliu:
I got the binary from pip wheel. There is no M1 machine we can use for CI, we have to manually build the JNI.

lanking520:
@frankfliu do we need special build instruction for M1 chip machine? Or it is the same for us to build on Intel chip?

lanking520 (approved):

PR #1661, opened 2022-05-18
Fixes bug in NDArray.oneHot() API

pdradx:
The on and off value parameters were misplaced.

## Description ##


frankfliu:
@pdradx 
Thanks a lot for your contribution. Without contributions like yours it'd be impossible to keep DJL going with the high standards.

lanking520 (approved):

PR #1666, opened 2022-05-20
[TabNet/WIP] Implement GhostBatchNorm

DiaaAj:
## Description ##

`GhostBatchNormalization` is one of the components that build up TabNet. It works similar to `BatchNorm` except that it splits each batch into smaller mini-batches and each of them is normalized separately.

Changes:
* **BatchNorm**: remove the final notation from the builder class as GhostBatchNormalization extends BatchNorm and the builder is required to be extended with it for usage.
*  **GhostBatchNorm**: Extends BatchNorm and overrides `internalForward()` such that it splits the batch into mini-batches using `StackBatchifier` before applying batch normalization. `split()` divides the size of the original batch and ceil it so if the original batch size isn't divisible by the virtual batch size, the remaining samples will be placed in the last mini-batch (this behaviour is inspired from PyTorch chunk). Builder methods are duplicated from BatchNorm due to return type, this can be resolved using generics if needed.
* **GhostBatchNormTest**: tests the splitting behavior in GhostBatchNorm. I will add more for `forwardInternal` in later commits.

### DIFF
--- a/api/src/main/java/ai/djl/nn/norm/BatchNorm.java
+++ b/api/src/main/java/ai/djl/nn/norm/BatchNorm.java
@@ -10,6 +10,7 @@
  * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
  * and limitations under the License.
  */
+
 package ai.djl.nn.norm;
 
 import ai.djl.Device;
@@ -282,34 +283,22 @@ public class BatchNorm extends AbstractBlock {
 
     /** The Builder to construct a {@link BatchNorm}. */
     public static class Builder extends BaseBuilder<Builder> {
-        /**
-         * Builds a {@link BatchNorm} block.
-         *
-         * @return the {@link BatchNorm} block
-         */
         Builder() {}
 
-        /**
-         * Builds the new {@link BatchNorm}.
-         *
-         * @return the new {@link BatchNorm}
-         */
+        /** {@inheritDoc} */
         @Override
         public BatchNorm build() {
             return new BatchNorm(this);
         }
 
-        /**
-         * Returns this {code Builder} object.
-         *
-         * @return this {@code BaseBuilder}
-         */
+        /** {@inheritDoc} */
         @Override
         public Builder self() {
             return this;
         }
     }
 
+    /** The Builder to construct a {@link BatchNorm} type of {@link ai.djl.nn.Block}. */
     public abstract static class BaseBuilder<T extends BaseBuilder<T>> {
 
         protected int axis = 1;
@@ -318,7 +307,7 @@ public class BatchNorm extends AbstractBlock {
         protected boolean center = true;
         protected boolean scale = true;
 
-        public BaseBuilder() {}
+        protected BaseBuilder() {}
 
         /**
          * Set the axis in which channel is specified. Defaults to 1.
@@ -375,10 +364,18 @@ public class BatchNorm extends AbstractBlock {
             return self();
         }
 
-        /** {@inheritDoc} */
+        /**
+         * Builds the new {@link BatchNorm}.
+         *
+         * @return the new {@link BatchNorm}
+         */
         public abstract BatchNorm build();
 
-        /** {@inheritDoc} */
+        /**
+         * Returns this {code Builder} object.
+         *
+         * @return this {@code BaseBuilder}
+         */
         public abstract T self();
     }
 }
--- a/api/src/main/java/ai/djl/nn/norm/BatchNorm.java
+++ b/api/src/main/java/ai/djl/nn/norm/BatchNorm.java
@@ -77,7 +77,7 @@ public class BatchNorm extends AbstractBlock {
     private Parameter runningMean;
     private Parameter runningVar;
 
-    BatchNorm(Builder builder) {
+    BatchNorm(BaseBuilder<?> builder) {
         super(VERSION);
         axis = builder.axis;
         epsilon = builder.epsilon;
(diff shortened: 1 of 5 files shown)


api/src/main/java/ai/djl/nn/norm/BatchNorm.java
```diff
@@ -281,7 +281,7 @@ public static Builder builder() {
     }
 
     /** The Builder to construct a {@link BatchNorm}. */
-    public static final class Builder {
+    public static class Builder {
```
zachgk, line 284: What we typically do for this is create two builders: `BatchNorm.BaseBuilder` and `BatchNorm.Builder extends BatchNorm.BaseBuilder`. Then, you also have the `GhostBatchNorm.Builder extends BatchNorm.BaseBuilder`. If you do this, you won't need to override all those methods in `GhostBatchNorm.Builder` with the super.

Here are some examples. Refer to `AbstractImageFolder.ImageFolderBuilder` for an abstract or base builder. Refer to `ImageFolder.Builder` as the final builder.

There are a few changes to make. The setters in the base builders should return type `T` instead of type `Builder`. The `self` and `build` methods should go on the final builders.

api/src/test/java/ai/djl/nn/norm/GhostBatchNormTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.testng.annotations.Test;
+
+public class GhostBatchNormTest {
+
+    private GhostBatchNorm gbn;
+    private NDManager manager;
+
+    @BeforeTest
+    void initialize() {
+        this.manager = NDManager.newBaseManager();
```
zachgk, line 18: The primary purpose of an NDManager is to manage memory. So, you need to close the manager when you are done using it. You can either do this by adding an `@AfterTest` to contain the close statement or by using a try-with-resources:

```java
try(NDManager manager = NDManager.newBaseManager()) {
  // Do stuff with manager here
}
// manager is automatically closed when exiting the try block
```

api/src/main/java/ai/djl/nn/norm/GhostBatchNorm.java
```diff
@@ -0,0 +1,121 @@
...
+ * smaller sub-batches aka <em>ghost batches</em>, and normalize them individually to have a mean of
+ * 0 and variance of 1 and finally concatenate them again to a single batch. Each of the
+ * mini-batches contains a virtualBatchSize samples.
+ */
+public class GhostBatchNorm extends BatchNorm {
+
+    private int virtualBatchSize;
+    private Batchifier batchifier;
+
+    GhostBatchNorm(Builder builder) {
```
zachgk, line 21: Can you make the constructor protected? The way we typically build it is by doing `GhostBatchNorm.builder().optVirtualBatchSize(100).build()` so it is cleaner to only allow using this constructor through the builder

api/src/main/java/ai/djl/nn/norm/GhostBatchNorm.java
```diff
@@ -0,0 +1,121 @@
...
+        super(new BatchNorm.Builder());
```
zachgk, line 22:
```suggestion
        super(builder);
```

Otherwise, any setting you make for BatchNorm such as `optAxis()` or `optScale()` would not be set properly

api/src/main/java/ai/djl/nn/norm/GhostBatchNorm.java
```diff
@@ -0,0 +1,121 @@
...
+import ai.djl.training.ParameterStore;
+import ai.djl.translate.Batchifier;
+import ai.djl.translate.StackBatchifier;
+import ai.djl.util.PairList;
+
+/**
+ * {@link GhostBatchNorm} is similar to {@link BatchNorm} except that it splits a batch into a
+ * smaller sub-batches aka <em>ghost batches</em>, and normalize them individually to have a mean of
+ * 0 and variance of 1 and finally concatenate them again to a single batch. Each of the
+ * mini-batches contains a virtualBatchSize samples.
```
zachgk, line 14: Can you add a link to the source (like the paper) that GhostBatchNorm comes from? It would help people who come across this class and want to better understand it. Then, they can more easily get to the paper and read all the additional information it has

zachgk (changes requested):
Good work @DiaaAj! I requested a few changes on the PR, but most of it is just unfamiliarity with DJL or DJL conventions. Otherwise, it looks really good

zachgk (commented):
Thanks for making the changes. It looks good now

> GhostBatchNormTest: tests the splitting behavior in GhostBatchNorm. I will add more for forwardInternal in later commits.

Were you still planning to add a test for GhostBatchNorm forward? Or, is there anything else you wanted to do before moving this PR out of draft?

DiaaAj:
Scanner Javadoc styling errors should be resolved now. For the forward test, I tried to write a unit test for it but getting  

> the mode of batchNorm in MXNet should align with the mode of GradientCollector

I'm trying to find a workaround for this error. We can either move this PR out of draft and add the unit test in a follow-up PR or keep it in draft until the test is done.

api/src/main/java/ai/djl/nn/norm/GhostBatchNorm.java
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
frankfliu, line 2:
```suggestion
 * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

DiaaAj:
I have created a test for `GhostBatchNorm` forward and made the following 2 changes to resolve 2 bugs

1. Created 2 new methods (`batchify` and `squeezeExtraDimensions`) inside `GhostBatchNorm` to remove first dimension in the batchified `NDList` as `StackBatchifier` turned out to add and extra dimension at the beginning e.g (2,1,5) -> (1,2,1,5)
2. Fix sub-batches normalization loop as it didn't perform the operation in-place

I have completed changes from my end in this PR. If you don't have any other comments feel free to move it out of PR

DiaaAj:
As a side note, the expected values in the test came from the python implementation of ghost batch normalization

zachgk (approved):
It think it looks ready to merge. Great work @DiaaAj

PR #1671, opened 2022-05-23
[training] Add Cyclical Tracker

### LINKED ISSUE #37: Dynamic learning parameter - time period
## Description
Dynamically modified learning parameter over time. We want learning parameter to change dynamically when network is learning:
- constant learning parameter at the start then **cycle** mode. We want to start with constant learning parameter, then after N epoch start **cycle** - like saw, from min. to max. definition of learning parameter (up and down)
- it should be configurable - we want to start with static learning parameter and then **cycle**, we want to start with **cycle** at the beginning,.. 
- any combination with constant and **cycle**, for example 0-10 epoch constant learning parameter, then **cycle**(whatever type) (11-20), then again constant parameter (21-30),..

**Cycle** (learning parameter alternation) mode should have multiple types, with configurable time interval and min., max. learning parameter value, for example:
- like saw wave (up and down /\ /\ /\ )
- like first half of saw wave (just up / / / )
- like second half of saw wave (just down \ \ \ )
- we can go any combination (/\ - / / - \ \ )

### Will this change the current api? How?
Yes, implement learning parameter which dynamically changes over iteration/epoch

### Who will benefit from this feature?
Everybody, this will enable network to learn more quickly.

## References
Here is dynamic learning parameter change implementation, kind of saw wave over time:
https://github.com/eclipse/deeplearning4j/blob/b5f0ec072f3fd0da566e32f82c0e43ca36553f39/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/schedule/CycleSchedule.java
We want to extend this implementation - see above.

dandansamax:
## Description ##

Add Cyclical Tracker according to cyclical learning rate policy, as detailed in the paper [Cyclical Learning Rates for Training Neural Networks](https://arxiv.org/abs/1506.01186).

close #37.

### DIFF
--- a/api/src/main/java/ai/djl/training/tracker/CyclicalTracker.java
+++ b/api/src/main/java/ai/djl/training/tracker/CyclicalTracker.java
@@ -1,0 +1,280 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package ai.djl.training.tracker;
+
+import ai.djl.util.Preconditions;
+
+/**
+ * {@code CyclicalTracker} is an implementation of {@link Tracker} which is a policy of learning
+ * rate adjustment that increases the learning rate off a base value in a cyclical nature, as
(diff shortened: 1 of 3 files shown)


api/src/main/java/ai/djl/training/tracker/CyclicalTracker.java
```diff
@@ -0,0 +1,280 @@
...
+    private int stepSizeDown;
+    private int totalSize;
+    private float stepRatio;
+    private ScaleFunction scaleFunction;
+    private boolean scaleModeCycle;
+
+    /**
+     * Creates a new instance of {@code CosineTracker}.
+     *
+     * @param builder the builder to create a new instance of {@code CosineTracker}
```
zachgk, line 37:
```suggestion
     * Creates a new instance of {@code CyclicalTracker}.
     *
     * @param builder the builder to create a new instance of {@code CyclicalTracker}
```

zachgk (approved):

frankfliu (approved):

PR #1683, opened 2022-05-29
Added "capped" state to NDManager

patins1:
This change adds the capability to "cap" a NDManager instance.
In the capped state, it is not possible to add new resources to the manager (except for submanagers).
The model developer would choose to do so when he is sure that all resources (usually NDArray) that are supposed to be attached are attached and he expects no new resources to be added in the future, thus preventing unintentional addition of resources that could go undetected otherwise and could pose memory leaks.
Sub managers are by design still allowed to be attached and have an independent "cap" status.

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
+++ b/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
@@ -47,6 +47,7 @@ public abstract class BaseNDManager implements NDManager {
     protected ConcurrentHashMap<String, AutoCloseable> resources;
     protected ConcurrentHashMap<String, TempResource> tempResources;
     protected AtomicBoolean closed = new AtomicBoolean(false);
+    protected AtomicBoolean capped = new AtomicBoolean(false);
 
     protected BaseNDManager(NDManager parent, Device device) {
         this.parent = parent;
@@ -237,6 +238,12 @@ public abstract class BaseNDManager implements NDManager {
 
     /** {@inheritDoc} */
     @Override
+    public void cap() {
+        this.capped.set(true);
+    }
+
+    /** {@inheritDoc} */
+    @Override
     public NDManager getParentManager() {
         return parent;
     }
@@ -270,6 +277,15 @@ public abstract class BaseNDManager implements NDManager {
     /** {@inheritDoc} */
     @Override
     public synchronized void attachInternal(String resourceId, AutoCloseable resource) {
+        if (capped.get()) {
+            throw new IllegalStateException("NDManager is capped for addition of resources.");
+        }
+        attachUncappedInternal(resourceId, resource);
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public synchronized void attachUncappedInternal(String resourceId, AutoCloseable resource) {
         if (closed.get()) {
             throw new IllegalStateException("NDManager has been closed already.");
         }
@@ -452,7 +468,7 @@ public abstract class BaseNDManager implements NDManager {
             try {
                 if (!detached) {
                     if (manager.isOpen()) {
-                        resource.attach(manager);
+                        resource.returnResource(manager);
                     } else {
                         resource.close();
                     }
--- a/api/src/main/java/ai/djl/ndarray/NDManager.java
+++ b/api/src/main/java/ai/djl/ndarray/NDManager.java
@@ -1390,6 +1390,13 @@ public interface NDManager extends AutoCloseable {
     boolean isOpen();
 
     /**
+     * Caps this manager to prevent unintentional attachment of resources. This is useful to detect
+     * memory leaks at an early point in time. The attachment of sub managers is still allowed after
+     * this method has been called.
+     */
+    void cap();
+
+    /**
      * Returns the parent {@code NDManager}.
      *
      * @return the parent {@code NDManager}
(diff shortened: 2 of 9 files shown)


api/src/main/java/ai/djl/ndarray/BaseNDManager.java
```diff
@@ -270,6 +277,13 @@ public String toString() {
     /** {@inheritDoc} */
     @Override
     public synchronized void attachInternal(String resourceId, AutoCloseable resource) {
+        if (capped.get()) {
+            throw new IllegalStateException("NDManager is capped for addition of resources.");
+        }
+        attachInternalUncapped(resourceId, resource);
```
zachgk, line 283: Is creating the additional method of `attachInternalUncapped` just for calling it with managers? It might be better to keep `attachInternal` and `attachInternalUncapped` combined and just add an additional check like `if (capped.get() && !(resource instanceof NDManager))`. Also, can you add a test for capping?

api/src/main/java/ai/djl/ndarray/NDManager.java
```diff
@@ -1389,6 +1389,13 @@ default NDArray truncatedNormal(
      */
     boolean isOpen();
 
+    /**
+     * Caps this manager to prevent unintentional attachment of resources. This is useful to detect
+     * memory leaks at an early point in time. The attachment of sub managers is still allowed after
+     * this method has been called.
+     */
+    void cap();
```
zachgk, line 1397: @frankfliu What are your thoughts on this?

zachgk (changes requested):

patins1:
I see two use cases for `NDManager.cap()`, once I call it after i set up the features/labels with a toplevel manager (a), and once i call it on the model's manager (b) just before I do the training (so after the model's parameters are created).  I would even argue that it would be useful that EasyTrain.fit() - as it is a standardized training loop - should call `NDManager.cap()` for both managers just at the start because EasyTrain.fit() is not supposed to attach any new resources to any of those two managers.  

My current pull request works only for PyTorch, so I was experimenting with MXNet as well and found one issue for both engines: the indexing operation used to assemble batches from the input features currently attaches the extracted data first to the manager (a) before being attached to the trainer's manager. This slows down the batching algorithm by some factor, I want to create a new pull request just addressing this issue. Also, I found some bugs in there that temporary attached resources are actually returned to a different original manager, I am gonna write a test to verify.
  
> @patins1 This is a great initiative. However, I don't think this easy to use. The biggest challenge is when to call `NDManager.cap()` and which `NDManager` instance is causing leak.

frankfliu:
@patins1 
The leak can happen at different levels, Model level, ParameterStore level or Predictor level. Set `capped` won't detect leak at subNDManager. To detect memory leak, figure out which NDManager causing leaking is more important.

The solution I proposed is to set a global watermark which applies to all NDManagers, if any NDManager breach the threshold, we can through exception and call `BaseNDManager.debugDump()` to print out the entire resource tree. And we can quickly identify where the leak happen.

api/src/main/java/ai/djl/ndarray/BaseNDManager.java
patins1: With  the return-resources fix in the other pull request, resource can also be an NDArray, therefore that logic

patins1:
I fully understand the global watermark solution, I see only the problem in setting a watermark not too high so that memory leaks can go unnoticed and not too low that it actually prevents learning when e.g. the number of features increases. Therefore I prefer to use an analytical approach rather than a statistical/holistic approach. I have two other ideas that I would appreciate your feedback on:

1.  `BaseNDManager.resources` is typed as `ConcurrentHashMap`. How about using something similar to `WeakHashMap` (however, we need the values to be week and not the keys). Then the resources would be freed automatically when they are not anymore referenced (other than from `BaseNDManager.resources`)
2. In `EasyTrain.fit()`'s outer for-loop, we could count the total number of allocated `NDArray` after the first iteration, and after any subsequent iteration of the loop we assert that this number has not increased. This approach would share one feature with the watermark solution: determine the total number of `NDArray` instances. Maybe this can be solved by also storing a reference from parent managers to their children in order to recursively calculate the number.

frankfliu:
@patins1 

I assume the watermark is only enabled for debugging/troubleshooting purpose. And usually for a long running case. It's should be that hard to figure out the watermark value.

We used to use `WeakHashMap` or `SoftHashMap` in the past, but the Engine NDArray is operated in async way, Both PyTorch and MXNet `close()` is not thread-safe. We must ensure nobody still reference to the NDArray when it's closing.
With `WeakHashMap`, the NDArray might be closed in GC thread, while engine operator thread is still using the memory, and that will cause segfault.

patins1:
It is easier to think of a memory allocation limit, e.g. 10 gigabyte , e.g. I experienced an allocation over 60 gigabyte last year by native libraries (by the amount of reduced disk space on my hard drive..), so i started thinking how to solve this kind of problem. So it was not really natural to me to restrict the number of resources but more the size of memory they allocate in the native memory space which is obviously swapped to disk. The "cap" functionality seems easier to me than calculating that allocated memory which I wouldn't know how to do. My further thinking was that there are ndmanagers in the responsibility of the model developer (ie. me) and he can determine to cap or not to cap depending on the application situation, and then there are ndmanagers created internally e.g. by the trainer or the batch iterator where the model developer can trust that they have no leaks so capping doesn't need to be applied there. 

Using a watermark is more reliable to detect any possible leak, but it doesnt show you where the leak happens, and also resource-counting seems not a good match to measure memory usage, therefore "watermark " and  "cap" cover two different functionalities that cannot be  exchanged for one another, so i dont see why we could add both to DJL library.

frankfliu:
@patins1 

I agree we can have both. I just feel it's hard for developer to understand when to use this feature.

frankfliu (approved):

PR #1688, opened 2022-06-01
allow indexer to attach specific manager

patins1:
This feature allows to pass an NDManager to NDArray indexation operation so that the result is attached to the passed NDManager instead of to the original NDManager of the NDArray.
Thus, ArrayDataset.get() does not need to attach the result of the indexation a second time, thus batch creation is speeded up.

In my performance tests, I got following improvement:

Before:
It took 6872ms totally
It took 6876ms totally
It took 6825ms totally
It took 6923ms totally

After:
It took 6734ms totally
It took 6709ms totally
It took 6728ms totally
It took 6703ms totally

So the training time is reduced by 2.2% !

Also, this pull request fixes a bug that temporary resources are returned to the wrong manager.


engines/pytorch/pytorch-engine/src/test/java/ai/djl/pytorch/integration/ManagerAttachmentTest.java
```diff
@@ -0,0 +1,52 @@
...
+import ai.djl.ndarray.NDArray;
+import ai.djl.ndarray.NDManager;
+import ai.djl.ndarray.index.NDIndex;
+import ai.djl.ndarray.types.Shape;
+import ai.djl.pytorch.engine.PtNDArray;
+import ai.djl.pytorch.engine.PtNDManager;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class ManagerAttachmentTest {
```
zachgk, line 24: Instead of creating these tests for both MXNet engine and PyTorch engine, can you move it into the integration module? That is where we put the tests representing behavior shared for the entire engine spec. Engines that don't fully implement the spec can return an UnsupportedOperationException and it will not be considered failing the test

api/src/main/java/ai/djl/ndarray/NDArray.java
```diff
@@ -508,7 +508,18 @@ default void setScalar(NDIndex index, Number value) {
...
+    }
+
+    /**
+     * Returns a partial {@code NDArray}.
+     *
+     * @param manager the manager used to create the arrays
+     * @param index the section of this {@code NDArray} to return
+     * @return the partial {@code NDArray}
+     */
+    default NDArray get(NDIndex index, NDManager manager) {
```
zachgk, line 521: Minor thing, but I feel like it would be more consistent to put manager as the first argument

zachgk (commented):
This is a good change @patins1!

zachgk (approved):

PR #1691, opened 2022-06-02
bump up versions

lanking520:
## Description ##

Bump up versions in engine

### DIFF
--- a/engines/paddlepaddle/paddlepaddle-engine/src/main/java/ai/djl/paddlepaddle/jni/JniUtils.java
+++ b/engines/paddlepaddle/paddlepaddle-engine/src/main/java/ai/djl/paddlepaddle/jni/JniUtils.java
@@ -141,8 +141,4 @@ public final class JniUtils {
     public static String[] getInputNames(PaddlePredictor predictor) {
         return PaddleLibrary.LIB.getInputNames(predictor.getHandle());
     }
-
-    public static String getVersion() {
-        return "2.0.2";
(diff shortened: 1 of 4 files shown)


frankfliu (approved):

engines/paddlepaddle/paddlepaddle-engine/src/main/java/ai/djl/paddlepaddle/jni/JniUtils.java
```diff
@@ -143,6 +143,6 @@ public static String[] getInputNames(PaddlePredictor predictor) {
     }
 
     public static String getVersion() {
-        return "2.0.2";
+        return "2.2.2";
```
frankfliu, line 146:
```
        try (InputStream is =
                XgbEngine.class.getResourceAsStream("/paddlepaddle-engine.properties")) {
            Properties prop = new Properties();
            prop.load(is);
            return prop.getProperty("paddlepaddle-version");
        } catch (IOException e) {
            throw new AssertionError("Failed to load paddlapaddle-engine.properties", e);
        }
```

frankfliu: Add a test in unitest for getVersion

frankfliu (changes requested):

frankfliu (approved):

PR #1696, opened 2022-06-06
[osx m1] Allow DJL build on macOS M1 chip

frankfliu:
Change-Id: I63f6bcb9308647178e8d5583b3582408f6013590

## Description ##

### DIFF
--- a/testing/src/main/java/ai/djl/testing/TestRequirements.java
+++ b/testing/src/main/java/ai/djl/testing/TestRequirements.java
@@ -91,4 +91,11 @@ public final class TestRequirements {
             throw new SkipException("This test requires a non-windows os.");
         }
     }
+
+    /** Requires that the test runs on x86_64 arch. */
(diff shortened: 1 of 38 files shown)


testing/src/main/java/ai/djl/testing/TestRequirements.java
```diff
@@ -91,4 +91,11 @@ public static void notWindows() {
             throw new SkipException("This test requires a non-windows os.");
         }
     }
+
+    /** Requires that the test runs on x86_64 arch. */
+    public static void notArm() {
+        if ("aarch64".equals(System.getProperty("os.arch"))) {
+            throw new SkipException("This test requires a non-windows os.");
```
zachgk, line 98:
```suggestion
            throw new SkipException("This test requires a non-arm os.");
```

zachgk (approved):

PR #1702, opened 2022-06-08
[api] Refactor ServingTranslatorFactory

frankfliu:
Change-Id: I63cee2f2c8dcb9deec39e037a5291dbf0d68f045

## Description ##

### DIFF
--- a/api/src/main/java/ai/djl/translate/ServingTranslatorFactory.java
+++ b/api/src/main/java/ai/djl/translate/ServingTranslatorFactory.java
@@ -40,6 +40,7 @@ import java.util.Set;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.tools.JavaCompiler;
 import javax.tools.ToolProvider;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


lanking520:
Can you explain what is this change for?

api/src/main/java/ai/djl/translate/ServingTranslatorFactory.java
```diff
@@ -81,7 +82,7 @@ public Set<Pair<Type, Type>> getSupportedTypes() {
         Path libPath = modelDir.resolve("libs");
         if (!Files.isDirectory(libPath)) {
             libPath = modelDir.resolve("lib");
-            if (!Files.isDirectory(libPath)) {
+            if (!Files.isDirectory(libPath) && className == null) {
```
frankfliu, line 85: If the "lib/libs" folder doesn't exist, we should keep searching in the classpath, so we can load the translator from "deps" folder or system classpath

api/src/main/java/ai/djl/translate/ServingTranslatorFactory.java
```diff
@@ -98,10 +99,17 @@ private ServingTranslator findTranslator(Path path, String className) {
             Path classesDir = path.resolve("classes");
             compileJavaClass(classesDir);
 
-            List<Path> jarFiles =
-                    Files.list(path)
-                            .filter(p -> p.toString().endsWith(".jar"))
-                            .collect(Collectors.toList());
+            List<Path> jarFiles = new ArrayList<>();
+            if (Files.isDirectory(path)) {
```
frankfliu, line 103: If the folder doesn't exists, IOException will throw, this will prevent class loading from system classpath

lanking520 (approved):

PR #1704, opened 2022-06-09
[docs] Create README for DJL Android PyTorch 1.11 builds

JohnDoll2023:
## Description ##

Creates README for how to build DJL Android Apps with PyTorch 1.11

- Currently works with Style Transfer only


android/pytorch-native/README.md
```diff
@@ -0,0 +1,137 @@
...
+export NDK_VERSION=21.0.6113669
+echo "y" | sudo ${ANDROID_SDK_ROOT}/cmdline-tools/tools/bin/sdkmanager --install "ndk;${NDK_VERSION}"
+```
+<br>
+
+## Build PyTorch Android native binary from source
+See: [https://github.com/deepjavalibrary/djl/blob/master/.github/workflows/native_s3_pytorch_android.yml](https://github.com/deepjavalibrary/djl/blob/master/.github/workflows/native_s3_pytorch_android.yml)
+
+Prerequisites:
+* If you do not have `cmake` installed, run `brew install cmake`
```
zachgk, line 55: Can you move all of the prerequisite installations to the top in one big command. Most install tools are idempotent - they will be fine if it includes an installation you already have. So, just have one big `brew install cmake python android-sdk ...`.

android/pytorch-native/README.md
```diff
@@ -0,0 +1,137 @@
...
+* If you do not have the python library `pyyaml` installed, run `pip install pyyaml`
+* If you have `python3` installed, run `export PYTHON=python3`
```
zachgk, line 57: I would just include this into the script. Most systems nowadays should have python3. Maybe keep the comment to remove if you don't have python3

android/pytorch-native/README.md
```diff
@@ -0,0 +1,137 @@
...
+# From link above, set ANDROID_CLT equal to the name of Linux's latest SDK tools package release
+export ANDROID_CLT=commandlinetools-linux-8512546_latest.zip
+
+curl -O https://dl.google.com/android/repository/${ANDROID_CLT}
+unzip $ANDROID_CLT
+rm $ANDROID_CLT
+
+mv cmdline-tools tools
+```
+<br>
```
zachgk, line 40: You probably don't need these extra spaces

android/pytorch-native/README.md
```diff
@@ -0,0 +1,137 @@
+
+# Pytorch Android Build Procedure
+## Linux (Ubuntu 20.04) android-sdk install
+```
```
zachgk, line 4: In markdown, adjacent lines are considered part of the same paragraph. So, this says that the top level heading, second level heading, and code block are in the same paragraph. Instead, what you want is an empty line separating these different pieces.

Now, most parsers will actually interpret this correctly anyway. However, the one we use to build our docs site (https://docs.djl.ai) doesn't. So you will actually need to add all the separating line breaks here to have it render properly

zachgk (commented):
@frankfliu

zachgk (approved):

PR #1717, opened 2022-06-14
Split EasyTrain to add EasyTrainer

zachgk:
This creates a more extensible and non-final class EasyTrainer to contain
training functionality formerly in EasyTrain. There have been a few requests to
be able to modify EasyTrain and this should enable it.

EasyTrain is then left as a static wrapper around a shared instance of the
EasyTrain class. There shouldn't be any API changes required to use EasyTrain.

### DIFF
--- a/api/src/main/java/ai/djl/training/EasyTrain.java
+++ b/api/src/main/java/ai/djl/training/EasyTrain.java
@@ -29,6 +29,8 @@ import java.util.concurrent.ExecutorService;
 /** Helper for easy training of a whole model, a trainining batch, or a validation batch. */
 public final class EasyTrain {
 
+    public static final EasyTrainer TRAINER = new EasyTrainer();
+
     private EasyTrain() {}
 
     /**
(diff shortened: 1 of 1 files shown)


api/src/main/java/ai/djl/training/EasyTrain.java
```diff
@@ -29,6 +29,8 @@
 /** Helper for easy training of a whole model, a trainining batch, or a validation batch. */
 public final class EasyTrain {
 
+    public static final EasyTrainer TRAINER = new EasyTrainer();
```
KexinFeng, line 32: In the description, it says "non-final class EasyTrainer", but here it is final. Is this an error?
I'm wondering how this change enables users to modify EasyTrain. Is there a brief demo?

zachgk: The class `EasyTrainer` is not final, but this instance `TRAINER` which serves as the default `EasyTrainer` is final.

KexinFeng (approved):

zachgk:
Closing due to lack of interest

PR #1667, opened 2022-05-21
[basicdataset] Add Daily Delhi Climate Dataset

WHALEEYE:
## Description ##

Added Daily Delhi Climate Dataset, which is a time series dataset into DJL.
Since the dataset is not uploaded to s3 yet, the download process and the correctness of URI is not tested.

Close #1590.

PR #1718, opened 2022-06-15
Adding Movielens 100k tabular dataset to basicdatasets

siddvenk:
## Description ##

Adding the Movielens 100k dataset to the tabular datasets. This dataset can be used to train regression models, collaborative filtering models, Graph Neural Networks, and a host of other recommendation system models. 

https://grouplens.org/datasets/movielens/100k/

## Note ##
The tests right now are using a local instace of the repo since the metadata hasn't been published to S3. I'll update the tests accordingly once the metadata.json has been pushed to S3 and I can verify using the remote bucket works.

### DIFF
--- a/basicdataset/src/main/java/ai/djl/basicdataset/tabular/MovieLens100k.java
+++ b/basicdataset/src/main/java/ai/djl/basicdataset/tabular/MovieLens100k.java
@@ -1,0 +1,344 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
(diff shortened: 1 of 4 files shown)


siddvenk:
Ubuntu/Windows builds are failing because the associated unit test is failing. This is expected with the way I hard coded the path for the repository used in the test. Seems like there are 2 ways I can resolve this:
1. Use a relative path in the repo - test will still use a local repo but relative path should at least make finding the files work
2. Use the default repo that can only work once the metatdata has been added to the djl repo.

I see examples of both being used for the basicdataset tests - do we have a preference?

lanking520:
Not an expertise in this part, but I think you can try to follow: https://github.com/deepjavalibrary/djl/pull/1667 this PR to see how they do the test. @zachgk any thoughts?

siddvenk:
@lanking520 I was thinking of doing something similar to that. I wonder whether it makes sense to standardize across the basicdataset tests? Tests for Cifar10, Mnist, ImageNet, Wiki2Text are using a local repo in the tests, while others are using djl repo from what I can tell (they're not providing a repo when building, so it will use the default djl repo). 

Using the djl repo in the tests seems better to me to validate the data is stored correctly, configuration in the code is correct, and the data can be pulled to be built as a dataset. With local repo we can avoid some calls to the repo to fetch the metadata and files, but since the files are cached and don't change often I'm not sure if there's much benefit using a local repo in the tests.

zachgk (approved):
I modified the metadata to add a name to the license. That way, if we show the license name somewhere it would at least have some reasonable information conveyed.

The metadata is uploaded so you can modify the test to pull from DJL central rather local now

PR #1723, opened 2022-06-17
Ensure PreparedOneHotStringFeaturizer encodes categorical mappings co…

siddvenk:
Ensure PreparedOneHotStringFeaturizer encodes categorical mappings consistently across dataset usages

## Description ##

For Tabular datasets, the PreparedOneHotStringFeaturizer currently builds a list of values for a feature as we [iterate through the records](https://github.com/deepjavalibrary/djl/blob/master/basicdataset/src/main/java/ai/djl/basicdataset/tabular/TabularDataset.java#L92-L94). Since we build the OHE mapping [based on the order the values are read](https://github.com/deepjavalibrary/djl/blob/master/basicdataset/src/main/java/ai/djl/basicdataset/tabular/utils/Featurizers.java#L171-L179), if the order of values read differs across datasets (train/test/val) then the OHE mappings differ across usages where the same model would be used.

To fix this, we should sort the values (how we sort doesn't matter as long as it is consistent across dataset usages) before constructing the OHE mappings. This does assume that the the values of a certain feature we want to OHE are the same and appear at least once across all dataset usages, which I think is a fair assumption to make. Featurizers should not be concerned with whether the dataset we are reading from has been constructed appropriately or not.

I updated the MovieLens100k dataset as part of this change since I noticed this issue while trying to create an example using this dataset.

## Example##

As an example, consider the following where we want to OHE the State feature:
```
--Train Data--
|row#|State|...
|1|CA|...
|2|CO|...
|3|NY|...

--Test Data--
|row#|State|...
|1|CO|...
|2|NY|...
|3|CA|...
```
The training data is prepared such that the OHE values are `CA=[1,0,0]`, `CO=[0,1,0]`, `NY=[0,0,1]` and the model we train will learn/assume those mappings apply to the test set we want to evaluate on. But in testing, we would prepare the OHE values as `CA=[0,0,1]`, `CO=[1,0,0]`, `NY=[0,1,0]` but the model wouldn't know that these mappings have changed.

### DIFF
--- a/basicdataset/src/main/java/ai/djl/basicdataset/tabular/utils/Featurizers.java
+++ b/basicdataset/src/main/java/ai/djl/basicdataset/tabular/utils/Featurizers.java
@@ -17,6 +17,7 @@ import java.time.format.DateTimeFormatter;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 
 /** A utility class provides helper functions to create {@link Featurizer}. */
@@ -170,7 +171,8 @@ public final class Featurizers {
         @Override
         public void prepare(List<String> inputs) {
             map = new ConcurrentHashMap<>();
-            for (String input : inputs) {
+            TreeSet<String> uniqueInputs = new TreeSet<>(inputs);
+            for (String input : uniqueInputs) {
                 if (!map.containsKey(input)) {
                     map.put(input, map.size());
                 }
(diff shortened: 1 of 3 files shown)


basicdataset/src/main/java/ai/djl/basicdataset/tabular/utils/Featurizers.java
```diff
@@ -170,7 +171,8 @@ public PreparedOneHotStringFeaturizer() {
         @Override
         public void prepare(List<String> inputs) {
             map = new ConcurrentHashMap<>();
-            for (String input : inputs) {
+            TreeSet<String> uniqueInputs = new TreeSet<>(inputs);
```
siddvenk, line 174: We can definitely use Collections.sort(inputs) here as well to achieve the same thing. I opted for TreeSet since this is a public method and an in place sort could be considered an unintended side-effect.

zachgk (approved):

PR #1749, opened 2022-06-25
Add `put` feature with linear indexing on PyTorch engine

KexinFeng:
## Description ##

Add `put` feature support on PyTorch engine. This corresponds to `take` which get values according to linear indexing.

The relavant preceding PRs
[Support of take from pytorch #1627](https://github.com/deepjavalibrary/djl/pull/1627)

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/NDArray.java
+++ b/api/src/main/java/ai/djl/ndarray/NDArray.java
@@ -581,8 +581,8 @@ public interface NDArray extends NDResource, BytesSupplier {
     NDArray gather(NDArray index, int axis);
 
     /**
-     * Returns a partial {@code NDArray} pointed by index according to linear indexing, and
-     * the of output is of the same shape as index.
+     * Returns a partial {@code NDArray} pointed by index according to linear indexing, and the of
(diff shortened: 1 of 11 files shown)


api/src/main/java/ai/djl/ndarray/NDArray.java
```diff
@@ -581,14 +581,26 @@ default NDArray get(NDManager manager, long... indices) {
...
+    /**
+     * Set the entries of {@code NDArray} pointed by index according to linear indexing, to be the
+     * numbers in data, which is of the same shape as index.
+     *
+     * @param index select the entries of an {@code NDArray}
+     * @param data numbers to assign to the indexed entries
+     * @return The NDArray with updated values
+     */
+    default NDArray put(NDArray index, NDArray data) {
+        throw new UnsupportedOperationException("Not implemented yet.");
```
zachgk, line 601: We usually don't put the Unsupported default in NDArray because it makes it hard to keep track of for the "full" engines. Put it in the NDArrayAdapter and the full engines that are missing it (should be MX and TF)

zachgk (approved):

PR #1751, opened 2022-06-28
Create action to auto-close issues without response

zachgk:
This creates an action that runs daily after midnight. It will look for actions
with a given label (defaults to more-information-needed) that have not been
updated for 14 days and will close them.


lanking520:
Can we extend to 30 days?

frankfliu (approved):

PR #1771, opened 2022-07-05
[ort] Add support for registerCustomOpLibrary.

frankfliu:
Change-Id: Ic3ec4249adb432b2a54668b7a45481698fa83ea5

## Description ##


zachgk (commented):
Should we document this option somewhere?

lanking520 (approved):

Issue #1799, opened 2022-07-15
NDArray.get(NDArray) breaking behavior in 0.18.0


siddvenk:
## Description
There seems to be backwards incompatible behavior with the NDArray.get(NDArray) method. In DJLv0.17.0 the following code works as expected, but in DJLv0.18.0 it throws an IllegalArgumentException at the `xTile.get(...)` line. 

```java
try (NDManager manager = NDManager.newBaseManager()) {
            int nTrain = 5;
            NDArray xTrain = manager.randomUniform(0, 1, new Shape(nTrain)).mul(5).sort();
            NDArray xTile = xTrain.tile(new long[] {nTrain, 1});
            // Following line works as expected in v0.17 and returns the diagonal elements
            // Throws IllegalArgumentException in v0.18
            NDArray keys = xTile.get((manager.eye(nTrain)).reshape(new Shape(nTrain, -1)));
}
```

### Expected Behavior
The above code should work and return an NDArray (shape 1,5) with the diagonal elements from xTile.  

### Error Message
```
[ERROR] - {}
java.lang.IllegalArgumentException: Unknown argument: ND: (5, 5) cpu() float32
[[1., 0., 0., 0., 0.],
 [0., 1., 0., 0., 0.],
 [0., 0., 1., 0., 0.],
 [0., 0., 0., 1., 0.],
 [0., 0., 0., 0., 1.],
]

        at ai.djl.ndarray.index.NDIndex.addIndexItem(NDIndex.java:384) ~[api-0.19.0-SNAPSHOT.jar:?]
        at ai.djl.ndarray.index.NDIndex.addIndices(NDIndex.java:220) ~[api-0.19.0-SNAPSHOT.jar:?]
        at ai.djl.ndarray.index.NDIndex.<init>(NDIndex.java:130) ~[api-0.19.0-SNAPSHOT.jar:?]
        at ai.djl.ndarray.NDArray.get(NDArray.java:533) ~[api-0.19.0-SNAPSHOT.jar:?]
        at ai.djl.examples.bugs.NDArrayIndexingBug.main(NDArrayIndexingBug.java:32) ~[main/:?]
```

## How to Reproduce?
Here's the code I'm running. To reproduce the issue run this against either master or v0.18 tag. If you run this against v0.17 tag it works as expected. 

I added this code to a file in the examples/inference module and ran it via `./gradlew run -Dmain=ai.djl.examples.inference.NDArrayIndexingBug`

```java
[...]

PR #1801, opened 2022-07-15
[BugFix] Fix breaking behaviour in 0.18.0

KexinFeng:
## Description ##

This fixes [issue](https://github.com/deepjavalibrary/djl/issues/1799). 

The fix is to allow other datatype of index than just integer. The dataType conversion to `Long` has already implemented before in JniUtils

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/index/NDIndex.java
+++ b/api/src/main/java/ai/djl/ndarray/index/NDIndex.java
@@ -21,7 +21,6 @@ import ai.djl.ndarray.index.dim.NDIndexNull;
 import ai.djl.ndarray.index.dim.NDIndexPick;
 import ai.djl.ndarray.index.dim.NDIndexSlice;
 import ai.djl.ndarray.index.dim.NDIndexTake;
-import ai.djl.ndarray.types.DataType;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


api/src/main/java/ai/djl/ndarray/index/NDIndex.java
```diff
@@ -373,7 +373,7 @@ private int addIndexItem(String indexItem, int argIndex, Object[] args) {
                     if (array.getDataType() == DataType.BOOLEAN) {
                         indices.add(new NDIndexBooleans(array));
                         return argIndex + 1;
-                    } else if (array.getDataType().isInteger()) {
```
frankfliu, line 376: Why our integration test didn't catch this issue?
Can we add more test to ensure this not happen again?

api/src/main/java/ai/djl/ndarray/index/NDIndex.java
```diff
@@ -373,7 +373,7 @@ private int addIndexItem(String indexItem, int argIndex, Object[] args) {
...
+                    } else {
```
siddvenk, line 376: Would this cause issues if DataType is UNKNOWN or STRING? https://github.com/deepjavalibrary/djl/blob/master/api/src/main/java/ai/djl/ndarray/types/DataType.java#L28-L34 seems like what we need to cover in this code, and not UNKNOWN or STRING

api/src/main/java/ai/djl/ndarray/index/NDIndex.java
siddvenk: +1 we can add the case from the bug that reproduces this as part of our tests

KexinFeng: It was ot caught since previously the tests only tested integer index arrays. Now I have added float index array as well.

api/src/main/java/ai/djl/ndarray/index/NDIndex.java
KexinFeng: Thanks for spotting this. I'll update it.

frankfliu (approved):

siddvenk (approved):

PR #1802, opened 2022-07-16
Backward compatible with MXNet indexing.

KexinFeng:
This solves the bug reported in [issue](https://github.com/deepjavalibrary/djl/issues/1800). 

The new feature of indexing is not completely backward compatible, since `NDIndexPick` has been replaced by `NDIndexTake`. Now this problem is solved.

However, this is a temporary solution, only to make it backward compatible. `NDIndexPick` does not have the same definition as `NDIndexTake` when used to index array elements. For example, `NDIndexPick` cannot support larger-than-1-rank indexing array, while `NDIndexTake` can. So, as NDIndexElement,  `NDIndexPick` should be deprecated in the future.

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/index/full/NDIndexFullPick.java
+++ b/api/src/main/java/ai/djl/ndarray/index/full/NDIndexFullPick.java
@@ -52,22 +52,26 @@ public final class NDIndexFullPick {
         for (NDIndexElement el : index.getIndices()) {
             if (el instanceof NDIndexAll) {
                 axis++;
-            } else if (el instanceof NDIndexPick || el instanceof NDIndexTake) {
-                if (fullPick == null) {
-                    NDArray indexElem =
-                            el instanceof NDIndexPick
-                                    ? ((NDIndexPick) el).getIndex()
-                                    : ((NDIndexTake) el).getIndex();
-                    if (el instanceof NDIndexTake && !indexElem.getShape().isRankOne()) {
-                        throw new UnsupportedOperationException(
-                                "Only rank-1 indexing array is supported for pick");
-                    }
-                    fullPick = new NDIndexFullPick(indexElem, axis);
-                } else {
+            } else if (el instanceof NDIndexPick) {
+                if (fullPick != null) {
                     // Don't support multiple picks
                     throw new UnsupportedOperationException(
                             "Only one pick per get is currently supported");
                 }
+                NDArray indexElem = ((NDIndexPick) el).getIndex();
+                fullPick = new NDIndexFullPick(indexElem, axis);
+            } else if (el instanceof NDIndexTake) {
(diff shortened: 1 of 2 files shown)


api/src/main/java/ai/djl/ndarray/types/Shape.java
```diff
@@ -477,4 +477,24 @@ public static Shape decode(DataInputStream dis) throws IOException {
...
+     *
+     * <p>For example, an array with shape [1, 10, 1] returns true. Array with indeterminate size -1
+     * returns false.
+     *
+     * @return if the array is rank-1
+     */
+    public boolean isRankOne() {
+        int max = 1, ans = 1;
+        for (long s : shape) {
+            max = Math.max(max, Math.toIntExact(s));
```
frankfliu, line 492:
```suggestion
            int size =  Math.toIntExact(s);
            max = Math.max(max, size);
```

api/src/main/java/ai/djl/ndarray/types/Shape.java
```diff
@@ -477,4 +477,24 @@ public static Shape decode(DataInputStream dis) throws IOException {
...
+            ans *= s;
```
frankfliu, line 493:
```suggestion
            ans *= size;
```

api/src/main/java/ai/djl/ndarray/index/full/NDIndexFullPick.java
```diff
@@ -51,9 +52,17 @@ public static Optional<NDIndexFullPick> fromIndex(NDIndex index, Shape target) {
         for (NDIndexElement el : index.getIndices()) {
             if (el instanceof NDIndexAll) {
                 axis++;
-            } else if (el instanceof NDIndexPick) {
+            } else if (el instanceof NDIndexPick || el instanceof NDIndexTake) {
```
frankfliu, line 55: I think the following might be more readable:
```
            } else if (el instanceof NDIndexPick) {
                if (fullPick != null) {
                    // Don't support multiple picks
                    throw new UnsupportedOperationException(
                            "Only one pick per get is currently supported");
                }
                NDArray indexElem = ((NDIndexPick) el).getIndex();
                fullPick = new NDIndexFullPick(indexElem, axis);
            } else if (el instanceof NDIndexTake) {
                if (fullPick != null) {
                    // Don't support multiple picks
                    throw new UnsupportedOperationException(
                            "Only one pick per get is currently supported");
                }
                NDArray indexElem = ((NDIndexTake) el).getIndex();
                if (!indexElem.getShape().isRankOne()) {
                    throw new UnsupportedOperationException(
                            "Only rank-1 indexing array is supported for pick");
                }
                fullPick = new NDIndexFullPick(indexElem, axis);
```

lanking520 (approved):

PR #1808, opened 2022-07-19
Compatibility changes for Semantic Segmentation app

JohnDoll2023:
## Description ##

Adds compatibility for the DJL-Demo Semantic Segmentation app

- Transposes NDArray if dimensions matchup in BitmapImageFactory
- DataType for NDArray would change from the app due to using BitmapImageFactory and not BufferImageFactory, so added line to change DataType back to `UINT8` before returning image.

### DIFF
--- a/api/src/main/java/ai/djl/modality/cv/translator/SemanticSegmentationTranslator.java
+++ b/api/src/main/java/ai/djl/modality/cv/translator/SemanticSegmentationTranslator.java
@@ -110,10 +110,12 @@ public class SemanticSegmentationTranslator extends BaseImageTranslator<Image> {
             }
             int originW = (int) ctx.getAttachment("originalWidth");
             int originH = (int) ctx.getAttachment("originalHeight");
-            NDArray intRet = manager.create(bb, new Shape(height, width, CHANNEL), DataType.UINT8);
-            NDArray resized = NDImageUtils.resize(intRet, originW, originH);
+            NDArray fullImage =
+                    manager.create(bb, new Shape(height, width, CHANNEL), DataType.UINT8);
+            NDArray resized = NDImageUtils.resize(fullImage, originW, originH);
+            NDArray ret = resized.toType(DataType.UINT8, true);
 
-            return ImageFactory.getInstance().fromNDArray(resized);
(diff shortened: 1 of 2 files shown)


api/src/main/java/ai/djl/modality/cv/translator/SemanticSegmentationTranslator.java
```diff
@@ -110,10 +110,12 @@ public Image processOutput(TranslatorContext ctx, NDList list) {
             }
             int originW = (int) ctx.getAttachment("originalWidth");
             int originH = (int) ctx.getAttachment("originalHeight");
-            NDArray intRet = manager.create(bb, new Shape(height, width, CHANNEL), DataType.UINT8);
-            NDArray resized = NDImageUtils.resize(intRet, originW, originH);
+            NDArray fullImage =
+                    manager.create(bb, new Shape(height, width, CHANNEL), DataType.UINT8);
+            NDArray resized = NDImageUtils.resize(fullImage, originW, originH);
```
lanking520, line 115: Does image change data type after resizing?

JohnDoll2023: When calling from DJL repo, no. When calling from DJL-demo where the app is, yes. On the resize, DJL-Demo calls BitmapImageFactory (instead of BufferedImageFactory which is called by DJL) and this ends up changing the data type so this is a safeguard that prevents the datatype from changing before it is sent to be returned, where it must be of type `UINT8`.

frankfliu (approved):

api/src/main/java/ai/djl/modality/cv/translator/SemanticSegmentationTranslator.java
zachgk: I would be more inclined to treat that as a problem with resize. Maybe add a `toType` to the `resize` method that is being called to ensure that it doesn't change data types

JohnDoll2023: Will push change in new pr

PR #1811, opened 2022-07-19
Add MobileNetV1 and MobileNetV1 test into ModelZoo

warthecatalyst:
## Description ##
Add MobileNet version1 (reference paper https://arxiv.org/pdf/1704.04861.pdf) into ModelZoo and create a model test to check out whether the shapes and values of parameters meet our expectations


lanking520:
There is no content in this PR

warthecatalyst:
sorry I forgot to push my local changes

PR #1813, opened 2022-07-19
Fixes resize return DataType mismatch bug

JohnDoll2023:
## Description ##

Fixes the NDArray from being set to a float32 format

- When calling the resize method from the DJL-demo repo, the `PtNDArrayEx.java` is called and changes the `NDArray` type from `UINT8` to `float32`. This fix changes it back to it's original dataType that it was prior to being changed to `float32`.

### DIFF
--- a/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArrayEx.java
+++ b/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArrayEx.java
@@ -582,7 +582,8 @@ public class PtNDArrayEx implements NDArrayEx {
                                     new long[] {height, width},
                                     getInterpolationMode(interpolation),
                                     false)
-                            .transpose(0, 2, 3, 1);
+                            .transpose(0, 2, 3, 1)
+                            .toType(array.getDataType(), false);
             if (dim == 3) {
                 result = result.squeeze(0);
             }
--- a/api/src/main/java/ai/djl/modality/cv/translator/SemanticSegmentationTranslator.java
+++ b/api/src/main/java/ai/djl/modality/cv/translator/SemanticSegmentationTranslator.java
@@ -113,9 +113,8 @@ public class SemanticSegmentationTranslator extends BaseImageTranslator<Image> {
             NDArray fullImage =
                     manager.create(bb, new Shape(height, width, CHANNEL), DataType.UINT8);
             NDArray resized = NDImageUtils.resize(fullImage, originW, originH);
-            NDArray ret = resized.toType(DataType.UINT8, false);
 
-            return ImageFactory.getInstance().fromNDArray(ret);
+            return ImageFactory.getInstance().fromNDArray(resized);
         }
     }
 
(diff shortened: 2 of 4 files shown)


engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArrayEx.java
```diff
@@ -582,7 +582,8 @@ public PtNDArray resize(int width, int height, int interpolation) {
                                     new long[] {height, width},
                                     getInterpolationMode(interpolation),
                                     false)
-                            .transpose(0, 2, 3, 1);
+                            .transpose(0, 2, 3, 1)
+                            .toType(array.getDataType(), false);
```
frankfliu, line 586: The resize operation should not change data type. The pytorch ops only support float32, that's why we convert to float32, but we have to convert back to uint8 if input is uint8.

We also need to add output check in integration tests, and make sure this is the same for all engines.

JohnDoll2023: Added integration test

frankfliu: It turns out the result of Resize should be float32.

And we are manually convert NDArray to UINT8/INT8 manually before calling `ImageFactory.fromNDArray()`
I think we should call `NDArray.toType(UINT8, false)` inside  `ImageFactory.fromNDArray`

We can drop this PR for now.

zachgk: @frankfliu I'm not sure what you mean by the resize should be flaot32. Shouldn't it be the same as input type? The two main formats for images are uint8 from 0-255 and float32 from 0-1, and resize shouldn't be changing between them. Due to limitations in pytorch engine, it only works for float32. Currently, we convert to float32 to make it callable, but we should also convert back after to maintain the input format. It applies more broadly than just `ImageFactory.fromNDArray()`

frankfliu: Here is the problem
1. All the models take float32 as input, if we convert float32 to uint8, at one point we have to convert to float32 again
2. there actually three format: uint8, float32 (0,1) and float32 (-1,1), if we convert to uint8, and then convert to float32, the range messed up
3. For models that applies toTensor, uint8 isn't a problem, toTensor will convert any input to float32, however for tensorflow model, resize is their last transform (the model takes channel last image), we have to apply a toType(float32) for those models
4. PT resize operation requires float32 input, and TF automatically convert to float32 natively. We can define resize output should be always float32.

frankfliu: Our ImageFactory always convert image to UINT8 NDArray, that means both TF and PT resize() must convert to UINT8, then we must add a step to convert to float32 before feed into model.

We didn't expose resize as NDArray operation, what we exposed is `NDImageUtils.resize()`. If this API can accept both. UINT8/INT8/FLOAT32 input and output is always float32, it actually make sense.

PR #1817, opened 2022-07-20
Add MobileNetV1 into modelZoo and add corresponding test class MobileNetV1Test

warthecatalyst:
Add MobileNetV1 into modelZoo and add corresponding test class MobileNetV1Test
modify train method of ImageClassification Class in djl-zero, to train a MobileNet for fast cases and ResNet for large cases


model-zoo/src/main/java/ai/djl/basicmodelzoo/cv/classification/MobileNetV1.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    /** The Builder to construct a {@link MobileNetV1} object. */
+    public static final class Builder {
+        float batchNormMomentum = 0.9f;
+        float alpha = 1f; // width multiplier defined in the paper
+
+        long outSize = 10; // 10 as default for basic datasets like cifar10 or mnist
+
+        int[] filters = {32, 64, 128, 128, 256, 256, 512, 512, 1024, 1024};
```
zachgk, line 227: Here, you have filters as part of the builder but it isn't actually modifiable. This can be confusing because it is not how the Builder pattern typically works. There really a few straightforward optiosn:

1. Move the option to a hardcoded part of `mobilenet(..)` to more clearly indicate that it is not part of the Builder pattern
2. Create an option to modify `filters`. Then, you would also need to handle the case of different filter lengths. This can be either by adding a validation to require the specific length of 10 or by creating a loop so the block can be built with arbitrary filter lengths. Maybe this would include adding a strides array with the additional restriction that strides and filters have the same length.

zachgk (commented):
This looks pretty good. I just have two minor suggestions

zachgk (approved):

PR #1819, opened 2022-07-20
Update bitmap factory for Semantic App

JohnDoll2023:
## Description ##

This fixes the Semantic app issue where the image was grayscale and duplicated 9 times

- Does not transpose the array anymore, instead, just gets the shape height and width where they exist in the NDArray
- Checked all other apps and this fixed had no affect on them as they were working correctly both prior to and after this change

### DIFF
--- a/android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
+++ b/android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
@@ -85,6 +85,8 @@ public class BitmapImageFactory extends ImageFactory {
     @Override
     public Image fromNDArray(NDArray array) {
         Shape shape = array.getShape();
+        int height = (int) shape.get(1);
+        int width = (int) shape.get(2);
         if (shape.dimension() != 3) {
             throw new IllegalArgumentException("Shape should only have three dimension follow CHW");
         }
@@ -95,20 +97,20 @@ public class BitmapImageFactory extends ImageFactory {
             throw new UnsupportedOperationException("Grayscale image is not supported");
         } else if (shape.get(0) != 3) {
             if (shape.get(2) == 3) {
-                array = array.transpose(2, 0, 1);
-                shape = array.getShape();
+                height = (int) shape.get(0);
+                width = (int) shape.get(1);
             } else {
                 throw new IllegalArgumentException("First or last dimension should be number of channel with value 1 or 3");
             }
         }
-        int height = (int) shape.get(1);
-        int width = (int) shape.get(2);
-        int imageArea = width * height;
-        Bitmap bitmap = Bitmap.createBitmap(width, height, Bitmap.Config.ARGB_8888);
+        int finalWidth = width;
+        int finalHeight = height;
+        int imageArea = finalWidth * finalHeight;
+        Bitmap bitmap = Bitmap.createBitmap(finalWidth, finalHeight, Bitmap.Config.ARGB_8888);
         int[] raw = array.toUint8Array();
         IntStream.range(0, imageArea).parallel().forEach(ele -> {
-            int x = ele % width;
-            int y = ele / width;
+            int x = ele % finalWidth;
+            int y = ele / finalWidth;
             int red = raw[ele] & 0xFF;
             int green = raw[ele + imageArea] & 0xFF;
             int blue = raw[ele + imageArea * 2] & 0xFF;
--- a/android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
+++ b/android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
@@ -91,7 +91,10 @@ public class BitmapImageFactory extends ImageFactory {
         if (shape.dimension() != 3) {
             throw new IllegalArgumentException("Shape should only have three dimension follow CHW");
         }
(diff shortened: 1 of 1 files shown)


android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
```diff
@@ -86,24 +86,29 @@ public Image fromImage(Object image) {
...
-            array = array.transpose(2, 0, 1);
-            shape = array.getShape();
-        }
-        int height = (int) shape.get(1);
-        int width = (int) shape.get(2);
-        int imageArea = width * height;
-        Bitmap bitmap = Bitmap.createBitmap(width, height, Bitmap.Config.ARGB_8888);
-        int[] raw = array.toType(DataType.UINT8, false).toUint8Array();
+        int finalWidth = width;
+        int finalHeight = height;
```
lanking520, line 105: Why do we introduce final width and height? Can we reuse?

android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
```diff
@@ -86,24 +86,29 @@ public Image fromImage(Object image) {
...
+        int imageArea = finalWidth * finalHeight;
+        Bitmap bitmap = Bitmap.createBitmap(finalWidth, finalHeight, Bitmap.Config.ARGB_8888);
+        int[] raw = array.toUint8Array();
         IntStream.range(0, imageArea).parallel().forEach(ele -> {
-            int x = ele % width;
-            int y = ele / width;
+            int x = ele % finalWidth;
```
lanking520, line 110: This does not applies to HWC. HWC (Channel last) will display the data like 

For each height and width:
Show R,G,B (3 channel color)

CHW like

For each channel
 Hight and width color

lanking520 (changes requested):

android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
lanking520: So you may need to determine if here is CHW or HWC and customize your loading logic

android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
JohnDoll2023: In the [BufferedImageFactory](https://github.com/JohnDoll2023/djl/blob/master/api/src/main/java/ai/djl/modality/cv/BufferedImageFactory.java#L105), the width is set to final there so I was trying to follow the same convention

android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
JohnDoll2023: I have tested this on all the Android apps and this method works. Originally for my app, I was transposing the NDArray to CHW, but that caused an error in the image being shown in the app. @frankfliu and I were getting a 3x3 grid of mini grayscale images back, but when I took out the transpose, the image was shown correctly.

android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
```diff
@@ -86,24 +86,29 @@ public Image fromImage(Object image) {
...
         Shape shape = array.getShape();
+        int height = (int) shape.get(1);
+        int width = (int) shape.get(2);
         if (shape.dimension() != 3) {
             throw new IllegalArgumentException("Shape should only have three dimension follow CHW");
         }
         if (shape.get(0) == 1) {
             throw new UnsupportedOperationException("Grayscale image is not supported");
+        } else if (shape.get(0) != 3) {
+            if (shape.get(2) == 3) {
```
zachgk, line 97: Based on the error message, shouldn't the case of `shape.get(2) == 1` also be handled?

android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
zachgk: I would agree that we should reuse. The implication of having a variable named finalWidth is that there is also a non-final width and using it would not be right. But here, there are multiple paths to set the width, but it is not really changing. It may even make sense to change BufferedImageFactory as well. But for here, you can use width and height

JohnDoll2023: Ah I see why it is that way in BufferedImageFactory. The build will fail with error `local variables referenced from a lambda expression must be final or effectively final`. I will put `width` into a final variable and leave height just as height.

android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
JohnDoll2023: Added a check for this

android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
lanking520: Since you are allowing HWC data input, you should handle this case correctly. The data order is changed for HWC as mentioned above. It is ordered to be channel last where RGB color are come together. In here we assume data is CHW and that's why we always add the area size to skip a channel

PR #1833, opened 2022-07-26
Updated image factories for HWC and CHW

JohnDoll2023:
## Description ##

Both image factories now handle HWC and CHW formats

- Transpose is not available for all engines so we must handle the different formats without transposing to a single format type
- Test on the apps and within DJL repo

### DIFF
--- a/android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
+++ b/android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
@@ -110,7 +110,6 @@ public class BitmapImageFactory extends ImageFactory {
         int height = (int) shape.get(0);
         int width = (int) shape.get(1);
         Bitmap bitmap = Bitmap.createBitmap(width, height, Bitmap.Config.ARGB_8888);
-        int imageArea = height * width;
         for (int h = 0; h < height; ++h) {
             for (int w = 0; w < width; ++w) {
                 int pos = (h * width + w) * 3;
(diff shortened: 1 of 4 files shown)


android/core/src/main/java/ai/djl/android/core/BitmapImageFactory.java
```diff
@@ -91,20 +91,34 @@ public Image fromNDArray(NDArray array) {
...
         int[] raw = array.toType(DataType.UINT8, false).toUint8Array();
+        int imageArea = height * width;
         for (int h = 0; h < height; ++h) {
             for (int w = 0; w < width; ++w) {
-                int pos = (h * width + w) * 3;
+                int pos = (h * width + w);
                 int red = raw[pos];
-                int green = raw[pos + 1];
-                int blue = raw[pos + 2];
+                int green = raw[imageArea + pos];
```
frankfliu, line 120: Why you change here? the original implementation was correct

api/src/main/java/ai/djl/modality/cv/BufferedImageFactory.java
```diff
@@ -91,8 +91,24 @@ public Image fromNDArray(NDArray array) {
...
+            int height = (int) shape.get(1);
+            int width = (int) shape.get(2);
+            BufferedImage image = new BufferedImage(width, height, BufferedImage.TYPE_INT_RGB);
+            int[] raw = array.toType(DataType.UINT8, false).toUint8Array();
+            int[] pixels = new int[width * height];
+            int imageArea = height * width;
+            for (int h = 0; h < height; ++h) {
+                for (int w = 0; w < width; ++w) {
+                    int index = h * width + w;
+                    int pos = index;
```
frankfliu, line 103: redundant pos variable

lanking520 (approved):

PR #1844, opened 2022-07-29
[WIP] add DJL Android ONNXRuntime support

wxm2018:
## Description ##

- Add DJL Android support for the ONNXRuntime engine to enable it to use onnx format models.- 
- But now it is [WIP].


android/core/build.gradle
```diff
@@ -97,11 +97,19 @@ afterEvaluate {
     }
 }
 
+repositories {
```
lanking520, line 100: This can be removed

android/onnxruntime-native/README.md
```diff
@@ -0,0 +1,90 @@
+
+# ONNXRuntime Android Build Procedure
```
lanking520, line 2: I think the installation instruction is generally applicable to all versions. You can migrate the build instruction to Android folder and refer a link from README

android/onnxruntime-native/build.gradle
```diff
@@ -0,0 +1,201 @@
...
+
+
+
+
+
+
+
+
+
+//apply plugin: 'java-library'
```
lanking520, line 141: This is the setup for transient onnxruntime, @frankfliu what do you think?

lanking520 (changes requested):

android/onnxruntime-native/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+This procedure is meant only for developers who want to build PyTorch native binaries from source, this is not intended for regular users.
```
frankfliu, line 4: This section is copied from PyTorch module, we need write different one for ONNX

android/build.gradle
```diff
@@ -7,6 +7,7 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:4.2.2'
+        classpath 'com.github.kezong:fat-aar:1.3.8'
```
frankfliu, line 10: fat-aar may not work as you expected.
engine.jar bundle in aar won't work, it need to be dexed first.

I think we should just publish a pom only package as normal java project.

android/core/build.gradle
frankfliu, line 100: We should not need this.

We can use mavenLocal():
1. cd onnxruntime-native
2. gradle pTML

lanking520:
As discussed offline. Let's just rollback to the previous trasient dependency version

android/onnxruntime-native/build.gradle
frankfliu: Please try my PR: https://github.com/deepjavalibrary/djl/pull/1866

Your android project should be able to simply include:
```
implementation "ai.djl.android:onnxruntime:0.19.0-SNAPSHOT"
```

frankfliu (approved):

lanking520 (approved):

PR #1847, opened 2022-07-31
Add MobileNetV2 into ModelZoo and corresponding Test Class MobileNetV…

warthecatalyst:
Add a new class MobileNetV2 into ModelZoo and its corresponding test class MobileNetV2Test into intergration. I think it's almost complete except for some minor problems.It seems like my Block MobileNetV2 will fail when it comes to trainer.step(). The following code is when it goes wrong.

                int batchSize = 1;
                Shape inputShape = new Shape(batchSize, 3, 224, 224);
                trainer.initialize(inputShape);
                NDManager manager = trainer.getManager();
                NDArray input = manager.ones(inputShape);
                NDArray label = manager.ones(new Shape(batchSize, 1));
                Batch batch =
                        new Batch(
                                manager.newSubManager(),
                                new NDList(input),
                                new NDList(label),
                                batchSize,
                                Batchifier.STACK,
                                Batchifier.STACK,
                                0,
                                0);
                PairList<String, Parameter> parameters = mobilenet.getParameters();
                EasyTrain.trainBatch(trainer, batch);
                trainer.step();


model-zoo/src/main/java/ai/djl/basicmodelzoo/cv/classification/MobileNetV2.java
```diff
@@ -0,0 +1,410 @@
...
+    public static Block makeStage(
+            int repeat,
+            int inputChannels,
+            int outputChannels,
+            int stride,
+            int t,
+            float batchNormMomentum) {
+        SequentialBlock layers = new SequentialBlock();
+        layers.add(linearBottleNeck(inputChannels, outputChannels, stride, t, batchNormMomentum));
+        while (repeat - 1 > 0) {
```
zachgk, line 126: Can this be changed to a for loop? They are much easier to read then a while loop

model-zoo/src/main/java/ai/djl/basicmodelzoo/cv/classification/MobileNetV2.java
```diff
@@ -0,0 +1,410 @@
...
+                                .setFilters(outputChannels)
+                                .setKernelShape(new Shape(1, 1))
+                                .optBias(false)
+                                .build())
+                .add(BatchNorm.builder().optEpsilon(2E-5f).optMomentum(batchNormMomentum).build());
+
+        // if dimMatch,then add x
+        if (stride == 1 && inputChannels == outputChannels) {
+            block =
+                    new SequentialBlock()
```
zachgk, line 92: You don't need to wrap this in a SequentialBlock here. You can just directly return the ParallelBlock

model-zoo/src/main/java/ai/djl/basicmodelzoo/cv/classification/MobileNetV2.java
```diff
@@ -0,0 +1,410 @@
...
+                                    .build())
+                    .add(
+                            BatchNorm.builder()
+                                    .optEpsilon(2E-5f)
+                                    .optMomentum(builder.batchNormMomentum)
+                                    .build())
+                    .add(Activation.reluBlock());
+        }
+
+        Block bottleNeck1 =
```
zachgk, line 159: Rather than constructing each bottleNeck individually, can you create them by a for loop? Then, you can store them in an array and use `SequentialBlock.addAll()` to add all of them to `mobilenet` at once

model-zoo/src/main/java/ai/djl/basicmodelzoo/cv/classification/MobileNetV2.java
```diff
@@ -0,0 +1,410 @@
...
+                .add(bottleNeck1)
+                .add(bottleNeck2)
+                .add(bottleNeck3)
+                .add(bottleNeck4)
+                .add(bottleNeck5)
+                .add(bottleNeck6)
+                .add(bottleNeck7)
+                .add(conv1)
+                .add(Pool.globalAvgPool2dBlock())
+                .add(
```
zachgk, line 250: Here and on 265, you can use `SequentialBlock.addSingleton`. This lets you simplify your code a bit, but also quite importantly will reduce all of these levels of indentations

zachgk:
Also, is this performing better then MobileNetV1 during tests? It should be because V2 is an improvement on V1. Then, you should replace the V1 in DJL Zero

model-zoo/src/main/java/ai/djl/basicmodelzoo/cv/classification/MobileNetV2.java
warthecatalyst: got it!

model-zoo/src/main/java/ai/djl/basicmodelzoo/cv/classification/MobileNetV2.java
warthecatalyst: yeah,I realize that

model-zoo/src/main/java/ai/djl/basicmodelzoo/cv/classification/MobileNetV2.java
warthecatalyst: Yeah... At first I thought that we need to add a relu block at last, but I go through the paper again and find that a relublock is not needed...

zachgk:
It looks good to merge. Can you rebase it onto master?

warthecatalyst:
yeah



---- Replied Message ----
| From | Zach ***@***.***> |
| Date | 08/04/2022 09:38 |
| To | ***@***.***> |
| Cc | ***@***.******@***.***> |
| Subject | Re: [deepjavalibrary/djl] Add MobileNetV2 into ModelZoo and corresponding Test Class MobileNetV… (PR #1847) |

It looks good to merge. Can you rebase it onto master?

PR #1852, opened 2022-08-01
Catch security exception in get environment call

xyang16:
## Description ##

This PR is to fix https://github.com/deepjavalibrary/djl/discussions/1848

The customer run into the get environment security issue. This PR put the environment call into a utility function and catch the exception.

### DIFF
--- a/api/src/main/java/ai/djl/util/Utils.java
+++ b/api/src/main/java/ai/djl/util/Utils.java
@@ -16,6 +16,7 @@ import ai.djl.ndarray.NDArray;
 import ai.djl.nn.Parameter;
 
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.BufferedInputStream;
 import java.io.ByteArrayOutputStream;
@@ -30,7 +31,9 @@ import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Scanner;
 import java.util.regex.Matcher;
@@ -39,6 +42,8 @@ import java.util.stream.Collectors;
 
 /** A class containing utility methods. */
 public final class Utils {
+
+    private static final Logger logger = LoggerFactory.getLogger(Utils.class);
 
     private Utils() {}
 
@@ -326,7 +331,7 @@ public final class Utils {
     public static Path getEngineCacheDir() {
         String cacheDir = System.getProperty("ENGINE_CACHE_DIR");
         if (cacheDir == null || cacheDir.isEmpty()) {
-            cacheDir = System.getenv("ENGINE_CACHE_DIR");
+            cacheDir = Utils.getenv("ENGINE_CACHE_DIR");
             if (cacheDir == null || cacheDir.isEmpty()) {
                 return getCacheDir();
             }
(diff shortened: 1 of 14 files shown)


frankfliu (approved):

api/src/main/java/ai/djl/util/Utils.java
```diff
@@ -377,4 +382,35 @@ public static Path getNestedModelDir(Path modelDir) {
...
+     * @param name the name of the environment variable
+     * @return the string value of the variable, or {@code null} if the variable is not defined in
+     *     the system environment or security manager doesn't allow access to the environment
+     *     variable
+     */
+    public static String getenv(String name) {
+        try {
+            return System.getenv(name);
+        } catch (SecurityException e) {
+            logger.error("Security manager doesn't allow access to the environment variable");
```
frankfliu, line 398:
```suggestion
            logger.warn("Security manager doesn't allow access to the environment variable");
```

api/src/main/java/ai/djl/util/Utils.java
```diff
@@ -377,4 +382,35 @@ public static Path getNestedModelDir(Path modelDir) {
...
+
+    /**
+     * Gets the value of the specified environment variable.
+     *
+     * @param name the name of the environment variable
+     * @return the string value of the variable, or {@code null} if the variable is not defined in
+     *     the system environment or security manager doesn't allow access to the environment
+     *     variable
+     */
+    public static String getenv(String name) {
```
frankfliu, line 394: I think it's better have an overload function:
```
public static String getenv(String name, String def) {
```

frankfliu (commented):
You can merge this PR create a follow up PR to improve the API.

api/src/main/java/ai/djl/util/Utils.java
xyang16: Changed.

api/src/main/java/ai/djl/util/Utils.java
xyang16: Added

api/src/main/java/ai/djl/util/Utils.java
```diff
@@ -377,4 +382,54 @@ public static Path getNestedModelDir(Path modelDir) {
...
+        try {
```
frankfliu, line 395: return getenv(name, null);

api/src/main/java/ai/djl/util/Utils.java
```diff
@@ -377,4 +382,49 @@ public static Path getNestedModelDir(Path modelDir) {
...
+     *
+     * @return the environment as a map of variable names to values
+     */
+    public static Map<String, String> getenv() {
+        try {
+            return System.getenv();
+        } catch (SecurityException e) {
+            logger.warn("Security manager doesn't allow access to the environment variable");
+        }
+        return Collections.unmodifiableMap(new HashMap<>());
```
frankfliu, line 428:
```suggestion
        return Collections.emptyMap();
```

api/src/main/java/ai/djl/util/Utils.java
```diff
@@ -377,4 +382,49 @@ public static Path getNestedModelDir(Path modelDir) {
         }
         return modelDir.toAbsolutePath();
     }
+
+    /**
+     * Gets the value of the specified environment variable.
+     *
+     * @param name the name of the environment variable
+     * @param def a default value
+     * @return the string value of the variable, or {@code null} if the variable is not defined in
```
frankfliu, line 391:
```suggestion
     * @return the string value of the variable, or {@code def} if the variable is not defined in
```

api/src/main/java/ai/djl/util/Utils.java
```diff
@@ -30,7 +31,9 @@
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
```
frankfliu, line 34:
```suggestion
```

xyang16: Done

api/src/main/java/ai/djl/util/Utils.java
xyang16: Done

api/src/main/java/ai/djl/util/Utils.java
xyang16: Done

api/src/main/java/ai/djl/util/Utils.java
xyang16: Done

PR #1649, opened 2022-05-13
Add support of `take` on MXNet engine

KexinFeng:
This pr follows https://github.com/deepjavalibrary/djl/pull/1627

[`take` function in MXNet ](https://mxnet.apache.org/versions/1.6.0/api/r/docs/api/mx.symbol.take.html?highlight=take) is defined differently from numpy or pytorch. It is implemented by `flatten` and `take` in MXNet.

PR #1869, opened 2022-08-04
Bulk Batch creation

patins1:
This pull requests optimizes Batch creation for ArrayDataset when using StackBatchifier .
My DJL application learns the data now in 58550ms  rather than 79843ms so 36% faster!
For applications not using range-based indexing but indices-based indexing, my DJL application would run in 66341ms , so still 20% faster. This is true for PyTorch, i get a test failure for MxNet, so this solution maybe restricted to PyTorch

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/index/full/NDIndexFullPick.java
+++ b/api/src/main/java/ai/djl/ndarray/index/full/NDIndexFullPick.java
@@ -17,7 +17,6 @@ import ai.djl.ndarray.index.NDIndex;
 import ai.djl.ndarray.index.dim.NDIndexAll;
 import ai.djl.ndarray.index.dim.NDIndexElement;
 import ai.djl.ndarray.index.dim.NDIndexPick;
-import ai.djl.ndarray.index.dim.NDIndexTake;
 import ai.djl.ndarray.types.Shape;
 
 import java.util.Optional;
--- a/api/src/main/java/ai/djl/ndarray/index/full/NDIndexFullPick.java
+++ b/api/src/main/java/ai/djl/ndarray/index/full/NDIndexFullPick.java
@@ -27,7 +27,6 @@ public final class NDIndexFullPick {
 
     private NDArray indices;
     private int axis;
-    private boolean indexTake;
 
     /**
      * Constructs a new {@link NDIndexFullPick}.
@@ -73,7 +72,6 @@ public final class NDIndexFullPick {
                             "Only rank-1 indexing array is supported for pick");
                 }
                 fullPick = new NDIndexFullPick(indexElem, axis);
-                fullPick.setIndexTake(true);
             } else {
                 // Invalid dim for fullPick
                 return Optional.empty();
@@ -98,23 +96,5 @@ public final class NDIndexFullPick {
      */
     public int getAxis() {
         return axis;
-    }
-
-    /**
-     * Returns if a "take" rather than a "pick" operation is intended.
-     *
-     * @return <code>true</code> for "take"
-     */
-    public boolean isIndexTake() {
-        return indexTake;
-    }
-
-    /**
-     * Sets if a "take" rather than a "pick" operation is intended.
-     *
-     * @param indexTake <code>true</code> for "take"
-     */
-    public void setIndexTake(boolean indexTake) {
-        this.indexTake = indexTake;
     }
 }
--- a/api/src/main/java/ai/djl/ndarray/index/full/NDIndexFullPick.java
+++ b/api/src/main/java/ai/djl/ndarray/index/full/NDIndexFullPick.java
@@ -27,6 +27,7 @@ public final class NDIndexFullPick {
 
     private NDArray indices;
     private int axis;
+    private boolean indexTake;
 
     /**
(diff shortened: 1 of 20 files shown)


patins1:
The problems with MxNet is fixed by using "take" operation instead of "pick".
TrainMnist executes now 4 times faster for MxNet .

lanking520:
@KexinFeng can you help to check on the NDIndex part changes?

KexinFeng:
@patins1 It looks like `BulkDataIterable` is not covered by any unit test. You mentioned that using `BulkDataIterable` is more efficient. Could you add a unit test that covers this class you added?

KexinFeng:
> The problems with MxNet is fixed by using "take" operation instead of "pick".
> TrainMnist executes now 4 times faster for MxNet .

Could you integrate it with the existing `take` function in [PR](https://github.com/deepjavalibrary/djl/pull/1649)?

api/src/main/java/ai/djl/ndarray/index/full/NDIndexFullPick.java
```diff
@@ -72,6 +73,7 @@ public static Optional<NDIndexFullPick> fromIndex(NDIndex index, Shape target) {
                             "Only rank-1 indexing array is supported for pick");
                 }
                 fullPick = new NDIndexFullPick(indexElem, axis);
+                fullPick.setIndexTake(true);
```
KexinFeng, line 76: Here, I tried to use take for MXNet engine too see https://github.com/deepjavalibrary/djl/pull/1802, and the PR tests all passed. But it caused the https://github.com/deepjavalibrary/djl/issues/1800. Here in your edition, setting fullPick.setIndexTake(true); will trigger this problem again.

KexinFeng (changes requested):
If you could add a unit test that covers `BulkDataIterable` and that shows the necessity of using `take` on MXNet engine, it'd be great.

api/src/main/java/ai/djl/ndarray/index/full/NDIndexFullPick.java
patins1: `NDArray.take(NDArray)` returns a shape that is not useful for the functionality of this pull request:

`Returns a partial NDArray pointed by index according to linear indexing, and the output is of the same shape as index.`

Given `ndArray` of shape e.g. `(a,b,c,d)` and `index` of shape `(e) `, then I assumed initially that `ndArray.take(index)` would return an array of shape` (e,b,c,d)` .  Therefore I restyled my initial implementation, using NDIndexFullTake .  If anybody has a better idea, let me know, at least the test cases go green now.

KexinFeng:
@patins1 Thanks for dealing with this issue in such a timely manner! But it looks like the unittest that covers the new file api/src/main/java/ai/djl/training/dataset/BulkDataIterable.java is still missing. You mentioned you have tested it, which display efficiency increase. Could you add it into the proper unit test file? 

Also you mentioned 
> The problems with MxNet is fixed by using "take" operation instead of "pick".
> TrainMnist executes now 4 times faster for MxNet .

It'd be better to have unit test for this too, which will prevent future edition from breaking this fix.

Thanks!

patins1:
Tests added

KexinFeng:
@siddvenk Hi Siddarth, in this pr, we have changed the definition of get(NDArray index) from pick to take, and have given the warning. This will affect the result you mentioned in https://github.com/deepjavalibrary/djl/issues/1800. See the test in commit "add index test and clean code" 8ffac00e6584c96500e6ac96ba35cc45513f1f4e. Pick can still be used though by addPickDim().
This change is for the purpose of making it consistent with numpy and pytorch engine, also here it is mor efficient.
Should we update the relevant part mentioned in https://github.com/deepjavalibrary/djl/issues/1800?

KexinFeng (approved):

zachgk (approved):
There is one small problem I noticed, but otherwise it looks good. I think we should just merge it for now and we can fix it in a later PR

KexinFeng:
> @siddvenk Hi Siddarth, in this pr, we have changed the definition of get(NDArray index) from pick to take, and have given the warning. This will affect the result you mentioned in #1800. See the test in commit "add index test and clean code" [8ffac00](https://github.com/deepjavalibrary/djl/commit/8ffac00e6584c96500e6ac96ba35cc45513f1f4e). Pick can still be used though by addPickDim(). This change is for the purpose of making it consistent with numpy and pytorch engine, also it is more efficient as shown here. Should we update the relevant part of the _Dive into Deep Learning for Java_ book mentioned in #1800?

KexinFeng:
> @siddvenk Hi Siddarth, in this pr, we have changed the definition of get(NDArray index) from pick to take, and have given the warning. This will affect the result you mentioned in #1800. See the test in commit "add index test and clean code" [8ffac00](https://github.com/deepjavalibrary/djl/commit/8ffac00e6584c96500e6ac96ba35cc45513f1f4e). Pick can still be used though by addPickDim(). This change is for the purpose of making it consistent with numpy and pytorch engine, also it is more efficient as shown here. Should we update the relevant part of the _Dive into Deep Learning for Java_ book mentioned in #1800?

PR #1871, opened 2022-08-04
[android] Adds formatJava task for android project

frankfliu:
Change-Id: If15efb1ad2a744b1bfadb7bf89a09a265a81082b

## Description ##

### DIFF
--- a/tools/gradle/android-formatter.gradle
+++ b/tools/gradle/android-formatter.gradle
@@ -1,0 +1,40 @@
+buildscript {
+    repositories {
+        mavenCentral()
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
+    }
+    dependencies {
+        classpath 'com.google.googlejavaformat:google-java-format:1.15.0'
+    }
+}
+
+apply plugin: JavaFormatterPlugin
+
+import com.google.googlejavaformat.java.Formatter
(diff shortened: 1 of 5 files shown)


tools/gradle/android-formatter.gradle
```diff
@@ -0,0 +1,40 @@
...
+
+apply plugin: JavaFormatterPlugin
+
+import com.google.googlejavaformat.java.Formatter
+import com.google.googlejavaformat.java.ImportOrderer
+import com.google.googlejavaformat.java.JavaFormatterOptions
+import com.google.googlejavaformat.java.Main
+import com.google.googlejavaformat.java.RemoveUnusedImports
+
+class JavaFormatterPlugin implements Plugin<Project> {
```
zachgk, line 21: Shouldn't we also add the formatter verification?

zachgk (approved):

tools/gradle/android-formatter.gradle
frankfliu: It's a bit tricky to do that:
1. Android was not part of continuous build, code changes won't very immediately verified at PR time
2. We only run connectAndroidTest nightly, it's not really a check task, not sure if we want cAT depends on verify
3. I'm not able to scan sourceSets for Android project, it currently include add .java file in the project folder, if we added any generated .java file during the build, verifyJava may fail

Issue #1886, opened 2022-08-10
pytorch-engine:0.18.0 causes memory leak when using NDManager.newBaseManager()


925781609:
## Description
1. When using  pytorch-engine:0.18.0 `NDManager.newBaseManager()` creates a `PtNDManager`, it will call `ai.djl.pytorch.engine.PtNDManager#newSubManager`, and execute: 
```java
   PtNDManager manager = new PtNDManager(this, device);
   attachUncappedInternal(manager.uid, manager);
   return manager;
```
2. Method `attachUncappedInternal` is implemented by `BaseNDManager` and attaches the created PtNDManager to its field `resources`. 
```java
    resources.put(resourceId, resource);
```

3. The created PtNDManger will never be released even it is closed. 
```java
   public void close() {
        if (!closed.getAndSet(true)) {
             // ignore some code
            parent.detachInternal(uid);
            resources.clear();
            tempResources.clear();
        }
    }

```
The `parent` is `PtNDManager$SystemManager` and  parent's `detachInternal` does nothing.
```java
@Override
 public void detachInternal(String resourceId) {}
```

So in the end, the created PtNDManger will not be sweeped by JVM GC.

4. When downgrade  pytorch-engine to version 0.17.0, the problem is solved. Because the  `newSubManager` calls `PtNDManager$SystemManger#attachInternal`. `PtNDManager$SystemManger#attachInternal` does nothing.
```java
   PtNDManager manager = new PtNDManager(this, device);
    attachInternal(manager.uid, manager);
    return manager;
```

```java
 @Override
  public void attachInternal(String resourceId, AutoCloseable resource) {}
```

### Expected Behavior
The `SystemManager` will not attach the created PtNDManger to its field `resources` or release PtNDManger when it is closed.

### Error Message

<img width="701" alt="image" src="https://user-images.githubusercontent.com/20756551/183803585-2737751d-6614-470c-80b7-311ddd478457.png">

## How to Reproduce?
1. use pytorch-engine version 0.18.0
2.  execute the code below as many times as possible and will cause OOM eventually.
```java
try (NDManager manager = NDManager.newBaseManager(Device.cpu())) {
    // do something here
}
```
3. maven dependencies
```xml
        <dependency>
            <groupId>ai.djl.pytorch</groupId>
            <artifactId>pytorch-engine</artifactId>
            <version>0.18.0</version>
            <exclusions>
                <exclusion>
                    <artifactId>jna</artifactId>
[...]

PR #1887, opened 2022-08-10
Fixes memory leak when using NDManager.newBaseManager()

925781609:
…DManager (#1886)

## Description ##

When using pytorch-engine:0.18.0 `NDManager.newBaseManager()`, it will call `attachUncappedInternal` to attach the created PtNDManger to PtNDManager$SystemManager's `resources` field, but never detach it, thus cause memory leak.

By implementing `attachUncappedInternal` in PtNDManager$SystemManager without doing anything, avoid PtNDManager cannot be sweeped by JVM


lanking520 (approved):

frankfliu:
@925781609 

Thank you so much for reporting and fixing this memory leak issue.

PR #1888, opened 2022-08-10
Creates a SystemNDManager interface

### LINKED ISSUE #1886: pytorch-engine:0.18.0 causes memory leak when using NDManager.newBaseManager()

zachgk:
fixes #1886

This creates an interface for SystemNDManagers. That way, the behavior of
skipping over various functions can be moved to the BaseNDManager instead of
each individual SystemManager

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
+++ b/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
@@ -315,6 +315,9 @@ public abstract class BaseNDManager implements NDManager {
     /** {@inheritDoc} */
     @Override
     public synchronized void attachInternal(String resourceId, AutoCloseable resource) {
+        if (this instanceof SystemNDManager) {
+            return;
+        }
         if (capped.get()) {
             throw new IllegalStateException("NDManager is capped for addition of resources.");
         }
@@ -324,6 +327,9 @@ public abstract class BaseNDManager implements NDManager {
     /** {@inheritDoc} */
     @Override
     public synchronized void attachUncappedInternal(String resourceId, AutoCloseable resource) {
+        if (this instanceof SystemNDManager) {
+            return;
+        }
         if (closed.get()) {
             throw new IllegalStateException("NDManager has been closed already.");
         }
@@ -349,6 +355,9 @@ public abstract class BaseNDManager implements NDManager {
     @Override
     public void tempAttachInternal(
             NDManager originalManager, String resourceId, NDResource resource) {
+        if (this instanceof SystemNDManager) {
+            return;
+        }
         if (closed.get()) {
             throw new IllegalStateException("NDManager has been closed already.");
         }
@@ -358,6 +367,9 @@ public abstract class BaseNDManager implements NDManager {
     /** {@inheritDoc} */
     @Override
     public synchronized void detachInternal(String resourceId) {
+        if (this instanceof SystemNDManager) {
+            return;
+        }
         if (closed.get()) {
             // This may happen in the middle of BaseNDManager.close()
             return;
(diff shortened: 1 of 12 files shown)


frankfliu:
I'm actually not sure if this is better than override in each SystemNDManager.

In each methods to check instanceof seems odd to me.

zachgk:
It may be a bit odd, but it makes it far easier to implement the NDManager API and to do it correctly. I noticed at least a few engines that were missing some overloads when I was doing this. So, it seems like trying to remember every overload in every engine may be too error prone to rely on

frankfliu:
Do we need update lightGBM?

frankfliu (approved):

PR #1909, opened 2022-08-16
add nightly aarch64 and gpu builds

siddvenk:
Add nightly tests for aarch64, cuda102, cuda113. These mirror the previous tests in codebuild, and have been updated to work with pytorch 1.12.1.

Here's the latest run with the new jobs all working https://github.com/deepjavalibrary/djl/actions/runs/2870780589

~With pytorch 1.12.1, I noticed something strange with the binaries. This was the error I was seeing when testing pytorch gpu builds (across all cuda versions):~

~`
Caused by: java.lang.UnsatisfiedLinkError: /root/.djl.ai/pytorch/1.12.1-cu116-linux-x86_64/libtorch_cuda_cu.so: libcublas-2854e16e.so.11: cannot open shared object file: No such file or directory
`~

~Seems like the `libtorch_cuda_cu.so` binary we published is expecting a dependent binary `libcublas-2854e16e.so.11`, which doesn't exist because the real library is `libcublas.so.11`. I'm not sure where the suffix `-2854e16e` is coming from.~

~For now I have symlinked it to work.~

Pytorch is working now

### DIFF
--- a/.github/workflows/nightly_publish.yml
+++ b/.github/workflows/nightly_publish.yml
@@ -88,10 +88,119 @@ jobs:
       - name: Test with Gradle
         run: ./gradlew :integration:test "-Dai.djl.default_engine=TensorFlow"
 
+  test-aarch64:
+    if: github.repository == 'deepjavalibrary/djl'
+    runs-on: [ self-hosted, aarch64 ]
+    container: amazonlinux:2
+    timeout-minutes: 30
+    needs: create-runners
+    steps:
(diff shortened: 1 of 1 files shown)


siddvenk:
https://github.com/deepjavalibrary/djl/pull/1916 will fix the issue with the libcublas binary - will wait for that to merge and then update this pr

lanking520 (approved):

PR #1926, opened 2022-08-18
[tokenizers] load truncation and padding settings from options, updat…

siddvenk:
[tokenizers] load truncation and padding settings from options, update apis

## Description ##
- loading truncation/padding settings from options when creating new tokenizer instance
- Updated APIs for setting truncation/padding post tokenizer creation. Of note:
  - maxLength parameter is shared by both truncation and padding, and settable by its own method
  - No support for stride in this implementation
 - Updated unit test coverage - used the supported combinations from https://huggingface.co/docs/transformers/v4.20.1/en/pad_truncation to guide test cases

### DIFF
--- a/extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
+++ b/extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
@@ -19,6 +19,9 @@ import ai.djl.modality.nlp.preprocess.Tokenizer;
 import ai.djl.util.NativeResource;
 import ai.djl.util.Utils;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
@@ -33,7 +36,19 @@ import java.util.Map;
  */
 public final class HuggingFaceTokenizer extends NativeResource<Long> implements Tokenizer {
 
+    private static final Logger logger = LoggerFactory.getLogger(HuggingFaceTokenizer.class);
+    private static final long DEFAULT_TRUNCATION_LENGTH = 512L;
+    private static final long DEFAULT_PADDING_LENGTH = 0L;
+    private static final long DEFAULT_STRIDE = 0L;
+    private static final long DEFAULT_PAD_TO_MULTIPLE_OF = 0L;
+
     private boolean addSpecialTokens;
+    private TruncationStrategy truncationStrategy;
+    private PaddingStrategy paddingStrategy;
+    private long truncationLength;
+    private long paddingLength;
+    private long stride;
+    private long padToMultipleOf;
 
     private HuggingFaceTokenizer(long handle, Map<String, String> options) {
         super(handle);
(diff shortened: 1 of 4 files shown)


extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
```diff
@@ -305,60 +306,98 @@ public String decode(long[] ids) {
...
+     * @param truncation whether to enable truncation
+     * @param padding whether to enable padding
+     * @param maxLength maximum length of sequences to truncate and/or pad to
      */
-    public void setTruncationAndPadding(
-            TruncationStrategy truncationStrategy,
-            PaddingStrategy paddingStrategy,
-            long maxLength) {
-        setTruncationAndPadding(truncationStrategy, paddingStrategy, maxLength, 0, 0);
+    public void setTruncationAndPadding(boolean truncation, boolean padding, long maxLength) {
```
zachgk, line 313: This looks like you are really trying to have two separate actions for truncation and padding here. Maybe split it into two functions?

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
```diff
@@ -305,60 +306,98 @@ public String decode(long[] ids) {
...
+     * @param padding the type of padding behavior to use
+     * @param maxLength maximum length of sequences to truncate and/or pad to
      */
-    public void setTruncationAndPadding(
-            TruncationStrategy truncationStrategy,
-            PaddingStrategy paddingStrategy,
-            long maxLength,
-            long stride) {
-        setTruncationAndPadding(truncationStrategy, paddingStrategy, maxLength, stride, 0);
+    public void setTruncationAndPadding(String truncation, String padding, long maxLength) {
```
zachgk, line 324: Is there a reason for the intermediate string format for truncation here? Right now, it goes from `Boolean -> String -> TruncationStrategy`. I recommend picking one "standardized" format which is probably the typed `TruncationStrategy`. Then, if you have alternative strategies (boolean, maybe String) then it only goes from the alternative to the standardized format.

Beyond that, I would personally lean towards supporting the enum over a String format. It is easier and more reliable to use from Java code and easier to document. Speaking of which, it may be worth adding some javadoc comments to the enum to explain the different truncation and padding strategies

siddvenk: Yea this is something I'm debating as well - prior to this we were leaning on the enum option where users would pass in an enum value to indicate the behavior. The overall motivation for this change is to adhere more to the fast tokenizer interface, which lets users pass in a boolean, String, or enum https://huggingface.co/transformers/v4.4.2/main_classes/tokenizer.html. 

On our side, I think it makes sense to support the boolean parameter, and then one of either enum or string. I'd prefer enum for the same reasons you pointed out. 

The two options I see are to expose `boolean and enum` methods, or `boolean and string` methods, of which i prefer the former.

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
siddvenk: I think keeping it as a single method is more in line with the interface exposed by transformers for fast tokenizers https://huggingface.co/docs/transformers/v4.20.1/en/main_classes/tokenizer#transformers.PreTrainedTokenizerFast.set_truncation_and_padding. 

Though, based on which implementation the user chooses they can modify truncation and padding independently. Here's the rough interface provided at each level in hugging face:
- Transformers FastTokenizer: 1 method to setTruncationAndPadding https://huggingface.co/docs/transformers/v4.20.1/en/main_classes/tokenizer#transformers.PreTrainedTokenizerFast.set_truncation_and_padding - i think the primary motivation for this is that since truncation and padding can share some settings (max_length), making sure they are set together prevents the user from making mistakes
- Tokenizer python bindings: independent methods for truncation and padding https://github.com/huggingface/tokenizers/blob/main/bindings/python/py_src/tokenizers/implementations/base_tokenizer.py
- Tokenizer Rust: independent methods for truncation and padding https://github.com/huggingface/tokenizers/blob/main/bindings/python/src/tokenizer.rs

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
```diff
@@ -303,86 +338,118 @@ public String decode(long[] ids) {
...
+            updateTruncationAndPadding();
+        }
+    }
+
+    /** Enables truncation to only truncate the second item. */
+    public void setTruncateSecondOnly() {
+        if (truncation != TruncationStrategy.ONLY_SECOND) {
+            truncation = TruncationStrategy.ONLY_SECOND;
+            updateTruncationAndPadding();
+        }
```
siddvenk, line 378: Should we also provide `maxLength` as an argument? Right now there doesn't seem like an way to use these strategies with custom lengths via the API.

frankfliu (approved):

PR #1940, opened 2022-08-23
[tokenizer] Use builder for HuggingfaceTokenizer

frankfliu:
1. Add builder for HuggingfaceTokenizer
2. Make HuggingfaceTokenizer immutable

Change-Id: I84724b83051cb2079046b49893d96485d8fdcd3b

## Description ##

### DIFF
--- a/extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
+++ b/extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
@@ -30,6 +30,7 @@ import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * {@code HuggingFaceTokenizer} is a Huggingface tokenizer implementation of the {@link Tokenizer}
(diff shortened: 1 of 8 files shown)


extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
```diff
@@ -535,4 +480,152 @@ static PaddingStrategy fromValue(String value) {
...
+        public HuggingFaceTokenizer build() throws IOException {
+            String tokenizerName = options.get("tokenizer");
+            if (tokenizerName != null) {
+                return HuggingFaceTokenizer.newInstance(tokenizerName, options);
+            }
+            if (tokenizerPath == null) {
+                throw new IllegalArgumentException("Missing tokenizer path.");
+            }
+            return HuggingFaceTokenizer.newInstance(tokenizerPath, options);
+        }
```
siddvenk, line 629: I think it is unlikely a user would do this, but whether they use the builder or newInstance to create a tokenizer they could set maxLength and/or padToMultiple of without explicitly enabling padding/truncation. I think there are 3 things we could do in that case:

1. Nothing, leave as is
2. Log a warning letting the user know no truncation/padding will be used even though it looks like they are trying to use it
3. Silently enable truncation/padding in this case. 

I'd prefer option 2 or 1, what do you think?

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
```diff
@@ -535,4 +480,152 @@ static PaddingStrategy fromValue(String value) {
...
+        /**
+         * Sets maxLength for padding and truncation.
+         *
+         * @param maxLength the length to truncate and/or pad sequences to
+         * @return this builder
+         */
+        public Builder optMaxLength(int maxLength) {
+            options.put("maxLength", String.valueOf(maxLength));
+            return this;
+        }
```
siddvenk, line 590: I think this is unlikely, but a user could pass in a length of 0 here by accident - should we treat 0 similar to how we treat -1 currently (unset, the default returned from our JNI)?

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
frankfliu: I think option 2 is preferable.
But setting maxLength/padToMultipleOf when padding and truncation is disabled is harmless. I don't feel strong to print a warning to user. User might want to always set the `maxLength` and switch on/off padding

siddvenk: I think this is fine as is. The situation I was thinking of is when neither truncation or padding are explicitly set, but user set the maxlength or pad to multiple of. It's harmless as you said, and i think it's fair to assume in this case the user knows what they are doing.

siddvenk (approved):

PR #1969, opened 2022-08-26
[converter] Handles models that requires toke_type_ids case

frankfliu:
Change-Id: I130d45c058d74d0ff30103fb5c02a4d0aa5f59e8

## Description ##


extensions/tokenizers/src/main/java/ai/djl/huggingface/translator/QuestionAnsweringTranslator.java
```diff
@@ -73,8 +73,8 @@ public String processOutput(TranslatorContext ctx, NDList list) {
         NDArray endLogits = list.get(1).duplicate();
 
         // exclude <CLS>, TODO: exclude impossible ids properly and handle max answer length
-        startLogits.set(new NDIndex(0), 0);
-        endLogits.set(new NDIndex(0), 0);
+        startLogits.set(new NDIndex(0), -100000);
```
siddvenk, line 76: What is the significance of this number? I see a similar thing being done extensions/tokenizers/src/main/python/question_answering_converter.py

frankfliu: This is just a hacky way to exclude the first token ([CLS]) from `argmax()`. Python code did the same trick.

demq: I can suggest just copying the "context" parts of the start/end logits in `NDArray endLogits = list.get(1).duplicate();` .  Also note that the classifier tokens in Q/A are often used to determine if the question is answerable ("is_impossible" == true) by ensuring the best answer start*end scores are above those for the corresponding classifier tokens.

siddvenk (approved):

PR #1971, opened 2022-08-28
[example] add an example of time sereis model inferrence

Carkham:
add an example of inference on M5 Forecasting - Accuracy with DeepAR

## Description ##

This is an example of how to use a time series model for inference on real world data, we chose the dataset from [https://www.kaggle.com/competitions/m5-forecasting-accuracy](https://www.kaggle.com/competitions/m5-forecasting-accuracy). We customize the corresponding metrics to evaluate the inference ability of the model.


frankfliu:
Let create the dataset first.

KexinFeng (approved):

KexinFeng:
> Let create the dataset first.

The dataset used in this inference example is downloaded from the public website. What dataset is still needed?

PR #1973, opened 2022-08-29
[timeseries] add all frequencies support.

Carkham:
## Description ##

Now user can use all frequencies for time series inference including "10min", "12D", "24M", "7W" etc.
And taking into account the precision problem, turn float multiple into double multiple in the get lags process

### DIFF
--- a/extensions/timeseries/src/main/java/ai/djl/timeseries/timefeature/Lag.java
+++ b/extensions/timeseries/src/main/java/ai/djl/timeseries/timefeature/Lag.java
@@ -31,17 +31,51 @@ public final class Lag {
      * correspond to the same `season` (+/- `delta`) in previous `k` cycles. Here `delta` and `k`
      * are chosen according to the existing code.
      *
-     * @param freqStr Frequency string of the form [multiple][granularity] such as "12H", "1D" etc.
+     * @param freqStr Frequency string of the form [multiple][granularity] such as "12H", "1D", "6T"
+     *     etc.
      * @param lagUb The maximum value for a lag
      * @return a list of lags
      */
     public static List<Integer> getLagsForFreq(String freqStr, int lagUb) {
         List<List<Integer>> lags;
-        if ("D".equals(freqStr)) {
-            lags = makeLagsForDay(1);
-            lags.addAll(makeLagsForWeek(1f / 7f));
-        } else {
-            throw new IllegalArgumentException("invalid frequency: freq now must be D");
+        TimeOffset timeOffset = TimeOffset.toOffset(freqStr);
+        switch (timeOffset.getName()) {
+            case "Q":
+                if (timeOffset.getN() != 1) {
+                    throw new IllegalArgumentException(
+                            "Only multiple 1 is supported for quarterly. Use x month instead.");
+                }
+                lags = makeLagsForMonth(3. * timeOffset.getN());
+                break;
+            case "M":
+                lags = makeLagsForMonth(timeOffset.getN());
+                break;
+            case "W":
+                lags = makeLagsForWeek(timeOffset.getN());
+                break;
+            case "D":
+                lags = makeLagsForDay(timeOffset.getN());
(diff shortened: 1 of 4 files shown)


extensions/timeseries/src/main/java/ai/djl/timeseries/timefeature/TimeOffset.java
```diff
@@ -0,0 +1,87 @@
...
+    public String getName() {
+        return name;
+    }
+
+    /**
+     * Return the multiple of {@code TimeOffset}.
+     *
+     * @return the multiple of {@code TimeOffset}
+     */
+    public double getN() {
```
siddvenk, line 75: Thoughts on changing this to `getMultipleOfTimeOffset()`? I think it would be more clear than `getN()`

extensions/timeseries/src/main/java/ai/djl/timeseries/timefeature/Lag.java
```diff
@@ -72,7 +106,34 @@ public static List<Integer> getLagsForFreq(String freqStr) {
         return getLagsForFreq(freqStr, 1200);
     }
 
-    private static List<List<Integer>> makeLagsForDay(float multiple) {
+    private static List<List<Integer>> makeLagsForSecond(double multiple) {
```
siddvenk, line 109: It looks like we can refactor all of these methods into a single method of the form `makeLagsForTimeStep(int numCycles, int multiplier, double multiple)` or something similar. Is there a reason we have multiple private methods for each supported time-step/frequency?

extensions/timeseries/src/main/java/ai/djl/timeseries/timefeature/TimeOffset.java
Carkham: Got it.

extensions/timeseries/src/main/java/ai/djl/timeseries/timefeature/Lag.java
Carkham: I originally thought that writing this would make it more clear to the user which piece of the lags to generate for different frequencies. But now it seems more concise to refactor them into one method.

siddvenk: I think that's a fair point, especially if these were public methods. Even if we want to keep all these methods for clarity, I think a good chunk of the body can be refactored out into a separate method since between methods only a couple numbers differ.

Carkham: Yes, that's great. Thanks!

extensions/timeseries/src/main/java/ai/djl/timeseries/timefeature/TimeOffset.java
```diff
@@ -0,0 +1,87 @@
...
+/** This is a class use to get multiple and granularity from frequency string. */
+public class TimeOffset {
+
+    private static String oPattern = "(?<multiple>\\d*)(?<granularity>\\w+)";
+
+    private String name;
+    private int n;
+
+    /**
+     * Constructs a new {@code DateOffset} instance.
```
siddvenk, line 28: `DateOffset` -> `TimeOffset`

extensions/timeseries/src/main/java/ai/djl/timeseries/timefeature/TimeOffset.java
```diff
@@ -0,0 +1,87 @@
...
+ * and limitations under the License.
+ */
+
+package ai.djl.timeseries.timefeature;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/** This is a class use to get multiple and granularity from frequency string. */
+public class TimeOffset {
```
siddvenk, line 20: How are users going to use this class? 

It looks like you're trying to use it as a utility to parse a certain string format, but it's not an accepted type at the moment for any method.

siddvenk (approved):

Issue #1970, opened 2022-08-28
set Function of NDArray, like set(int[] data) should compare the datatype of NDArray and datatype of data


warthecatalyst:
## Description
NDArray anchors = manager.zeros(new Shape(3),Datatype.FLOAT32);
int[] data = new int[]{1,2,3};
anchors.set(data);

### Expected Behavior
should throw an exception when datatypes mismatch

KexinFeng:
What is the current behaviour? I think instead of throwing a datatypes mismatch, another way is to automatically convert the datatype.

warthecatalyst:
it doesn't set the NDArray to corresponding data

PR #1975, opened 2022-08-29
[fix] add data type check to array setter

KexinFeng:
## Description ##

Fix https://github.com/deepjavalibrary/djl/issues/1970. The test code is therein.

## details
A poential place to convert the datatype from the input data type to the target array data type is in
https://github.com/deepjavalibrary/djl/blob/aeb9135686605aab846ada67df800c954b5ebf63/api/src/main/java/ai/djl/ndarray/BaseNDManager.java#L428

Here it does the job of:
> 2\. Multiple dataType maps to the same Buffer type, we have to handle them differently

But it is not ideal to conver the Buffer type inside java
https://stackoverflow.com/questions/38745123/bytebufferasshortbuffer-cannot-be-cast-to-java-nio-floatbuffer
Cannot cast 'java.nio.HeapIntBuffer' to 'java.nio.FloatBuffer'

So here we simply throw an exception when datatypes don't match.

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
+++ b/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
@@ -403,10 +403,12 @@ public abstract class BaseNDManager implements NDManager {
      * @param expected the expected size
      * @throws IllegalArgumentException if buffer size is invalid
      */
-    public static void validateBufferSize(Buffer buffer, DataType dataType, int expected) {
+    public static void validateBuffer(Buffer buffer, DataType dataType, int expected) {
         boolean isByteBuffer = buffer instanceof ByteBuffer;
         DataType type = DataType.fromBuffer(buffer);
-        if (!isByteBuffer && type != dataType) {
+        if (type != dataType && !isByteBuffer) {
+            // It's ok if type != datatype and buffer is ByteBuffer,
+            // since buffer will be copied into ByteBuffer
             throw new IllegalArgumentException(
                     "The input data type: "
                             + type
--- a/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
+++ b/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
@@ -404,9 +404,18 @@ public abstract class BaseNDManager implements NDManager {
      * @throws IllegalArgumentException if buffer size is invalid
      */
     public static void validateBufferSize(Buffer buffer, DataType dataType, int expected) {
+        boolean isByteBuffer = buffer instanceof ByteBuffer;
+        DataType type = DataType.fromBuffer(buffer);
+        if (!isByteBuffer && type != dataType) {
+            throw new IllegalArgumentException(
+                    "The input data type: "
+                            + type
+                            + " does not match target array data type: "
+                            + dataType);
+        }
+
         int remaining = buffer.remaining();
-        int expectedSize =
-                buffer instanceof ByteBuffer ? dataType.getNumOfBytes() * expected : expected;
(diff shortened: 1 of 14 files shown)


api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
```diff
@@ -212,6 +212,16 @@ public NDArray get(NDIndex index) {
     /** {@inheritDoc} */
     @Override
     public void set(Buffer data) {
+        DataType arrayType = getDataType();
```
frankfliu, line 215: I don't think we check list this:
1. We should always allow ByteBuffer to set for any DataType
2. Multiple dataType maps to the same Buffer type, we have to handle them differently

KexinFeng: I find that in the implementation of set(Buffer data) in different engines, like MXNet, PyTorch, TensorFlow, the buffer is already converted to ByteBuffer before being fed into the engines. Do you mean we should add this to NDArrayAdapter?

KexinFeng: By check listing those, is it out of the purpose of automatically converting the input datatype into the target array datatype?

KexinFeng:
A poential place to convert he datatype from the input data type to the target array data type is in
https://github.com/deepjavalibrary/djl/blob/aeb9135686605aab846ada67df800c954b5ebf63/api/src/main/java/ai/djl/ndarray/BaseNDManager.java#L428

Here it does the job of:
> 2\. Multiple dataType maps to the same Buffer type, we have to handle them differently

But it is not ideal to conver the Buffer type inside java
https://stackoverflow.com/questions/38745123/bytebufferasshortbuffer-cannot-be-cast-to-java-nio-floatbuffer
Cannot cast 'java.nio.HeapIntBuffer' to 'java.nio.FloatBuffer'

frankfliu (approved):

api/src/main/java/ai/djl/ndarray/BaseNDManager.java
```diff
@@ -404,9 +404,18 @@ NDManager getAlternativeManager() {
      * @throws IllegalArgumentException if buffer size is invalid
      */
     public static void validateBufferSize(Buffer buffer, DataType dataType, int expected) {
+        boolean isByteBuffer = buffer instanceof ByteBuffer;
+        DataType type = DataType.fromBuffer(buffer);
+        if (!isByteBuffer && type != dataType) {
```
KexinFeng, line 409: The following mismatch case will escape the check:
`type != dataType && buffer is ByteBuffer`, but `dataType` is not one of the byte types = {DataType.UINT8, DataType.INT8, DataType.BOOLEAN}

```suggestion

if (arrayType != inputType ) {
          DataType[] types = {DataType.UINT8, DataType.INT8, DataType.BOOLEAN};
          if (!isByteBuffer || Arrays.stream(types).noneMatch(x -> x == dataTypeType)) {
          
```

frankfliu: ByteBuffer should always be allowed. It's the memory representation of all data types. All NDArray has `toByteBuffer()` method. The JNI only accept ByteBuffer, if we only accept matching Buffer, we have to covert ByteBuffer to Buffer and copy the Buffer to ByteBuffer to pass to JNI. It's not efficient for Hybrid engine. We trying to achieve 0 copy between pytorch and onnx engines. It rely on ByteBuffer.

PR #1982, opened 2022-09-02
Cache the PyTorch "files.txt".

demq:
The cached "files.txt" ensures DJL will still run when the machine is offline, and the native torch jar is not added to the classpath.

## Description ##

Projects that run DJL pytorch models might not work if the machine cannot access "https://publish.djl.ai/pytorch/files.txt" when the "detected" platform is a placeholder (for example the native torch jar is not added to the project).

### DIFF
--- a/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
+++ b/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
@@ -23,7 +23,6 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.lang.reflect.Method;
 import java.net.URL;
--- a/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
+++ b/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
@@ -23,7 +23,6 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.FileOutputStream;
 import java.lang.reflect.Method;
 import java.net.URL;
 import java.net.URLDecoder;
--- a/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
+++ b/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
@@ -26,7 +26,6 @@ import java.io.InputStream;
 import java.lang.reflect.Method;
 import java.net.URL;
 import java.net.URLDecoder;
-import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.file.Files;
 import java.nio.file.Path;
--- a/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
+++ b/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
@@ -392,6 +392,12 @@ public final class LibUtils {
         String link = "https://publish.djl.ai/pytorch/" + matcher.group(1);
         Path tmp = null;
         
+        try {
+            Files.createDirectories(cacheDir);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to create a pytorch cache directory", e);
+        }
+        
         String localFiles = cacheDir + File.separator + "files.txt";
         String localTempFiles = cacheDir + File.separator + "tmpfiles.txt";
         try (InputStream is = new URL(link + "/files.txt").openStream();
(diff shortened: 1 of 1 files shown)


frankfliu:
I don't think this is necessary. If the cache folder already exist, DJL won't access network to retrieve `files.txt`. The only case it will try to download `files.txt` is you run DJL on GPU machine which CUDA version is not support (e.g. 9.2). In this case, DJL will try to access `files.txt` and fallback to CPU. The workaround in this case is to set an environment variable: 
```
export PYTORCH_FLAVOR=cpu
```
Then DJL will directly load cpu version from cache folder.

demq:
> I don't think this is necessary. If the cache folder already exist, DJL won't access network to retrieve `files.txt`. The only case it will try to download `files.txt` is you run DJL on GPU machine which CUDA version is not support (e.g. 9.2). In this case, DJL will try to access `files.txt` and fallback to CPU. The workaround in this case is to set an environment variable:
> 
> ```
> export PYTORCH_FLAVOR=cpu
> ```
> 
> Then DJL will directly load cpu version from cache folder.

The files.txt is also accessed when only the CUDA minor version is "mismatched, for example on a particular machine I have the "detected" flavor = "cu117", which gets matched to "cu116" from the files.txt and I can still run DJL on GPU even in offline mode.

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
```diff
@@ -387,7 +391,31 @@ private static LibTorch downloadPyTorch(Platform platform) {
         }
         String link = "https://publish.djl.ai/pytorch/" + matcher.group(1);
         Path tmp = null;
-        try (InputStream is = new URL(link + "/files.txt").openStream()) {
+        
+        try {
```
frankfliu, line 395:
```
        Path indexFile = cacheDir.resolve(version + ".txt");
        if (Files.notExists(indexFile)) {
            Path tempFile = cacheDir.resolve(version + ".tmp");
            try (InputStream is = new URL(link + "/files.txt").openStream()) {
                Files.createDirectories(cacheDir);
                Files.copy(is, tempFile, StandardCopyOption.REPLACE_EXISTING);
                Utils.moveQuietly(tempFile, indexFile);
            } catch (IOException e) {
                throw new IllegalStateException("Failed to save pytorch index file", e);
            } finally {
                Utils.deleteQuietly(tempFile);
            }
        }

```

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
```diff
@@ -387,7 +391,31 @@ private static LibTorch downloadPyTorch(Platform platform) {
...
+            logger.warn("Could not download the PyTorch files: ", link);
+        }
+        
+        try {
+            Files.move(localTempFiles, localFiles, StandardCopyOption.REPLACE_EXISTING);
+        } catch (Exception e) {
+            logger.warn("Could not download the PyTorch files: ", link);
+        }            
+        
+        try (InputStream is =  new FileInputStream(localFiles.toFile())) {
```
frankfliu, line 418:
```suggestion
        try (InputStream is = Files.newInputStream(indexFile)) {
```

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
```diff
@@ -23,9 +23,13 @@
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.FileInputStream;
```
frankfliu, line 26:
```suggestion
```

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
```diff
@@ -23,9 +23,13 @@
...
+import java.io.FileOutputStream;
```
frankfliu, line 27:
```suggestion
```

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
```diff
@@ -23,9 +23,13 @@
...
 import java.lang.reflect.Method;
 import java.net.URL;
 import java.net.URLDecoder;
+import java.nio.channels.Channels;
```
frankfliu, line 31:
```suggestion
```

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
```diff
@@ -23,9 +23,13 @@
...
+import java.nio.channels.ReadableByteChannel;
```
frankfliu, line 32:
```suggestion
```

frankfliu:
When you create PR, would you please check "Allow edits and access to secrets by maintainers", so that I can push changes to your branch.

demq:
> "Allow edits and access to secrets by maintainers"

I think I did that -> 
Screen Shot 2022-09-02 at 4 42 52 pm

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/LibUtils.java
demq: Pushed an update with this change.

frankfliu (approved):

PR #1984, opened 2022-09-02
[huggingface] Adds Huggingface ModelZoo

frankfliu:
Change-Id: Ief99e4b5449c95a75582ded9a9e2549e582387c5

## Description ##

### DIFF
--- a/extensions/tokenizers/src/main/java/ai/djl/huggingface/zoo/HfModelZoo.java
+++ b/extensions/tokenizers/src/main/java/ai/djl/huggingface/zoo/HfModelZoo.java
@@ -14,7 +14,10 @@ package ai.djl.huggingface.zoo;
 
 import ai.djl.Application;
 import ai.djl.Application.NLP;
+import ai.djl.engine.Engine;
 import ai.djl.repository.Repository;
+import ai.djl.repository.Version;
+import ai.djl.repository.VersionRange;
 import ai.djl.repository.zoo.ModelZoo;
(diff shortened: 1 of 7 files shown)


extensions/tokenizers/src/main/java/ai/djl/huggingface/zoo/HfModelZoo.java
```diff
@@ -0,0 +1,123 @@
...
+            Path dir = Utils.getCacheDir().resolve("cache/repo/" + path);
+            if (Files.notExists(dir)) {
+                Files.createDirectories(dir);
+            } else if (!Files.isDirectory(dir)) {
+                logger.warn("Failed initialize cache directory: " + dir);
+                return Collections.emptyMap();
+            }
+            Type type = new TypeToken<Map<String, Map<String, Object>>>() {}.getType();
+
+            Path file = dir.resolve("models.json");
```
zachgk, line 94: I think we may want a models.json for every DJL release version. Otherwise, we may be suggested new models that are not supported for older versions of DJL

frankfliu: Currently `models.json` is per application, assume the same application will use the same Translator. So we don't really need each file per version. The risk would be some model may require enhancement in current Translator. What we can do is add a `requires="0.20.0+"` field, so we can filter out those future models that is not compatible with old Translator

zachgk (approved):

Issue #1774, opened 2022-07-06
NDArray.set() fails on linux with " Inplace update to inference tensor outside InferenceMode"


demq:
## Description
The NDArray.set() fails when trying to update tensors in a post-processing stage with a message: 
> Inplace update to inference tensor outside InferenceMode is not allowed.You can make a clone to get a normal tensor before doing inplace update.See https://github.com/pytorch/rfcs/pull/17 for more details.

This behavior is observed when running the code with PyTorch engine on a linux machine, the same code runs without any errors on a Mac M1 Pro. The work-around is to first duplicate the tensor by calling a NDArray.duplicate() and performing the .set() on the new tensor.

### Expected Behavior
The PyTorch implementation of the NDArray should either perform the tensor duplication when trying to modify the tensors outside of the InferenceMode, or these tensors should be made immutable. 

### Error Message

> Exception in thread "main" ai.djl.translate.TranslateException: ai.djl.engine.EngineException: Inplace update to inference tensor outside InferenceMode is not allowed.You can make a clone to get a normal tensor before doing inplace update.See https://github.com/pytorch/rfcs/pull/17 for more details.
> Caused by: ai.djl.engine.EngineException: Inplace update to inference tensor outside InferenceMode is not allowed.You can make a clone to get a normal tensor before doing inplace update.See https://github.com/pytorch/rfcs/pull/17 for more details.
> 	at ai.djl.pytorch.jni.PyTorchLibrary.torchMaskedPut(Native Method)
> 	at ai.djl.pytorch.jni.JniUtils.booleanMaskSet(JniUtils.java:416)
> 	at ai.djl.pytorch.engine.PtNDArrayIndexer.set(PtNDArrayIndexer.java:82)
> 	at ai.djl.ndarray.index.NDArrayIndexer.set(NDArrayIndexer.java:157)
> 	at ai.djl.ndarray.NDArray.set(NDArray.java:469)
> 	at ai.djl.ndarray.NDArray.set(NDArray.java:490)
> 	at processOutput(PtBertQATranslator.java:116)

## How to Reproduce?
Create a custom QATranslator, override the processOutput() method like
```
   public List<QAResult> processOutput(TranslatorContext ctx, NDList list) {
        NDManager manager = ctx.getNDManager();
        NDArray start_logits = list.get(0);
        boolean[] bad_tokens_mask = new boolean[128];
        NDArray nd_bad_tokens_mask = manager.create(bad_tokens_mask);
[...]

PR #2009, opened 2022-09-12
fix some bugs in pytorch based examples

siddvenk:
## Description ##

While changing to PyTorch as the default engine I noticed some bugs in examples that rely on PyTorch engine related to https://github.com/deepjavalibrary/djl/issues/1774. 

This changes fixes those examples. I guess this also shows that we are skipping some tests in our current CI, since this issue would have been caught before had the tests run automatically.

### DIFF
--- a/api/src/main/java/ai/djl/modality/cv/translator/BigGANTranslator.java
+++ b/api/src/main/java/ai/djl/modality/cv/translator/BigGANTranslator.java
@@ -40,15 +40,13 @@ public final class BigGANTranslator implements NoBatchifyTranslator<int[], Image
     /** {@inheritDoc} */
     @Override
     public Image[] processOutput(TranslatorContext ctx, NDList list) {
-        NDArray output = list.get(0).addi(1).muli(128).clip(0, 255);
-
+        NDArray output = list.get(0).duplicate().addi(1).muli(128).clip(0, 255);
         int sampleSize = (int) output.getShape().get(0);
         Image[] images = new Image[sampleSize];
 
         for (int i = 0; i < sampleSize; ++i) {
             images[i] = ImageFactory.getInstance().fromNDArray(output.get(i));
         }
-
         return images;
     }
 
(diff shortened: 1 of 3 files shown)


api/src/main/java/ai/djl/modality/cv/translator/BigGANTranslator.java
```diff
@@ -40,13 +40,15 @@ public BigGANTranslator(float truncation) {
     /** {@inheritDoc} */
     @Override
     public Image[] processOutput(TranslatorContext ctx, NDList list) {
-        NDArray output = list.get(0).addi(1).muli(128).clip(0, 255);
+        Image[] images;
+        try (NDArray output = list.get(0).duplicate().addi(1).muli(128).clip(0, 255)) {
```
frankfliu, line 44: use try for NDArray makes the code looks ugly.
We generally don't use try-with-resource in Translator.

examples/src/main/java/ai/djl/examples/inference/StyleTransfer.java
```diff
@@ -52,7 +52,10 @@ public static void main(String[] args) throws IOException, ModelException, Trans
         Image input = ImageFactory.getInstance().fromFile(Paths.get(imagePath));
         Image output = transfer(input, artist);
 
-        logger.info("Using PyTorch Engine. " + artist + " painting generated.");
+        logger.info(
```
frankfliu, line 55:
```
logger.info("Using PyTorch Engine. {} painting generated. Image saved in build/output/cyclegan", artist);
```

api/src/main/java/ai/djl/modality/cv/translator/BigGANTranslator.java
siddvenk: got it - i'll update

examples/src/main/java/ai/djl/examples/inference/StyleTransfer.java
siddvenk: updated

frankfliu (approved):

PR #2026, opened 2022-09-19
[timeseries] add time series dataset support for timeseries package

Carkham:
## Description ##

This pr implements `TimeSeriesDataset` for time series data preprocessing, this class defines a unified format for time series data preprocessing when training, and allows users to customize the source of data data and how to obtain it, we implement a format stored in .csv file here.

We also provide a new `Featurizer` for information on the date obtained from the data.

In the pre-processing section of the data, we added isTraining parameter to the transformer to determine whether it is a training mode.

Last, we create a new `InstanceSampler` to help split past data(context used as model input) and future data(used as label) when training


extensions/timeseries/src/main/java/ai/djl/timeseries/dataset/M5Forecast.java
```diff
@@ -0,0 +1,243 @@
...
+        public M5Forecast build() {
+            validate();
+            return new M5Forecast(this);
+        }
+
+        private void parseFeatures() {
+            if (mf == null) {
+                try (InputStream is =
+                                Objects.requireNonNull(
+                                        M5Forecast.class.getResourceAsStream("m5forecast.json"));
```
frankfliu, line 226: This file is pretty big, we should put this file on S3

Carkham: Thanks, do I need to corporate with my mentor to upload this?

extensions/timeseries/src/main/java/ai/djl/timeseries/dataset/M5Forecast.java
```diff
@@ -0,0 +1,246 @@
...
+        return new Builder();
+    }
+
+    private String getUsagePath(Usage usage) {
+        // We coarse graining the data by summing the sale amount every 7 days and rename the .csv
+        // file as 'weekly_***'
+        String usagePath;
+        switch (usage) {
+            case TRAIN:
+                usagePath = "weekly_sales_train_validation.csv";
```
frankfliu, line 109: Is there is a script to generate this file? I think we should upload this file to S3 as dataset

KexinFeng: Yes. This is a python script, which transforms the daily time series data to weekly data.

KexinFeng (approved):

PR #2027, opened 2022-09-19
[timeseries] add some basic block and deepAR model

Carkham:
## Description ##

This PR adds two new sections for creating trainable models.

The first part is embedding and scaler, which are responsible for encoding the input features and scaling the target values.

The second part is that we implemented a **deepAR** network model with the above-mentioned blcok, which refers to the implementation in [gluonTS](https://ts.gluon.ai/stable/api/gluonts/gluonts.mx.distribution.html). 

Since the model predicts a probability distribution, we need to override the different forward methods when training and inference with different block.

### DIFF
--- a/examples/src/main/java/ai/djl/examples/inference/timeseries/package-info.java
+++ b/examples/src/main/java/ai/djl/examples/inference/timeseries/package-info.java
@@ -1,0 +1,15 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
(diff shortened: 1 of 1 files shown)


extensions/timeseries/src/main/java/ai/djl/timeseries/block/FeatureEmbedder.java
```diff
@@ -0,0 +1,172 @@
...
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package ai.djl.timeseries.block;
```
frankfliu, line 14: Please add unit test

frankfliu (changes requested):

extensions/timeseries/src/main/java/ai/djl/timeseries/block/FeatureEmbedder.java
Carkham: Got it.

frankfliu (approved):

PR #2036, opened 2022-09-26
Adds TabNet model for tabular dataset in modelzoo

warthecatalyst:
TabNet model almost complete, except for its test and some minor bugs
What I've done in this pr:
1. add a SparseMax function in nn.core(already been merged)
2. add an activation function: tabNetGLU in nn.Activation and its corresponding test function testTabNetGLU in ActivationTest
3. add feature transformer, attention transformer and Decision Step in tabNet
4. tabNet is a Block which extends AbstractBlock and it's functions like forwardInternal, getOutputShapes and so on
5. tabNet has a builder class for users to use

TODO: test for tabNet


api/src/main/java/ai/djl/nn/Activation.java
```diff
@@ -258,6 +258,29 @@ public static NDList selu(NDList arrays) {
...
     }
 
+    /**
+     * Applies tabNetGLU activation(which is mostly used in tabNet) on the input {@link NDArray}.
+     *
+     * @param array the input {@link NDArray}
+     * @param units the half number of the resultant features
+     * @return the {@link NDArray} after applying tabNetGLU function
+     */
+    public static NDArray tabNetGLU(NDArray array, int units) {
```
zachgk, line 268: If it is (mostly) specialized to tabnet, then it is probably better to put the activation on the Tabnet class instead and leave the `Activation` class for general purpose activations. As long as you make it `public static`, it can still be accessed just as easily.

api/src/main/java/ai/djl/nn/core/SparseMax.java
```diff
@@ -10,7 +10,6 @@
  * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
  * and limitations under the License.
  */
-
```
zachgk, line 13: Try to avoid accidental changes like these

zachgk (commented):
One smaller comment, but otherwise it looks pretty good. Just add a working test, maybe a Translator, and we should merge it

zachgk (approved):

Issue #2018, opened 2022-09-15
Mac M1 GPU support via Pytorch and MPS


ajrnz:
Given that pytorch supports GPUs (via Metal/MPS) on Mac M1 machines (since version [pytorch 1.12](https://pytorch.org/blog/introducing-accelerated-pytorch-training-on-mac/), should it not be possible to enable this in the pytorch engine?

How much work would this be now that the JNI bindings are now ARM (aarch64)

The speed up is significant for Mac M1 users.

## References
https://pytorch.org/blog/introducing-accelerated-pytorch-training-on-mac/

lanking520:
Currently we are already using the offical torch 1.12.1 MacOS Arm64 wheel to build the JNI. So it has capabitlity to extend to GPU, but we didn't implement the MPS device yet. We are welcome for any contribution to this device addition. You can try to get it work by adding MPS device in the code

PR #2037, opened 2022-09-27
Add support for Apples' Metal Performance Shaders (MPS) in pytorch

demq:
## Description ##
Pytorch 1.12 supports Apple’s Metal Performance Shaders (MPS) for accelerated training/inference. A simple test on HuggingFace models shows 2x-3x improvements on the inference latency over the CPU cores on an Apple M1Pro with 8 CPU and 14 GPU cores:
```
--------------------------------------------------------------------------------
Model:              roberta-hf
Device:             cpu()
Samples:            34
Model Input Length: 384
Metrics:
	Total: 140.206 ± 5.880      msec
	Inference: 138.794 ± 5.713      msec
	Preprocess: 0.441 ± 0.553      msec
	Postprocess: 0.824 ± 0.567      msec
--------------------------------------------------------------------------------
Model:              roberta-hf
Device:             mps(-1)
Samples:            34
Model Input Length: 384
Metrics:
	Total: 65.500 ± 9.166      msec
	Inference: 41.235 ± 6.198      msec
	Preprocess: 1.059 ± 0.338      msec
	Postprocess: 23.206 ± 6.794      msec

--------------------------------------------------------------------------------
```
This improvement has been requested by me in the DJL help slack channel, as well as more recently in #2018 . 

- Adds a new "MPS" device type and maps to the corresponding pytorch device type.

Some caveats:
- The "MPS" mode only works if  `torch::jit::load(path, device, map);`  is called with `device= torch::nullopt` as in torch  the model gets deserialized in 'legacy' mode, which only support CPU and GPU devices. The model get desiralized on the CPU and then can be converted to MPS. For this, in traced torch model's directory  in the file "serving.properties" the `option.mapLocation=false` should be set, which triggers this implementation already present in DJL native module.
Alternatively, the the following check can be added the pytorch-native module: https://github.com/deepjavalibrary/djl/blob/f097502b9fe7607203320c4aecee14c759e29d5e/engines/pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_inference.cc#L49-L52

``` 
if (jmap_location && (device.is_cpu() || device.is_cuda())) {
```

-  An env variable s of hould be set`PYTORCH_ENABLE_MPS_FALLBACK=1` as a work around a pytorch implementation deficiency of 'aten::index.Tensor' for MPS, see https://github.com/pytorch/pytorch/issues/77764

### DIFF
--- a/api/src/main/java/ai/djl/Device.java
+++ b/api/src/main/java/ai/djl/Device.java
@@ -21,10 +21,10 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
- * The {@code Device} class provides the specified assignment for CPU/GPU processing on the {@code
+ * The {@code Device} class provides the specified assignment for CPU/GPU/MPS processing on the {@code
  * NDArray}.
  *
- * <p>Users can use this to specify whether to load/compute the {@code NDArray} on CPU/GPU with
+ * <p>Users can use this to specify whether to load/compute the {@code NDArray} on CPU/GPU/MPS with
  * deviceType and deviceId provided.
  *
  * @see <a href="https://d2l.djl.ai/chapter_deep-learning-computation/use-gpu.html">The D2L chapter
@@ -36,6 +36,7 @@ public final class Device {
 
     private static final Device CPU = new Device(Type.CPU, -1);
     private static final Device GPU = Device.of(Type.GPU, 0);
+    private static final Device MPS = Device.of(Type.MPS, -1);
 
     private static final Pattern DEVICE_NAME = Pattern.compile("([a-z]+)([0-9]*)");
 
@@ -45,7 +46,7 @@ public final class Device {
     /**
      * Creates a {@code Device} with basic information.
      *
-     * @param deviceType the device type, typically CPU or GPU
+     * @param deviceType the device type, typically CPU, GPU, or MPS
      * @param deviceId the deviceId on the hardware. For example, if you have multiple GPUs, you can
      *     choose which GPU to process the NDArray
      */
@@ -57,7 +58,7 @@ public final class Device {
     /**
      * Returns a {@code Device} with device type and device id.
      *
-     * @param deviceType the device type, typically CPU or GPU
+     * @param deviceType the device type, typically CPU, GPU, or MPS
      * @param deviceId the deviceId on the hardware.
      * @return a {@code Device} instance
      */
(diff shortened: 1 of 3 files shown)


api/src/main/java/ai/djl/Device.java
```diff
@@ -36,6 +36,7 @@ public final class Device {
 
     private static final Device CPU = new Device(Type.CPU, -1);
     private static final Device GPU = Device.of(Type.GPU, 0);
+    private static final Device MPS = Device.of(Type.MPS, -1);
```
frankfliu, line 39: MPS is PyTorch specific, for the time being, it's better keep it PyTorch only. We can add to `Device` class when it become standard.

api/src/main/java/ai/djl/Device.java
```diff
@@ -209,9 +219,19 @@ public static Device gpu(int deviceId) {
...
+     */
+    public static Device mps() {
+        return MPS;
+    }
+
     /** Contains device type string constants. */
     public interface Type {
         String CPU = "cpu";
         String GPU = "gpu";
+        String MPS = "mps";
```
frankfliu, line 235: We don't need make changes in `Device` class. Let's keep it private in PyTorch for now

api/src/main/java/ai/djl/Device.java
demq: I am not too sure how to avoid this part "cleanly" without having to rewrite a bunch of code in pytorche's JniUtils.java, as most of the functions there use a `Device` object to infer the PtDeviceType:
https://github.com/deepjavalibrary/djl/blob/8fce4ebf0eceda51abbce6448ca029fb5599899c/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/JniUtils.java#L191-L202

One way is to change the `PtDeviceType:: toDeviceType(Device device)` to somehow return the code for MPS device on osx-aarm64 systems, but I don't know how to do that:
https://github.com/deepjavalibrary/djl/blob/8fce4ebf0eceda51abbce6448ca029fb5599899c/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtDeviceType.java#L28-L38

api/src/main/java/ai/djl/Device.java
demq: I agree that this might create some false expectations form users that apple metal is supported in general, but the current solution means that apple silicone users don't need to rewrite any of their existing code.

The users can simply specify "mps" as their device type in  their model Criteria when running on Mac M1/M2 and get the acceleration working. Otherwise the developers would need to add pytorch and OSX-specific code to support "mps", making it less likely to be used. The device type, on the other hand, is usually just a system setting / cli argument to ensure let's say the GPU acceleration can be switched on/off.

frankfliu (approved):

PR #2043, opened 2022-09-28
Creates Ensembleable

zachgk:
This creates a new interface Ensembleable to be used for translator outputs that can be ensembled or averaged together. The interface will later be used as part of workflows within DJL Serving.


frankfliu:
I updated your PR, please take a look.

zachgk:
It mostly looks good. I'm glad you found a way to do it without needing the casting. I did add back a test for mismatched class names. I think that especially when using imported model zoos, the chances of issues here are likely and this should help prevent those issues from being hard to detect or debug

frankfliu (approved):

PR #2045, opened 2022-09-28
Handle error when forget to initialize a child block

zachgk:
This throws an error with a much clearer error message in case someone fails to initialize a child block. It should help greatly reduce the time it takes to debug the error in this situation.

### DIFF
--- a/api/src/main/java/ai/djl/nn/UninitializeParameterException.java
+++ b/api/src/main/java/ai/djl/nn/UninitializeParameterException.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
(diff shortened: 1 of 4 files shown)


api/src/main/java/ai/djl/nn/UninitializeParameterException.java
```diff
@@ -0,0 +1,74 @@
...
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package ai.djl.nn;
+
+/** Thrown to indicate that a {@link Parameter} was not initialized. */
+public class UninitializeParameterException extends RuntimeException {
```
siddvenk, line 16: nit: Should this be Uninitialize<ins>d</ins>ParameterException?

zachgk: It should. Good catch

frankfliu (approved):

PR #2054, opened 2022-09-30
Fix ames and various tabular improvements

zachgk:
First for Ames, this fixes a bug (misclassified column type) to fix ames. It also simplifies the Ames dataset by removing the featureToMap and replacing it with the standard usage of the categorical Featurizer.

For tabular, it moves the validation function from the builder to the constructor for tabular. This fixes the problem of how easy it is to forget the validation when creating a dataset. As proof of this, several datasets were missing the calls to validate. It adds two new tests for all features and no features in Ames.


basicdataset/src/main/java/ai/djl/basicdataset/tabular/TabularDataset.java
```diff
@@ -44,6 +44,13 @@ public TabularDataset(BaseBuilder<?> builder) {
         super(builder);
         features = builder.features;
         labels = builder.labels;
+
```
frankfliu, line 47: Avoid throw exception in Constructor. Move this logic into `validate()`

zachgk: We had a `validate()` function, but the difficulty is that there wasn't a mechanism to ensure that the datasets inhering `TabularDataset` would call validate(). Resulting from this, even we did not remember to validate in all of them. Besides that, for most users they will get the exception as a result of calling `datasetBuilder.build()` regardless. Is there some other way you suggest implementing this?

frankfliu (approved):

PR #2080, opened 2022-10-12
[api] Allow to show NDArray content in Debugger 2

KexinFeng:
## Description ##

The previous solution still suffers from a slowing down when printing the string outside the debugger mode. Here the solution is to utillize the existing pruning mechanism, and put the execution of `init(array)` also under the condition that the array is not too large.

A testing code is added.

### DIFF
--- a/docs/development/development_guideline.md
+++ b/docs/development/development_guideline.md
@@ -157,8 +157,8 @@ You can create your own NDArray renderer as follows:
 ![](img/custom_debug_view.png)
 
 Opening the IDE Settings/Preferences (`⌘ ,`), in *Java Type Renderer* section, you can enable array display by:
-- Change the "Use following expression" field to [toDebugString(true)](https://javadoc.io/static/ai.djl/api/0.19.0/ai/djl/ndarray/NDArray.html#toDebugString(boolean)) to show NDArray content
-- Change the "Use following expression" field to something like [toDebugString(100, 10, 10, 20, true)](https://javadoc.io/static/ai.djl/api/0.19.0/ai/djl/ndarray/NDArray.html#toDebugString(int,int,int,int,boolean))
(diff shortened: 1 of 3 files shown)


docs/development/development_guideline.md
```diff
@@ -156,12 +156,11 @@ IntelliJ allows you to [customize the data view in the Debug Tools](https://www.
 You can create your own NDArray renderer as follows:
 ![](img/custom_debug_view.png)
 
-Please make sure to:
-
-- Check the "On-demand" option, which causes IntelliJ to only render the NDArray when you click on the variable.
-- Change the "Use following expression" field to `toDebugString(true)` to show NDArray content
-- Change the "Use following expression" field to something like `toDebugString(1000, 10, 10, 20, true)`
+Opening the IDE settings (`⌘ ,`), in *Java Type Renderer* section, you can enable array display by:
+- Change the "Use following expression" field to [toDebugString(true)](https://javadoc.io/static/ai.djl/api/0.19.0/ai/djl/ndarray/NDArray.html#toDebugString-int-int-int-int-) to show NDArray content
```
frankfliu, line 160: The link is wrong, it doesn't go to the API

docs/development/development_guideline.md
```diff
@@ -156,12 +156,11 @@ IntelliJ allows you to [customize the data view in the Debug Tools](https://www.
...
+- Change the "Use following expression" field to something like [toDebugString(100, 10, 10, 20, true)](https://javadoc.io/static/ai.djl/api/0.19.0/ai/djl/ndarray/NDArray.html#toDebugString-int-int-int-int-)
```
frankfliu, line 161: this link also wrong, new API doesn't exist in 0.19.0

frankfliu (approved):

PR #2082, opened 2022-10-13
Enable train data cut

KexinFeng:
## Description ##

This feature enables the control over the number of training data.

Note:
Here the batch indices obtained from Sampler the interface is 
`Iterator<List<Long>> sample(RandomAccessDataset dataset);`
However, this is not convenient for training. Like in this case the cut of training data set, ideally should be implemented without modifying the library. In the future, we can consider utilizing the NDArray indexing features and enable `Iterator<NDArray>` as the type of indices. This will allow more flexibility, desired in training/experimenting.


examples/src/main/java/ai/djl/examples/training/transferlearning/TransferFreshFruitCustomizeSampler.java
```diff
@@ -0,0 +1,289 @@
...
+            return batchSize;
+        }
+
+        @Override
+        public void close() {
+            manager.close();
+            manager = null;
+        }
+
+        static class Iterate implements Iterator<List<Long>> {
```
zachgk, line 253: Can you use either `datasetBuilder.optLimit(limit)` or `dataset.subDataset(0, limit)` to get the limited dataset?

PR #2070, opened 2022-10-08
Transfer learning with pytorch engine on fresh fruit dataset

KexinFeng:
## Description ##
In this PR, the following features are enabled:
1. The transfer learning feature on PyTorch engine. User can opt in this feature by setting "retrain" to String "true" in `Criteria.builder().optOption(String key, String value)`, when loading the traced pretrained model.
2. The trainable variables(parameters) can have per variable loss function.

PR #2084, opened 2022-10-14
Randperm on PyTorch

KexinFeng:
## Description ##

- A new method `randPerm` is added.


engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtModel.java
```diff
@@ -92,10 +100,15 @@ public void load(Path modelPath, String prefix, Map<String, ?> options)
...
                             mapLocation,
                             extraFileKeys,
-                            extraFileValues);
+                            extraFileValues,
+                            retrain);
             for (int i = 0; i < extraFileKeys.length; i++) {
                 properties.put(extraFileKeys[i], extraFileValues[i]);
             }
+            // Freeze the parameters if not retrain
+            for (Pair<String, Parameter> paramPair : block.getParameters()) {
```
zachgk, line 109: Is there a reason that retrain has to be added as an option to the model? Can we not just direct users to call `freeze()` themselves if they don't wish to retrain the model? And as part of that, I am not sure if freezing the parameters is the default that users would expect

api/src/main/java/ai/djl/training/optimizer/Sgd.java
```diff
@@ -56,6 +57,9 @@ protected Sgd(Builder builder) {
     /** {@inheritDoc} */
     @Override
     public void update(String parameterId, NDArray weight, NDArray grad) {
+        if (learningRateTracker instanceof FixedPerVarTracker) {
+            ((FixedPerVarTracker) learningRateTracker).setParameterId(parameterId);
```
zachgk, line 61: It isn't great to use a system like this to pass a special case argument of parameterId here. Instead, maybe we could add a new kind of tracker named something like `ParameterTracker`. The ParamterTracker can require the parameter to get a new value, and all standard Trackers would also be a `ParameterTracker` if we have `Tracker extend ParameterTracker`. If we leave it as a special case like this, it would make it difficult if a new tracker that also required parameters were created or if new optimizers were created without the special handling for the FixedPerVarTracker

engines/pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_inference.cc
```diff
@@ -44,7 +44,25 @@ Java_ai_djl_pytorch_jni_PyTorchLibrary_moduleLoad__Ljava_lang_String_2_3IZ_3Ljav
     map[name] = "";
   }
 
-  JITCallGuard guard;
+  if (!jretrain) {
```
zachgk, line 47: I couldn't find the difference in behavior between with and without retrain. Is there one?

examples/src/main/java/ai/djl/examples/training/transferlearning/TransferFreshFruit.java
```diff
@@ -0,0 +1,196 @@
...
+    public static TrainingResult runExample(String[] args)
+            throws IOException, TranslateException, ModelException, URISyntaxException {
+        Arguments arguments = new Arguments().parseArgs(args);
+        if (arguments == null) {
+            return null;
+        }
+
+        // Also available at
+        // "https://mlrepo.djl.ai/model/cv/image_classification/ai/djl/pytorch/resnet18_embedding/0.0.1/traced_resnet18_embedding.pt.gz";
+        String modelUrls = "/Users/fenkexin/Desktop/transferDJL/code/base_nw.pt";
```
zachgk, line 72: You should update the modelUrls (and dataset urls if possible) to use the remote version so the example immediately works on everyone's machine before merging

examples/src/main/java/ai/djl/examples/training/transferlearning/TransferFreshFruit.java
```diff
@@ -0,0 +1,196 @@
...
+                });
+
+        return new DefaultTrainingConfig(new SoftmaxCrossEntropy("SoftmaxCrossEntropy"))
+                .addEvaluator(new Accuracy())
+                .optDevices(Engine.getInstance().getDevices(1))
+                .addTrainingListeners(TrainingListener.Defaults.logging(outputDir))
+                .addTrainingListeners(listener);
+    }
+
+    private static class SoftmaxCrossEntropy extends Loss {
```
zachgk, line 174: Why not use the standard `SoftmaxCrossEntropyLoss`?

api/src/main/java/ai/djl/ndarray/BaseNDManager.java
```diff
@@ -190,6 +190,12 @@ public NDArray randomInteger(long low, long high, Shape shape, DataType dataType
         throw new UnsupportedOperationException("Not supported!");
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public NDArray randPerm(long n) {
```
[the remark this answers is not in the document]
KexinFeng: Yeah, got it.

api/src/main/java/ai/djl/translate/StackBatchifier.java
```diff
@@ -77,7 +77,11 @@ public NDList batchify(NDList[] inputs) {
                     NDArray currInput = input.get(i);
                     if (!currInput.getShape().equals(kindDataShape)) {
                         throw new IllegalArgumentException(
-                                "You cannot batch data with different input shapes", e);
+                                "You cannot batch data with different input shapes"
+                                        + currInput.getShape()
```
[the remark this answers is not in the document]
KexinFeng: I see.

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtModel.java
KexinFeng: I see. Yes, this should be replaced by `baseBlock.freezeParameters(false);`. I'll edit it in the pr https://github.com/deepjavalibrary/djl/pull/2070.

The reason the default is to set the baseBlock frozen() is that even the pretrained model is not frozen, its learning rate cannot be very large, or the same as the subsequent blocks. Ie the pretrained parameters should not be trained too much. So it is safe to freeze those parameters by default. Maybe we can add some comments about this.

examples/src/main/java/ai/djl/examples/training/transferlearning/TransferFreshFruit.java
KexinFeng: Yeah. This is updated in https://github.com/deepjavalibrary/djl/pull/2070.

examples/src/main/java/ai/djl/examples/training/transferlearning/TransferFreshFruit.java
KexinFeng: The standard `SoftmaxCrossEntropyLoss` doesn't take the output of the softmax function; it currently either take the values before applying softmax, or the values after applying `logit = log(softmax)`.
Here, I built the network to output softmax. This is mainly because it is natural to understand (as probabilities). Correspondingly, I also corrected the calculation in Accuracy.java. You can take a look if it is correct.

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtModel.java
```diff
@@ -78,12 +78,20 @@ public void load(Path modelPath, String prefix, Map<String, ?> options)
...
+            boolean retrain = false;
             // load jit extra files
             if (options != null) {
                 if (options.containsKey("extraFiles")) {
                     extraFileKeys = ((String) options.get("extraFiles")).split(",");
                     extraFileValues = new String[extraFileKeys.length];
                 }
+                if (options.containsKey("retrain")) {
+                    String value = (String) options.get("retrain");
+                    retrain =
```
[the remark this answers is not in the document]
KexinFeng: Got it. This will be modified in https://github.com/deepjavalibrary/djl/pull/2070.

zachgk (approved):

PR #2105, opened 2022-10-26
[timeseries] Update M5Forecast dataset and its unittest

frankfliu:
## Description ##


extensions/timeseries/src/main/resources/ai/djl/timeseries/dataset/m5forecast_parser.json
```diff
@@ -0,0 +1,3100 @@
+{
```
zachgk, line 1: Maybe we should keep this in S3/metadata instead

zachgk (approved):
One suggestion. Otherwise, LGTM

extensions/timeseries/src/main/resources/ai/djl/timeseries/dataset/m5forecast_parser.json
frankfliu: I think it's better keep it with the code:
1. This is tightly coupled with the code.
2. The s3 location become confusing, it doesn't make sense to keep it in unittest dataset
3. Users has to copy this file if they want to download dataset from Kaggle

PR #2120, opened 2022-11-02
Refactor unit test TestRequirements, add missing TestRequirements

siddvenk:
## Description ##

Two sets of changes in this PR:
- Remove duplicate usage of TestRequirements.someRequirement() in unit tests and move them to a setUp method.
- Adds missing TestRequirements on some timeseries unit tests to not run on Arm

### DIFF
--- a/extensions/timeseries/src/test/java/ai/djl/timeseries/block/BlockTest.java
+++ b/extensions/timeseries/src/test/java/ai/djl/timeseries/block/BlockTest.java
@@ -20,16 +20,24 @@ import ai.djl.ndarray.types.DataType;
 import ai.djl.ndarray.types.Shape;
 import ai.djl.nn.Parameter;
 import ai.djl.testing.Assertions;
+import ai.djl.testing.TestRequirements;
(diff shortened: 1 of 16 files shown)


extensions/timeseries/src/test/java/ai/djl/timeseries/block/BlockTest.java
```diff
@@ -20,17 +20,24 @@
...
 import org.testng.annotations.Test;
 
 import java.util.Arrays;
 import java.util.List;
 
 public class BlockTest {
 
+    @BeforeClass
+    public void setUp() {
+        TestRequirements.notArm();
```
frankfliu, line 38: Add TODO here, we should be able to support ARM when we add PyTorch support

siddvenk: Good point - I'll add those to all the timeseries tests

frankfliu (approved):

lanking520 (approved):

PR #2129, opened 2022-11-03
LightGBM inference result matches input type

zachgk:
This makes the LightGBM inference return the same as the input type (either float32 or float64) rather than always returning float64 per the API.

In addition, it adds a bit nicer handling for creation from a Buffer in that it will also accept either a FloatBuffer or a DoubleBuffer.

### DIFF
--- a/engines/ml/lightgbm/src/main/java/ai/djl/ml/lightgbm/jni/JniUtils.java
+++ b/engines/ml/lightgbm/src/main/java/ai/djl/ml/lightgbm/jni/JniUtils.java
@@ -18,6 +18,8 @@ import ai.djl.ml.lightgbm.LgbmNDArray;
 import ai.djl.ml.lightgbm.LgbmNDManager;
 import ai.djl.ml.lightgbm.LgbmSymbolBlock;
 import ai.djl.ndarray.NDArray;
+import ai.djl.ndarray.types.DataType;
+import ai.djl.util.Pair;
 
 import com.microsoft.ml.lightgbm.SWIGTYPE_p_double;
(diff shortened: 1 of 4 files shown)


engines/ml/lightgbm/src/main/java/ai/djl/ml/lightgbm/jni/JniUtils.java
```diff
@@ -92,12 +100,29 @@ public static double[] inferenceMat(SWIGTYPE_p_p_void model, int iterations, Lgb
...
-            long length = lightgbmlib.int64_tp_value(outLength);
-            double[] values = new double[(int) length];
-            for (int i = 0; i < length; i++) {
-                values[i] = lightgbmlib.doubleArray_getitem(outBuffer, i);
+            int length = Math.toIntExact(lightgbmlib.int64_tp_value(outLength));
+            if (a.getDataType() == DataType.FLOAT32) {
+                ByteBuffer bb = ByteBuffer.allocateDirect(length * 4);
+                FloatBuffer wrapped = bb.asFloatBuffer();
+                for (int i = 0; i < length; i++) {
+                    wrapped.put((float) lightgbmlib.doubleArray_getitem(outBuffer, i));
```
frankfliu, line 108: Can we call `lightgbm.floatArray_getitem()`?

zachgk: The behavior of the lightgbm engine is to always return a double array and this is the utility provided by lightgbm to get from a JNI `double[]` data structure. So if we want to return a float[], we need to convert it from double to float

frankfliu (approved):

PR #2137, opened 2022-11-05
[Tokenizer] fix char offset

### LINKED ISSUE #2112: Returned character token spans are not correct for some inputs
## Description
For some inputs the character spans returned by `Encoding.getCharTokenSpans()` does not correspond to the correct character span in the input string.

### Expected Behavior
`Encoding.getCharTokenSpans()` should return the correct character span, and it should return the same character spans as when using the same tokenizer in Python.

### Error Message
No error message.

### Steps to reproduce
The error can be reproduced with this short Scala snippet
```
    val someText = "just some string including a £9.00 price"
    val tokenizer = HuggingFaceTokenizer.newInstance("distilbert-base-uncased")
    val encodedText = tokenizer.encode(someText)

    val charSpans = encodedText.getCharTokenSpans()
    val tokens = encodedText.getTokens()
    for (i <- 1 until charSpans.length) {
        println(s"$token: ${tokens(i)}, orig string: ${someText.substring(charSpans(i).getStart, charSpans(i).getEnd)}, span start=${charSpans(i).getStart}, span end=${charSpans(i).getEnd}")
    }
```
This script has the following output
```
token: just, orig string: just, span start=0, span end=4
token: some, orig string: some, span start=5, span end=9
token: string, orig string: string, span start=10, span end=16
token: including, orig string: including, span start=17, span end=26
token: a, orig string: a, span start=27, span end=28
token: £, orig string: £9, span start=29, span end=31
token: ##9, orig string: ., span start=31, span end=32
token: ., orig string: 0, span start=32, span end=33
token: 00, orig string: 0 , span start=33, span end=35
```
So for some reason the 6th token ("£") gets a character span of two characters, and then the rest of the spans are off by one.

Running this very similar Python script
```
tokenizer = AutoTokenizer.from_pretrained("distilbert-base-uncased")

some_string = "just some string including a £9.00 price"
encoded_input = tokenizer(some_string, return_offsets_mapping=True)

tokens = tokenizer.convert_ids_to_tokens(encoded_input["input_ids"])
for token, span in zip(tokens[1:-1], encoded_input["offset_mapping"][1:-1]):
    print(f"token: {token}, orig string: {some_string[span[0]:span[1]]}, span start={span[0]}, span end={span[1]}")
```
gives the following output
```
token: just, orig string: just, span start=0, span end=4
token: some, orig string: some, span start=5, span end=9
token: string, orig string: string, span start=10, span end=16
token: including, orig string: including, span start=17, span end=26
token: a, orig string: a, span start=27, span end=28
token: £, orig string: £, span start=29, span end=30
token: ##9, orig string: 9, span start=30, span end=31
token: ., orig string: ., span start=31, span end=32
token: 00, orig string: 00, span start=32, span end=34
token: price, orig string: price, span start=35, span end=40
```
And the 6th token ("£") gets a character span of one characters, as expected.

## Environment Info

"ai.djl.huggingface:tokenizers:0.19.0"

lanking520:
## Description ##

Fixed char offset issues.
fix #2112

### DIFF
--- a/extensions/tokenizers/src/test/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizerTest.java
+++ b/extensions/tokenizers/src/test/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizerTest.java
@@ -67,8 +67,8 @@ public class HuggingFaceTokenizerTest {
                 new CharSpan(14, 17),
                 new CharSpan(18, 21),
                 new CharSpan(22, 25),
-                new CharSpan(26, 30),
-                new CharSpan(31, 32),
+                new CharSpan(26, 27),
+                new CharSpan(28, 29),
                 null
             };
             int expectedLength = charSpansExpected.length;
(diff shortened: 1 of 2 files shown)


frankfliu (approved):

extensions/tokenizers/src/test/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizerTest.java
```diff
@@ -389,4 +389,29 @@ public void testTruncationAndPaddingForPairInputs() throws IOException {
             Assert.assertEquals(encoding.getIds().length, 8);
         }
     }
+
+    @Test
+    public void testSpecialTokenHandling() throws IOException {
```
frankfliu, line 394: We might not need this test, the test above this already has special characters.

andreabrduque:
> @andreabrduque FYI

Uh thanks for this one. It was indeed returning UTF-8 Bytes instead of the right char spans :)

PR #2139, opened 2022-11-07
[doc] Remove `block` which seems to provide no purpose

asbachb:
## Description ##

Removed `block` from a code block which seems to be placed by mistake. Needs to be removed when using copy & paste anyway.


asbachb:
Confirmed by @frankfliu as intentional:

> The block in jupyter note is not a mistake. In Juputer notebook, the last express means:
> ```java
> System.out.println(block)
> ```
> With this block it print out the block strucutre

Issue #2159, opened 2022-11-17
Pytorch EngineException: Inference tensors cannot be saved for backward


aminnasiri:
## Description
I developed an example of Rank Classification using BERT on Amazon Review dataset. [This guide](https://docs.djl.ai/engines/pytorch/pytorch-engine/index.html) and It worked fine with Apache MXNET engine, but it is throwing EngineException on Pytorch engine.

### Expected Behavior
I am expecting to see the Pytorch engine is working fine too.

### Error Message
` Error: ai.djl.engine.EngineException: Inference tensors cannot be saved for backward. To work around you can make a clone to get a normal tensor and use it in autograd.
	at ai.djl.pytorch.jni.PyTorchLibrary.torchNNLinear(Native Method)
	at ai.djl.pytorch.jni.JniUtils.linear(JniUtils.java:1189)
	at ai.djl.pytorch.engine.PtNDArrayEx.linear(PtNDArrayEx.java:390)
	at ai.djl.nn.core.Linear.linear(Linear.java:183)
	at ai.djl.nn.core.Linear.forwardInternal(Linear.java:88)
	at ai.djl.nn.AbstractBaseBlock.forwardInternal(AbstractBaseBlock.java:126)
	at ai.djl.nn.AbstractBaseBlock.forward(AbstractBaseBlock.java:91)
	at ai.djl.nn.SequentialBlock.forwardInternal(SequentialBlock.java:209)
	at ai.djl.nn.AbstractBaseBlock.forward(AbstractBaseBlock.java:91)
	at ai.djl.training.Trainer.forward(Trainer.java:175)
	at ai.djl.training.EasyTrain.trainSplit(EasyTrain.java:122)
	at ai.djl.training.EasyTrain.trainBatch(EasyTrain.java:110)
	at ai.djl.training.EasyTrain.fit(EasyTrain.java:58)
	at com.thinksky.classification.TrainModel.trainModel(TrainModel.java:96)
	at com.thinksky.classification.TrainModel_ClientProxy.trainModel(Unknown Source)
	at com.thinksky.PredictResource.executeQuery(PredictResource.java:104)
	at com.thinksky.PredictResource_VertxInvoker_executeQuery_315faff7c0f6c7728fd1c92cfb1b39aa7f024059.invokeBean(Unknown Source)
	at io.quarkus.vertx.runtime.EventConsumerInvoker.invoke(EventConsumerInvoker.java:41)
[...]

Issue #2144, opened 2022-11-08
Inference tensors cannot be saved for backward.


SuperMaskv:
I don't know if this is a bug. I'm trying to follow the [official tutorial](https://docs.djl.ai/jupyter/rank_classification_using_BERT_on_Amazon_Review.html) using pytorch engine. 

Here is my code and exception.
```java
String modelUrl = "https://resources.djl.ai/test-models/traced_distilbert_wikipedia_uncased.zip";
Criteria<NDList, NDList> criteria = Criteria.builder()
    .optApplication(Application.NLP.WORD_EMBEDDING)
    .setTypes(NDList.class, NDList.class)
    .optModelUrls(modelUrl)
    .optProgress(new ProgressBar())
    .build();

ZooModel<NDList, NDList> embedding = criteria.loadModel();
Predictor<NDList, NDList> embedder = embedding.newPredictor();
SequentialBlock classifier = new SequentialBlock()
    .add(
            ndList -> {
                NDArray data = ndList.singletonOrThrow();
                long batchSize = data.getShape().get(0);
                long maxLen = data.getShape().get(1);
                NDList inputs = new NDList();
                inputs.add(data.toType(DataType.INT64, false));
                inputs.add(data.getManager().full(data.getShape(), 1, DataType.INT64));
                inputs.add(data.getManager().arange(maxLen).toType(DataType.INT64, false).broadcast(data.getShape()));
                try {
                    return embedder.predict(inputs);
                } catch (TranslateException e) {
                    throw new RuntimeException(e);
                }
            }
    )
    .add(Linear.builder().setUnits(768).build())
    .add(Activation::relu)
    .add(Dropout.builder().optRate(0.2f).build())
    .add(Linear.builder().setUnits(5).build())
    .addSingleton(nd -> nd.get(":,0"));

Model model = Model.newInstance("review_classification");
model.setBlock(classifier);

DefaultVocabulary vocabulary = DefaultVocabulary.builder()
    .addFromTextFile(embedding.getArtifact("vocab.txt"))
    .optUnknownToken("[UNK]")
    .build();

int maxTokenLen = 64;
int batchSize = 8;
int limit = Integer.MAX_VALUE;

BertFullTokenizer tokenizer = new BertFullTokenizer(vocabulary, true);
CsvDataset awsDataset = getDataset(batchSize, tokenizer, maxTokenLen, limit);
RandomAccessDataset[] datasets = awsDataset.randomSplit(7, 3);
[...]

PR #2173, opened 2022-11-21
Fix TrainAmazonReviewRanking example for PyTorch

siddvenk:
## Description ##

This PR fixes and issue with using a Predictor under the PyTorch engine in the TrainAmazonReviewRanking transfer learning example. 

Issues: https://github.com/deepjavalibrary/djl/issues/2144, https://github.com/deepjavalibrary/djl/issues/2159

The issue is that using a predictor enforces Inference context on the tensors, so when we do transfer learning it can lead to an EngineException like
```
[INFO ] - forward P50: 308.169 ms, P90: 308.169 ms
Exception in thread "main" ai.djl.engine.EngineException: Inference tensors cannot be saved for backward. To work around you can make a clone to get a normal tensor and use it in autograd.
```

This does cause the pretrained block weights to be modified during transfer learning since they are operating under grad mode. This is the same mechanism we use in the other transfer learning examples (where we use the pretrained block in a new model), so this seems to be the route we recommend to users.

An alternative solution here is to provide a TranslatorOptions that duplicates the output NDArrays from the predictor block. 

Note: With the recent change to include [Translators in datasets](https://github.com/deepjavalibrary/djl/commit/7ceb7cd3a95d89a4672eedd14cd6c53341948c9c), we might be able to leverage that functionality here to replace the lambda block. I don't like having to instantiate a ParameterStore via forward in the lambda block here, but at the moment the example will work out of the box for both MXNet and PyTorch now.

### DIFF
--- a/examples/src/main/java/ai/djl/examples/training/transferlearning/TrainAmazonReviewRanking.java
+++ b/examples/src/main/java/ai/djl/examples/training/transferlearning/TrainAmazonReviewRanking.java
@@ -21,7 +21,6 @@ import ai.djl.basicdataset.tabular.utils.Feature;
 import ai.djl.basicdataset.tabular.utils.Featurizer;
 import ai.djl.engine.Engine;
 import ai.djl.examples.training.util.Arguments;
-import ai.djl.inference.Predictor;
 import ai.djl.metric.Metrics;
 import ai.djl.modality.nlp.DefaultVocabulary;
 import ai.djl.modality.nlp.Vocabulary;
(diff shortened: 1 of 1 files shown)


lanking520 (approved):

examples/src/main/java/ai/djl/examples/training/transferlearning/TrainAmazonReviewRanking.java
```diff
@@ -144,24 +144,21 @@ private static CsvDataset getDataset(
...
-                                                    .full(data.getShape(), 1, DataType.INT64),
-                                            data.getManager()
-                                                    .arange(data.getShape().get(1))
-                                                    .toType(DataType.INT64, false)
-                                                    .broadcast(data.getShape())));
-                        } catch (TranslateException e) {
-                            throw new IllegalArgumentException("embedding error", e);
-                        }
+                        return embedder.forward(
+                                new ParameterStore(),
```
frankfliu, line 153: Can we avoid calling forward manually?

siddvenk: Yes ideally I want to avoid this as well - I am wondering whether we can use the TranslatorOptions mechanism to convert the dataset output to the compatible model input? Haven't looked into it much yet.

Agreed that we should avoid calling forward... I'll investigate some options.

PR #2182, opened 2022-11-24
Update SpProcessor.java

Crazy-LittleBoy:
fix: the class need be 'public'

## Description ##


frankfliu:
All the functionality can be access by `SpTokenizer` and `SpVocabulary`. User might forget to close native resource when directly using `SpProcessor`, that why we try to make it private.

Can you tell me a bit more about your use case? In which case you need `SpProcessor` to be public?

lanking520:
I don't think this is necessary, you can still get it through: https://github.com/deepjavalibrary/djl/blob/master/extensions/sentencepiece/src/main/java/ai/djl/sentencepiece/SpTokenizer.java#L100. Ideally, we do not expect user to use SProcessor directly in case of memory leaks

frankfliu:
Please let us know if you have a specific use case.

PR #2188, opened 2022-11-26
made the sites copyright year dyanamic

agesa3:
## Updated the copyright year to be dynamic ##
Made the copyright year dynamic rather than static,hence the year will be updating itself automatically.

### DIFF
--- a/index.html
+++ b/index.html
@@ -225,7 +225,8 @@
                         <div class="footer-copyright">
                             <div class="container swiper-no-swiping">
                                 <div>
-                                    Copyright 2019 <a class="blue-text text-lighten-3 text-underLine"
+                                   Copyright <span class="an-1">{{ new Date().getFullYear() }}</span>
+                                   <a class="blue-text text-lighten-3 text-underLine"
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -225,7 +225,8 @@ <h5 class="grey-text text-lighten-4 an-1 ">
                         <div class="footer-copyright">
                             <div class="container swiper-no-swiping">
                                 <div>
-                                    Copyright 2019 <a class="blue-text text-lighten-3 text-underLine"
+                                   Copyright <span class="an-1">{{ new Date().getFullYear() }}</span>
```
frankfliu, line 228: Thank for your contribution. 

I think the single year copy right should state the year which the first time it published. I think this should not be dynamic

agesa3: what about having something like 2019-current year?

frankfliu: That should work.

agesa3: Ok let me make that change then.

frankfliu (approved):

Issue #2185, opened 2022-11-25
Support for AdamW Optimizer


adepase:
## Description
Are there plans to support AdamW Optimizer?
Today there is only Adam, but it results less performant than AdamW, as several papers report.
Pytorch supports it, I don't know if this is supported also by other engines.

NLP tasks will certainly benefit from that (but literature shows that also other fields could benefit).

For example, I'm trying to repeat a result shown in https://aclanthology.org/2021.naacl-industry.38.pdf with Distilbert and CLINC150 and I'm missing the 96.3 percentage for about a 4%. I tried changing other hyperparameters and config, but the only difference seems to be in the optimizer (and ok, there is not an exact description of how is the softmax classifier exactly, but I'm rather convinced that the issue could be in the Optimizer and, anyway, AdamW is quite used since many years, so it would be great to be able to use it; and ok, there is maybe another issue in the minibatch management for models trained with variable length - I'll report this apart, but it has minor impact - fraction of percentage, it seems, in the above task).

## References
Original Paper: https://arxiv.org/abs/1711.05101
Pytorch implementation: https://pytorch.org/docs/stable/optim.html
Some results pratically showing why it should be used: https://github.com/egg-west/AdamW-pytorch

adepase:
You're great, guys! I'll test it asap! Thank you!

PR #2206, opened 2022-12-02
Implement AdamW on Pytorch and MXNet

KexinFeng:
## Description ##
https://github.com/deepjavalibrary/djl/issues/2185

It is currently implemented only in PyTorch both MXNet.

Old Description:
It looks like the JNA is not available on MXNet. There is a JNA called `_adamw_update`. But the number of inputs is 5 different than adam or pytorch adamw. The required parameters are different (it requires `eta` ). It needs further investigation on this JNA.

```java
WARNING: JNA: Callback ai.djl.mxnet.engine.MxParameterServer$OptimizerCallback@4f936da8 threw the following exception
ai.djl.engine.EngineException: MXNet engine call failed: MXNetError: Cannot find argument 'rescale_grad', Possible Arguments:
----------------
lr : float, required
    Learning rate
beta1 : float, optional, default=0.899999976
    The decay rate for the 1st moment estimates.
beta2 : float, optional, default=0.999000013
    The decay rate for the 2nd moment estimates.
epsilon : float, optional, default=9.99999994e-09
    A small constant for numerical stability.
wd : float, optional, default=0
    Weight decay augments the objective function with a regularization term that penalizes large weights. The penalty scales with the square of the magnitude of each weight.
eta : float, required
    Learning rate schedule multiplier
clip_gradient : float, optional, default=-1
    Clip gradient to the range of [-clip_gradient, clip_gradient] If clip_gradient <= 0, gradient clipping is turned off. grad = max(min(grad, clip_gradient), -clip_gradient).
, in operator _adamw_update(name="", beta2="0.999", rescale_grad="1.0", epsilon="1.0E-8", wd="0.01", beta1="0.9", clip_gradient="-1.0", lr="0.031622566")
```


api/src/main/java/ai/djl/training/optimizer/Adam.java
```diff
@@ -46,6 +46,7 @@ public class Adam extends Optimizer {
 
     private Map<String, Map<Device, NDArray>> means;
     private Map<String, Map<Device, NDArray>> variances;
+    private boolean adamw;
```
zachgk, line 49: It's probably cleaner to have a separate AdamW class. Especially as it gives a place for separate documentation for it. For simplicity, you can still leave the internal implementations connected and/or have one of them extend the other (or extend a third class)

zachgk (commented):
Do you think it would make sense to also change the default optimizer to AdamW? Or, maybe set the optimizer as AdamW for PyTorch and Adam otherwise to deal with the lack of AdamW support in MXNet

api/src/main/java/ai/djl/training/optimizer/AdamW.java
```diff
@@ -0,0 +1,194 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
frankfliu, line 2:
```suggestion
 * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

integration/src/main/java/ai/djl/integration/tests/training/OptimizerTest.java
```diff
@@ -172,6 +172,36 @@ public void testAdam() {
         }
     }
 
+    @Test
+    public void testAdamW() {
+        if ("MXNet".equals(Engine.getDefaultEngineName())) return;
```
frankfliu, line 177:
```suggestion
        TestRequirements.engine("PyTorch");
```

integration/src/main/java/ai/djl/integration/tests/training/OptimizerTest.java
```diff
@@ -172,6 +173,36 @@ public void testAdam() {
         }
     }
 
+    @Test
+    public void testAdamW() {
+        TestRequirements.engine("PyTorch");
```
zachgk, line 178: Is it possible to remove this TestRequirements? In the integration tests, we have an additional criteria that an `UnsupportedOperationException` does not count as failing the test. The suite could be considered as ensuring that an engine, to what extent it implements the DJL API, does so correctly. Not implementing a part of it is then not considered a failure.

Ideally, can you throw the exception with other engines (ie MXNet) so the test will be fine even for all engines. It also helps to clarify to end users that it is a known gap that the AdamW engine will not work and possibly why (even if the answer is we didn't implement this yet).

frankfliu (approved):

Issue #2210, opened 2022-12-05
memory leak and duration increase during training


enpasos:
## Description
During training on GPU I experience
- 4 MiB memory leak on GPU per epoch (looks constant)
- duration increase about 1 min per epoch (looks linear)

### Expected Behavior
no memory leak, roughly constant duration per epoch

## How to Reproduce?
I set up a toy app based on [djl mnist](https://github.com/deepjavalibrary/djl/blob/master/examples/docs/train_mnist_mlp.md) to reproduce the problem I experience:
```
git clone https://github.com/enpasos/reproducebug1.git
cd reproducebug1
gradlew build
java -jar app/build/libs/app-0.0.1-SNAPSHOT.jar
```

## Environment Info

- GPU: NVIDIA GeForce RTX 3090
- CPU: AMD Ryzen 9 3950X 16-Core Processor
- RAM: 64 GB
- OS: Edition	Windows 11 Pro, Version	22H2, Betriebssystembuild	22623.1020
- GPU Driver: 522.25
- CUDA SDK: 11.6.2
- CUDNN: cudnn-windows-x86_64-8.5.0.96_cuda11
- Java: Corretto-17.0.3.6.1
- DJL: 0.21.0-SNAPSHOT  (05.12.2022)
- PYTORCH: 1.12.1

enpasos:
I also see the **same result on Ubuntu** 20.04.4 on wsl2 which is stacked on the same HW and OS as in the environment info.

enpasos:
Further observations:
- memory leak seem to happen in jumps of 2 MiB after some `trainSplit` calls.
- with MLP-model the same memory leak, but duration does not increase.

enpasos:
Conclusions for me: 
- I found no way to get rid of the memory leak. Therefore I have to restart the JVM before the GPU runs out of memory.
- I found a workaround way to get rid of the duration increase: Downgrade DJL to 0.19.0 and correspondingly CUDA to 11.3.1, PyTorch to 1.11.0.

I would greatly appreciate help in finding and eliminating the cause of the problem.

enpasos:
Narrowing down the **cause for the duration increase** (running time grows quadratically with epoch instead of linear):

- As exchanging in the stack (below) only DJL from 0.20.0 to 0.19.0 resolves the duration increase problem, the cause should be found somewhere in the changes from DJL 0.19.0 to DJL 0.20.0. Stack:
    - GPU: NVIDIA GeForce RTX 3090
[...]

Issue #2024, opened 2022-09-17
pytorch doesn't clear out gradients between `backward()` calls in the gradient collector


ajrnz:
I've been following along the linear regression code in the Dive into Deep Learning book. When using the pytorch engine the gradients are not cleared between backward propagation runs.

For example (this code is scala but that makes no difference):

```scala
val x = mn.arange(4.0f)
x.setRequiresGradient(true)
println(s"x=$x")

for(i <- 0 until 4) {
  Using.resource(Engine.getInstance().newGradientCollector()){ gc =>
      val y = x.dot(x) * 2
      gc.backward(y)
  }
  println(s"gradient=${x.getGradient()}")
}
```

The following code shows:
```
x=ND: (4) cpu() float32 hasGradient
[0., 1., 2., 3.]

gradient=ND: (4) cpu() float32
[ 0.,  4.,  8., 12.]

gradient=ND: (4) cpu() float32
[ 0.,  8., 16., 24.]

gradient=ND: (4) cpu() float32
[ 0., 12., 24., 36.]

gradient=ND: (4) cpu() float32
[ 0., 16., 32., 48.]
```

ie the gradients keep getting summed between runs when using the pytorch engine but the correct output (below) when swapping out the pytorch dependency for the mxnet one (identical code)
```
x=ND: (4) cpu() float32 hasGradient
[0., 1., 2., 3.]

gradient=ND: (4) cpu() float32
[ 0.,  4.,  8., 12.]

gradient=ND: (4) cpu() float32
[ 0.,  4.,  8., 12.]

gradient=ND: (4) cpu() float32
[ 0.,  4.,  8., 12.]

gradient=ND: (4) cpu() float32
[ 0.,  4.,  8., 12.]
```

This is under MacOS both x86 and ARM (CPU).

siddvenk:
This is the expected behavior for PyTorch - by default if a PyTorch Tensor (or in this case a DJL NDArray backed by the PyTorch engine) enables`setRequireGradient`, the gradients are accumulated with calls to `.backward()`. This behavior is documented in the [PyTorch documentation](https://pytorch.org/tutorials/recipes/recipes/zeroing_out_gradients.html). So, multiple calls to backward will result in the behavior you observed here with PyTorch.

[...]

PR #2232, opened 2022-12-15
Specify the block that PtGradientCollector belongs to

KexinFeng:
The zeroGradients() call (added in https://github.com/deepjavalibrary/djl/pull/2101 to solve https://github.com/deepjavalibrary/djl/issues/2024) when constructing PtGradientCollector() compromises the efficiency and memory usage, as shown in https://github.com/deepjavalibrary/djl/issues/2210. Thus the block has to be specified so that the zeroGradients is called only on its parameters, and the search scope is narrowed down. Another relevant PR is https://github.com/deepjavalibrary/djl/pull/2111. 

The original PtGradientCollector() is left unchanged in order to be backward compatible. A new PtGradientCollector(Block) is added so that the gradient collector is restricted to operate on a certain block. Then multiple such collectors can coexist. @zachgk I added this point to the doc, please take a look. The relevant PR https://github.com/deepjavalibrary/djl/pull/2111.

**In essence, this PR is to disable `zeroGradients()` in the default case, unless the block is specified when creating a GradientCollector.** This is because of the issues inside global search in zeroGradients.

---
To solve this issue the issue https://github.com/deepjavalibrary/djl/issues/2024 in the Dive into Deep Learning book, how the gradient is accumulated should be specified in the future, either "write" or "add". Also "retain_graph" should also be fixed. Under certain condition error should be thrown, for example, if retain_graph=false and repetative backward is called. This way the ambiguity in the defination of repetative backward call, as shown in https://github.com/deepjavalibrary/djl/issues/2024, will be clarified.

On pytorch, by default gradients are added for repetative backward call and if retain_graph=true
```python
    x1 = pt.tensor([1., 2., 3., 4.], requires_grad=True)
    x2 = pt.tensor([5., 6., 7., 8.], requires_grad=True)
    u = x1 * x2
    z = u * x1
    z.sum().backward(retain_graph=True)
    print("\n")
    print(x1.grad, x2.grad)

    z.sum().backward(retain_graph=True)
    print(x1.grad, x2.grad)

    z.sum().backward()
    print(x1.grad, x2.grad)
```
Output:
```
tensor([10., 24., 42., 64.]) tensor([ 1.,  4.,  9., 16.])
tensor([ 20.,  48.,  84., 128.]) tensor([ 2.,  8., 18., 32.])
tensor([ 30.,  72., 126., 192.]) tensor([ 3., 12., 27., 48.])
```
---
Note:
* If repetative backward needs to be called, `retain_graph=True` should be set. This is not an commenly seen usecase in practice. See also https://github.com/deepjavalibrary/djl/pull/2111.
* `Block` currently does not have a direct call feature like the following but requires `parameterStore`.
```python
m = nn.Linear(20, 30)
input = torch.randn(128, 20)
output = m(input)
```
---
Relavant stackoverflow issue:
https://stackoverflow.com/questions/70998652/djl-gradientcollector-try-with-resources-initialiser-error

### DIFF
--- a/api/src/main/java/ai/djl/engine/Engine.java
+++ b/api/src/main/java/ai/djl/engine/Engine.java
@@ -15,6 +15,7 @@ package ai.djl.engine;
 import ai.djl.Device;
 import ai.djl.Model;
 import ai.djl.ndarray.NDManager;
+import ai.djl.nn.Block;
 import ai.djl.nn.SymbolBlock;
 import ai.djl.training.GradientCollector;
 import ai.djl.training.LocalParameterServer;
@@ -305,6 +306,16 @@ public abstract class Engine {
      * @return a new instance of {@link GradientCollector}
      */
     public abstract GradientCollector newGradientCollector();
+
+    /**
+     * Returns a new instance of {@link GradientCollector}.
+     *
+     * @param block the block that the gradient collector attaches
+     * @return a new instance of {@link GradientCollector}
+     */
+    public GradientCollector newGradientCollector(Block block) {
+        throw new UnsupportedOperationException("Not supported.");
+    }
 
     /**
      * Returns a new instance of {@link ParameterServer}.
--- a/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtGradientCollector.java
+++ b/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtGradientCollector.java
@@ -15,6 +15,8 @@ package ai.djl.pytorch.engine;
 
 import ai.djl.ndarray.NDArray;
 import ai.djl.ndarray.NDManager;
+import ai.djl.nn.Block;
+import ai.djl.nn.Parameter;
 import ai.djl.pytorch.jni.JniUtils;
 import ai.djl.training.GradientCollector;
 
@@ -37,8 +39,18 @@ public final class PtGradientCollector implements GradientCollector {
                     "A PtGradientCollector is already collecting. Only one can be collecting at a"
                             + " time");
         }
+    }
 
-        zeroGradients();
+    /**
+     * Constructs a new {@code PtGradientCollector} instance.
+     *
+     * @param block the block that the gradient collector belongs. It is seen as a set of trainable
+     *     parameters.
+     */
+    public PtGradientCollector(Block block) {
+        gradModel = JniUtils.isGradMode();
+        JniUtils.setGradMode(true);
+        zeroGradients(block);
     }
 
     /** {@inheritDoc} */
@@ -74,6 +86,20 @@ public final class PtGradientCollector implements GradientCollector {
         for (NDArray array : systemManager.getManagedArrays()) {
             if (array.hasGradient()) {
                 array.getGradient().subi(array.getGradient());
+            }
+        }
+    }
+
+    /**
+     * Sets all the gradients within the block to zero.
+     *
+     * @param block the target block to run zeroGradients on
+     */
+    public void zeroGradients(Block block) {
+        for (Parameter param : block.getParameters().values()) {
+            NDArray array = param.getArray();
+            if (array.hasGradient()) {
+                JniUtils.zeroGrad((PtNDArray) array);
             }
         }
     }
--- a/integration/src/main/java/ai/djl/integration/tests/training/GradientCollectorIntegrationTest.java
+++ b/integration/src/main/java/ai/djl/integration/tests/training/GradientCollectorIntegrationTest.java
@@ -25,6 +25,7 @@ import ai.djl.ndarray.types.Shape;
 import ai.djl.nn.Block;
 import ai.djl.nn.Blocks;
 import ai.djl.nn.Parameter;
+import ai.djl.nn.ParameterList;
 import ai.djl.nn.SequentialBlock;
 import ai.djl.nn.core.Linear;
 import ai.djl.testing.Assertions;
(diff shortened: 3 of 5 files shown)


engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtGradientCollector.java
```diff
@@ -37,8 +39,18 @@ public PtGradientCollector() {
                     "A PtGradientCollector is already collecting. Only one can be collecting at a"
                             + " time");
         }
+    }
 
-        zeroGradients();
```
zachgk, line 41: I think we still want the zeroGradients here. Otherwise, it won't work correctly for cases without a block

api/src/main/java/ai/djl/training/GradientCollector.java
```diff
@@ -28,7 +28,8 @@
  * and added together.
  *
  * <p>Due to limitations in most engines, the gradient collectors are global. This means that only
- * one can be used at a time. If multiple are opened, an error will be thrown.
+ * one can be used at a time. If multiple are opened, an error will be thrown. However, if the
+ * collector is specified to operate on a certain block, then multiple collectors can coexist.
```
zachgk, line 32: You can't actually have multiple (even on separate blocks) based on the engine architectures. Maybe in a future engine or if they make the training mode non-global. Or, we can just treat that as an engine limitation and then throw `UnsupportedOperationExceptions` while the test suite should expect it to work rather than expect it to fail.

api/src/main/java/ai/djl/engine/Engine.java
```diff
@@ -306,6 +307,14 @@ public int getGpuCount() {
...
+    /**
+     * Returns a new instance of {@link GradientCollector}.
+     *
+     * @param block the block that the gradient collector attaches
+     * @return a new instance of {@link GradientCollector}
+     */
+    public abstract GradientCollector newGradientCollector(Block block);
+
     /**
      * Returns a new instance of {@link ParameterServer}.
```
zachgk, line 319: Can you also update `trainer.newGradientCollector()` to use this? You can change it's definition to:

```java
    public GradientCollector newGradientCollector() {
        return manager.getEngine().newGradientCollector(getModel().getBlock());
    }
```

KexinFeng: The tricky part is that inside the newGradientCollector(), the zeroGradient is not called, while inside newGradientCollector(Block block), it is. This is to avoid the efficiency and memory leak issue with global search scope as discussed  last week.

api/src/main/java/ai/djl/training/GradientCollector.java
KexinFeng: Got it. I'll remove it then. Or maybe we can put it to TODO. Here is a relavant usecase: https://stackoverflow.com/questions/70998652/djl-gradientcollector-try-with-resources-initialiser-error

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtGradientCollector.java
KexinFeng: The reason of removing it was mainly due to efficiency and memory issue https://github.com/deepjavalibrary/djl/issues/2210.
Should we add this back and then do
```
    public GradientCollector newGradientCollector() {
        return manager.getEngine().newGradientCollector(getModel().getBlock());
    }
```
Does this avoid the global search scope issue?

integration/src/main/java/ai/djl/integration/tests/training/GradientCollectorIntegrationTest.java
```diff
@@ -107,16 +108,25 @@ public void testZeroGradients() {
...
+            block.initialize(manager, DataType.FLOAT32, inputShape);
+            ParameterList list = block.getParameters();
+            NDArray weight = list.get("weight").getArray();
+            NDArray bias = list.get("bias").getArray();
 
             for (int i = 0; i < 3; i++) {
-                try (GradientCollector gc = Engine.getInstance().newGradientCollector()) {
-                    NDArray b = a.mul(2);
-                    gc.backward(b);
+                try (GradientCollector gc = Engine.getInstance().newGradientCollector(block)) {
```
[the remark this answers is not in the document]
KexinFeng: The test of newGradientCollector(block) has been covered in `testClearGradients`. Do we need to add it or do we change the name of this test to be, e.g. testNewGradientCollectorWithClearGradients?

api/src/main/java/ai/djl/engine/Engine.java
KexinFeng: Another difference is 
```java
        boolean wasCollecting = isCollecting.getAndSet(true);
        if (wasCollecting) {
            throw new IllegalStateException(
                    "A PtGradientCollector is already collecting. Only one can be collecting at a"
                            + " time");
        }
```
is not called inside `PtGradientCollector(Block block)`.

api/src/main/java/ai/djl/training/GradientCollector.java
KexinFeng: Here I was guessing: inside `PtGradientCollector(Block block)`, if the following is not called unlike that in `PtGradientCollector()`, would it allow multiple GradientCollectors to coexist?
```java
        boolean wasCollecting = isCollecting.getAndSet(true);
        if (wasCollecting) {
            throw new IllegalStateException(
                    "A PtGradientCollector is already collecting. Only one can be collecting at a"
                            + " time");
        }
```

api/src/main/java/ai/djl/engine/Engine.java
enpasos: In `zeroGradients` `array.getGradient().subi(array.getGradient());` is still used. 
Couldn't it be replaced with `JniUtils.zeroGrad((PtNDArray) array);` as used in `zeroGradients(Block block)`?

KexinFeng: > As @zachgk mentioned ... in case you go with the approach it must also work when using a trainer.

That would be ideal. But it has the efficient and memory issue for now. So we should disable it for now.

> Are you sure that you get all relevant NDArrays with Gradient by getting the parameter NDArrays from block?

The gradientCollector is specified to a block. By defination it does search only within a block.

PR #2250, opened 2022-12-20
[audio] aarch64 support for audio extension

siddvenk:
## Description ##

updates dependencies and adds aarch64 support for audio extension

### DIFF
--- a/extensions/audio/build.gradle
+++ b/extensions/audio/build.gradle
@@ -3,7 +3,7 @@ group 'ai.djl.audio'
 dependencies {
     api project(":api")
     api project(":basicdataset")
-    api ("org.bytedeco:javacv:1.5.7") {
+    api ("org.bytedeco:javacv:1.5.8") {
         exclude group: "org.bytedeco", module: "opencv"
         exclude group: "org.bytedeco", module: "openblas"
         exclude group: "org.bytedeco", module: "flycapture"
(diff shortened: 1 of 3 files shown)


extensions/audio/build.gradle
```diff
@@ -21,9 +21,17 @@ dependencies {
     }
     api "com.github.wendykierp:JTransforms:3.1"
 
-    runtimeOnly "org.bytedeco:ffmpeg:5.0-1.5.7:macosx-x86_64"
-    runtimeOnly "org.bytedeco:ffmpeg:5.0-1.5.7:linux-x86_64"
-    runtimeOnly "org.bytedeco:ffmpeg:5.0-1.5.7:windows-x86_64"
+    String arch = System.properties["os.arch"] == "aarch64" ? "arm64" : "x86_64"
+    String os = System.properties["os.name"].toLowerCase(Locale.ROOT)
+    if (os.contains("mac")) {
```
frankfliu, line 26: We should not pick dependencies based on current os.
If we publish the package on mac (our current CI), then only mac user can use this package when download from maven.

siddvenk: I see, that makes sense. I modified the change here to also include the aarch64 deps rather than pick on current os

frankfliu (approved):

PR #2273, opened 2022-12-31
optionally close orphaned NDArrays using Java garbage collection

enpasos:
## Description ##

This PR is a solution suggestion to https://github.com/deepjavalibrary/djl/issues/2210.

A striking example of the impact shows the following figure (to reproduce see https://github.com/deepjavalibrary/djl/issues/2210):
 
grafik

Design considerations: [DJLDiscussionInputVersion4.pdf](https://github.com/deepjavalibrary/djl/files/10315716/DJLDiscussionInputVersion4.pdf)

Opt-in:  `SwitchGarbageCollection.on();`

The implementation here is done for PyTorch, but the solution approach is general.


lanking520:
This approach is very similar to reference counting, which is used widely in C++. Here I just leave a reference implementation: https://github.com/almson/almson-refcount will take a look after holiday.

Thanks for your contribution! A great mile to explore something alternative. But I would try to benchmark in concurrent mode, to see what happened in parallel block

lanking520:
Another concern is the implementation on the reference queue: seemed all operation is involved with a synchronous checkQueue operation, what is the impact if we frequently adding/removing resources? How much chances we will be blocked by that

lanking520:
In the proxyMaker implementation you used a UUID random generation (synchronize call) and not picking up the uid. UID is also unique across the resources

enpasos:
> Another concern is the implementation on the reference queue: seemed all operation is involved with a synchronous checkQueue operation, what is the impact if we frequently adding/removing resources? How much chances we will be blocked by that

On the `ReferenceQueue` there are two fast methods `enqueue` and `poll` called once per reference. I wouldn't expect notable blocking from there.

enpasos:
> In the proxyMaker implementation you used a UUID random generation (synchronize call) and not picking up the uid. UID is also unique across the resources

I'm not sure if I got your concern right. It is not necessary to use an UUID here - an Integer used right - would also do. This key just needs to be an object that is unique within the `PtNDArrayProxyMaker`. It gets garbage collected together with the proxy when the user has no more references to the proxy while there still is a reference from the NDManager to the corresponding value of type `NDArray`. If your concern is about the particular UID implementation I could replace it with an Integer counter.

api/src/main/java/ai/djl/ndarray/gc/WeakHashMapWrapper.java
```diff
@@ -0,0 +1,161 @@
...
+public class WeakHashMapWrapper<K, V> implements Map<K, V> {
+
+    private final WeakHashMap<K, V> map = new WeakHashMap<>();
+    private final ReferenceQueue<Object> queue = new ReferenceQueue<>();
+
+    private final Set<WeakReferenceWrapper<K, V>> weakReferenceWrapperSet = new HashSet<>();
+
+    private void checkQueue() {
+        for (Reference<?> ref; (ref = queue.poll()) != null; ) {
+            synchronized (queue) {
```
frankfliu, line 40: This implementation may have big performance impact:
1. every NDArray api call (DynamicInvocationHandler.invoke), will call checkQueue() method
2. in this call, it loop through the reference queue
3. this is synchronized call, will cause thread contention in heavy multithreading case

api/src/main/java/ai/djl/ndarray/BaseNDManager.java
```diff
@@ -72,6 +72,25 @@ public final Device defaultDevice() {
         return getEngine().defaultDevice();
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public void closeSubManagersWithoutResources() {
```
frankfliu, line 77: These two class is not needed.

api/src/main/java/ai/djl/util/NativeResource.java
```diff
@@ -14,58 +14,42 @@
...
-
-    protected final AtomicReference<T> handle;
-    private String uid;
-
-    protected NativeResource(T handle) {
-        this.handle = new AtomicReference<>(handle);
-        uid = handle.toString();
-    }
-
+public interface NativeResource<T> extends AutoCloseable {
```
frankfliu, line 23: Does this need to be interface?

api/src/main/java/ai/djl/nn/norm/LayerNorm.java
```diff
@@ -178,7 +178,7 @@ public void loadMetadata(byte loadVersion, DataInputStream is)
     }
 
     /** The Builder to construct a {@link LayerNorm}. */
-    public static final class Builder {
+    public static class Builder {
```
frankfliu, line 181: This change is not relevant to this PR. Please revert this class.

api/src/main/java/ai/djl/ndarray/gc/WeakReferenceWrapper.java
```diff
@@ -0,0 +1,38 @@
...
+package ai.djl.ndarray.gc;
+
+import java.lang.ref.ReferenceQueue;
+import java.lang.ref.WeakReference;
+
+/**
+ * {@code WeakReferenceWrapper<K, V>} extends a {@link WeakReference}. It uses an object of type K
+ * as the referent has an object property of type V.
+ */
+public class WeakReferenceWrapper<K, V> extends WeakReference<Object> {
```
frankfliu, line 22: This class is unnecessary. You can keep it for now. We discuss after other issue is resolved.

frankfliu:
I looked a bit more in this solution, I think this doesn't really work in multi-threading case.
You are using a global queue to store WeakReference, and trying close the NDArray on any NDArray operations.

Consider there are two threads, NDArray 'x' is created in thread 1, and immediately get GCed.
Thread 2 calling a NDArray operation, and picked up 'x' from the queue and trying to delete 'x', this may lead to crash.

PyTorch internally using reference count, and delete NDArray is not thread-safe. If both thread reduced the count and see it become 0, both thread will try to free the memory and causing segfault. A Tensor in PyTorch maybe sliced internally and java only holding one of its handle. The NDArray must be deleted in the same thread that PyTorch internal Tensor is holding.

enpasos:
@frankfliu Thx a lot for looking into the suggested solution. I will try my best to solve the issues you mentioned tomorrow. 

One comment on the threading issue: If I understand you right it is necessary to have the closing of an NDArray done by the same thread as the creation. If so I would replace the global Queue by a ThreadLocal Queue. 

One more point for now: I must confess that in the heat of the moment during the last merges with the master branch I didn't pay attention to my memory leak test ... I'll try to fix that tomorrow as well.

frankfliu:
ThreadLocal has its own issue. It's hard to clean up ThreadLocal memory. We do see some customer create and distroy thread a lot.

enpasos:
> ThreadLocal has its own issue. It's hard to clean up ThreadLocal memory. We do see some customer create and distroy thread a lot.

What would you suggest?

enpasos:
> One more point for now: I must confess that in the heat of the moment during the last merges with the master branch I didn't pay attention to my memory leak test ... I'll try to fix that tomorrow as well.

I repaired the merge (everythink merged but "A temporary solution to issue 2210 (#2304)"). Memory leak test works now. I will now try to solve the issues mentioned in the comments.

api/src/main/java/ai/djl/nn/norm/LayerNorm.java
enpasos: It is reverted here. I put it into the separate PR https://github.com/deepjavalibrary/djl/pull/2309.

api/src/main/java/ai/djl/ndarray/BaseNDManager.java
enpasos: removed two added methods from NDManager hierarchy that are not necessary

api/src/main/java/ai/djl/ndarray/gc/WeakHashMapWrapper.java
enpasos: I completely agree that checkQueue() must not become a performance bottleneck. 
Fortunately the ReferenceQueue is made such that there is no "looping through the reference queue" on each call. Let's look together at the ReferenceQueue source code: 

grafik

If there are no objects on the queue it immediately returns. If there are objects all objects are removed on checkQueue() such that the next time (without garbage collection inbetween) it again returns immediately.

If for whatever reason the performance becomes worse and not much better as expected by switching on gc, the user may decide not to use this opt-in feature.

api/src/main/java/ai/djl/util/NativeResource.java
enpasos: I think yes. By fassading PtNDArray by a dynamic proxy the whole interface of PtNDArray must be a Java interface. NativeResource<T> is part of it.

enpasos:
> > ThreadLocal has its own issue. It's hard to clean up ThreadLocal memory. We do see some customer create and distroy thread a lot.
> 
> What would you suggest? Give it a try? ... I'll give it a try and keep your comment in mind.

I changed the code to use threadLocal queues.
In PtNDArrayProxyMaker a ThreadLocal holds the WeakHashMapWrappers for all the threads.
A smoketest with 1000 threads does not reveal any problems.

api/src/main/java/ai/djl/ndarray/gc/WeakReferenceWrapper.java
enpasos: My idea was to use a standard `WeakHashMap` for storing the `PtNDArrayImpl` without altering it.
There is to my knowledge no hook in mechanism to use the internal `ReferenceQueue` of the `WeakHashMap` to get informed about no more referenced `PtNDArrayImpl`. Therefore I put a wrapper in front of the `WeakHashMap` with its own `ReferenceQueue`.

frankfliu:
I looked a bit more into this PR, I don't think this solution can fly:
1. clean up NDArray relies on calling other NDArray's methods. If the code only create NDArray but not calling any operators, the NDArray won't be closed
2. The test code actually doesn't really work:
    1. mapSize() doesn't work as expected, when GC happened, the size won't be updated
    2. checkQueue() is not actually called, it triggered because the code manually invoked `mapSize()`, this won't happen in real world.

One workaround could be adding a `NDManager.gc()` function and let user manually call it when needed, it scans the current threads orphan NDArrays and close them. But I personally doesn't like `NDManager.gc()` solution. A better alternative would be implement `NDScope` class:

```
try (NDScope scope = new NDScope()) {
     // Create NDArrays
} // NDArrays created in this scope will be closed automatically
```

enpasos:
> The test code actually doesn't really work:
>     1. mapSize() doesn't work as expected, when GC happened, the size won't be updated
>     2. checkQueue() is not actually called, it triggered because the code manually invoked `mapSize()`, this won't happen in real world.

Looks like you looked at `ai.djl.pytorch.integration.gc.Main`. This Class was actually just a very basic smoke test for me when setting up the solution initially. I would rather delete it from this PR and set up a test suite https://github.com/deepjavalibrary/djl/issues/2290  following the tests in https://github.com/deepjavalibrary/djl/issues/2210. 

However, the `ai.djl.pytorch.integration.gc.Main` still works as expected for me: 
When I run it I see
```
> Task :engines:pytorch:pytorch-engine:Main.main()
[main] INFO ai.djl.pytorch.engine.PtEngine - Number of inter-op threads is 8
[main] INFO ai.djl.pytorch.engine.PtEngine - Number of intra-op threads is 8
[main] INFO ai.djl.pytorch.integration.gc.Main - number of NDArrays in NDManager hierarchy 3
[main] INFO ai.djl.pytorch.integration.gc.Main - number of NDArrays in map used of gc triggered NDArray closing 3
[main] INFO ai.djl.pytorch.integration.gc.Main - reference exists ...
[main] INFO ai.djl.pytorch.integration.gc.Main - weakHashMap size: 3
[main] INFO ai.djl.pytorch.integration.gc.Main - no reference exists, but likely not yet garbage collected ...
[main] INFO ai.djl.pytorch.integration.gc.Main - weakHashMap size: 3
\--- NDManager(e8ac5d45a276, gpu(0)) resource count: 1
    \--- NDManager(e847df9e9799, gpu(0)) resource count: 1
        \--- NDManager(db5ef3d00e38, gpu(0)) resource count: 3
            \--- NDArray(2057224517712, Shape(1))
            \--- NDArray(2057224517184, Shape(1))
            \--- NDArray(2057224517632, Shape(1))
[main] INFO ai.djl.pytorch.integration.gc.Main - no reference exists, and likely garbage collected ...
[main] INFO ai.djl.pytorch.integration.gc.Main - weakHashMap size: 0
\--- NDManager(e8ac5d45a276, gpu(0)) resource count: 1
    \--- NDManager(e847df9e9799, gpu(0)) resource count: 1
        \--- NDManager(db5ef3d00e38, gpu(0)) resource count: 0
\--- NDManager(e8ac5d45a276, gpu(0)) resource count: 1
    \--- NDManager(e847df9e9799, gpu(0)) resource count: 0
\--- NDManager(e8ac5d45a276, gpu(0)) resource count: 0
```

This test starts creating three NDArrays a,b,c. Because `SwitchGarbageCollection.on();` the `PtNDArrayImpl` are wrapped with a dynamic proxy. The user references are actually to the dynamic proxy but the NDManager holds the  `PtNDArrayImpl`. -> `weakHashMap size: 3`

When a,b,c are no longer referenced the dynamic proxies and the uid key are ready to be garbage collected. However, we need to wait for the next minor GC. Therefore still -> `weakHashMap size: 3`

To see a GC we do something we would not do in an application, we encourage GC and wait for 1 s.  And in this case checkQueue got three messages from GC and closes the corresponding NDArrays.  -> `weakHashMap size: 0`
I completely agree that calling checkQueue from size() in this example is something that would not happen in real life.

If I find time today I will set up an example for the testsuite. 

For me, `SwitchGarbageCollection.on();` makes a huge, positive difference in the application I run.

enpasos:
> ...
> 
>     1. clean up NDArray relies on calling other NDArray's methods. If the code only create NDArray but not calling any operators, the NDArray won't be closed
> 
> One workaround could be adding a `NDManager.gc()` function and let user manually call it when needed, it scans the current thread's orphan NDArrays and close them. But I personally don't like `NDManager.gc()` solution. A better alternative would be implement `NDScope` class:
> 
> ```
> try (NDScope scope = new NDScope()) {
>      // Create NDArrays
> } // NDArrays created in this scope will be closed automatically
> ```

If SwitchGarbageCollection.on(); is not called there could be NDArrays that are not closed: a leak. If you suffer from such a situation SwitchGarbageCollection.on(); could help by getting rid of the leaked NDArrays. I agree that the suggested solution relys on a method call to an NDArray on the same thread to do any cleanup. As you suggested one could complement this 'automatic process' with a manual trigger that the user may call in case the 'automatic trigger' does not apply. I could add a suggestion for your NDManager.gc().

Do you think that the NDScope approach works together with the GC?

enpasos:
I added a method gc() to NDManager that explicitly calls checkQueue on WeakHashMapWrapper. Just to show this function, I added a small Main2 corresponding to Main. Both classes should be deleted. A test suite https://github.com/deepjavalibrary/djl/issues/2290 should provide realistic test cases.

enpasos:
As a starting point for a test suite https://github.com/deepjavalibrary/djl/issues/2290  I pasted some examples from the DJL Docs into 
```
git clone https://github.com/enpasos/djlperformancetesting.git
```
and tested for duration and memory:

grafik

enpasos:
I thought about your

> A better alternative would be implement `NDScope` class:
> 
> ```
> try (NDScope scope = new NDScope()) {
>      // Create NDArrays
> } // NDArrays created in this scope will be closed automatically
> ```

I think you could have this approach in principle without GC and without NDManager hierarchy. 
You could add a mechanism to move an object from one NDScope to another enclosing or nested NDScope.
You could have full control over the lifecycle of the NDArray's part that lives on the devices. 
One price you pay for this would be the clinch with the variable scope in Java. Suppose you have a long-lived variable and in an NDScope you create a new NDArray that is automatically attached to the variable while you are doing a calculation where you think of the value of the variable rather than the object. When you exit the NDScope, this new NDArray is closed, the Java object NDArray still exists, and the variable points to it. The next time you use the variable, where you want to work with the value, you get the equivalent to a NullPointerException. The compiler would not help you to find such errors. Maybe it is still better than getting memory leaks in running applications.

frankfliu:
@enpasos 

1. I don't like indeterministic behavior, it gives developer false impression that it works like GC
2. `NDManager.gc()` only looks better to me, but it also a bit misleading, it only works for current thread, need document the behavior
3. NDScope can have nested NDScope, the behavior has been implemented in Javacpp's `PointerScope` and it works pretty nice
4. NDScope can be combined with NDManager hierarchy
5. However NDScope is kind of redundant with `NDManager.gc()` solution, they can co-exist, but will cause confusion.

enpasos:
@frankfliu Thx a lot for looking into this option. I totally agree that deterministic behavior is very desirable.

enpasos:
I am closing this PR in favour of the deterministic solution [#2321](https://github.com/deepjavalibrary/djl/pull/2321).

PR #2279, opened 2023-01-02
[api] Improve hybrid engine operators

frankfliu:
- [integration] Increase test coverage for NDArrays.java
- [api] Improve hybrid engine operators

## Description ##

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
+++ b/api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
@@ -121,12 +121,18 @@ public abstract class NDArrayAdapter implements NDArray {
     /** {@inheritDoc} */
     @Override
     public DataType getDataType() {
+        if (isClosed) {
+            throw new IllegalStateException("Native resource has been release already.");
+        }
         return dataType;
     }
 
     /** {@inheritDoc} */
     @Override
     public Shape getShape() {
+        if (isClosed) {
+            throw new IllegalStateException("Native resource has been release already.");
+        }
         return shape;
     }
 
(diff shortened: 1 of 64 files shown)


api/src/main/java/ai/djl/ndarray/NDArrayAdapter.java
```diff
@@ -435,79 +444,103 @@ public NDArray pow(Number n) {
...
+        intern(manager.from(array));
+        return this;
     }
 
     /** {@inheritDoc} */
     @Override
     public NDArray modi(NDArray other) {
-        return getAlternativeArray().modi(other);
+        NDArray array = getAlternativeArray().modi(alternativeManager.from(other));
+        intern(manager.from(array));
```
lanking520, line 526: This conversion from the alternative manager to original manager is a wasteful of resource? Considering using ONNX GPU and PyTorch CPU at the same time. The heavier the chained NDArray operation have, the slower it will get. I would suggest the output NDArray to stay with alternative.

frankfliu: This is inplace call, the return NDArray is expected to be the same as original.

lanking520: But aren't the inner content are being copied? From operation is a data copying operation from 1 engine to another

frankfliu: Let's say you have a `OrtNDArray`, it contains a `OrtTensor` and `PtNDArray` as alternative array, once you called `ortArray.modi(1)`, the alternative array is updated, but `OrtTensor` is still point to the old one. If you call any operation using `OrtTensor`, the result is unexpected.

lanking520 (approved):

PR #2282, opened 2023-01-03
[docs] Update README for engines

frankfliu:
## Description ##

### DIFF
--- a/docs/engine.md
+++ b/docs/engine.md
@@ -1,12 +1,25 @@
 # Engines
 
-The [Engine](https://javadoc.io/doc/ai.djl/api/latest/ai/djl/engine/Engine.html) is one of the most fundamental classes in DJL. Most of the core functionality in DJL including NDArrays, NDManager, and Models are only interfaces. They form a tree of interfaces with the root as the Engine class.
+The [Engine](https://javadoc.io/doc/ai.djl/api/latest/ai/djl/engine/Engine.html) is one of
+the most fundamental classes in DJL. Most of the core functionality in DJL including NDArrays,
+NDManager, and Models are only interfaces. They form a tree of interfaces with the root as the Engine class.
 
-The implementations of these interfaces are provided by the various engines using a Java service loader. This means that DJL is able to take advantage of much of the performance optimization and hardware support work which has gone on in these engines. As they are updated, DJL can take advantage of the updates. And, DJL is able to freely switch between engines to keep up with performance advancements.
+The implementations of these interfaces are provided by the various engines using a Java service
+loader. This means that DJL is able to take advantage of much of the performance optimization and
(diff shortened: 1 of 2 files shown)


docs/engine.md
```diff
@@ -19,14 +32,29 @@ Currently, the engines that are supported by DJL are:
 - [PaddlePaddle](../engines/paddlepaddle/README.md) - supports basic inference
 - [TFLite](../engines/tflite/tflite-engine/README.md) - supports basic inference
 - [TensorRT](../engines/tensorrt/README.md) - supports basic inference
+- [XGBoost](../engines/ml/xgboost/README.md) - supports basic inference
+- [LightGBM](../engines/ml/lightgbm/README.md) - supports basic inference
 - [DLR](../engines/dlr/README.md) - supports basic inference
+- [Python](https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/README.md) - supports basic inference
```
siddvenk, line 38: I find it a bit confusing to include Python engine here since it's provided primarily for the model server use-case and I don't think users will typically be using it with core DJL. 

Since it is an engine, I understand listing alongside the others here. But, it feels like it's different enough than the other basic inference engines that we may want to include a bit more description here, or omit it from this doc?

docs/engine.md
```diff
@@ -19,14 +32,29 @@ Currently, the engines that are supported by DJL are:
...
-You can also choose the default engine manually. Each engine has a name which can be found in the engine's javadoc or README. You can set the default engine by setting either the "DJL_DEFAULT_ENGINE" environment variable or the "ai.djl.default_engine" Java property. Either one should be set to the name of the desired default engine.
+In order to choose an engine, it must be added into the Java classpath. Usually this means
+additional Maven or Gradle dependencies. Many engines require multiple dependencies be added,
+so look at the engine README for your desired engine to learn what dependencies are necessary.
+
+It is also possible to load multiple engines simultaneously. When DJL starts up, it chooses a
+default engine from the available engines. Most of the API that requires an engine such as
+`NDManager.newBaseManager()` or `Model.newInstance()` will internally use the default engine.
+DJL chooses the default engine based on a ranking of how much we recommend the engine, but before
+that it was chosen at random. For those calls, you can also choose the engine manually by getting
```
siddvenk, line 50: Should we include when (i.e. version number) this behavior changed? Or maybe we omit the detail about how it used to work and just state that we choose based on rank if no override is passed in?

docs/engine.md
frankfliu: Will remove this

docs/engine.md
frankfliu: 0.10.0 has been released for quite some time. Since we move fast, we'd like to document the state of the latest version and ignore the old behaviors.

siddvenk (approved):

PR #2305, opened 2023-01-07
Add support for torch::cuda::empty_cache().

demq:
## Description ##

This PR is based on slack help channel discussion with @frankfliu  regarding an ability to clear the CUDA memory cache for PyTorch engine. Here is the description of the relevant example:

It appears that the allocated GPU memory does not get released even if we "close" the model and the associated managers. It appears that this is just related to PyTorch memory management, where you have to explicitly call torch:cuda:empty_cache()  to release the memory for "competing" external processes.The simplest example would be:

`int slp_time = 10000;
long n_mem = 6*1024*1024*128;

NDManager manager = NDManager.newBaseManager(Device.gpu());
NDArray = manager.create(new Shape(n_mem), DataType.INT64);
array.set(new NDIndex(":"), 1); // no memory leaks

manager.close();
out.print(String.format("--\nSleeping after close for %d ms... ", slp_time));
Thread.sleep(slp_time);
out.println("Done!");`

While the process is "sleeping" after closing the manager, "nvidia-smi" still shows the GPU memory is still allocated, and if I were to run another CUDA program it would not be able to use the entire GPU memory/or the GPU memory would be "cached" on CPU RAM.

This PR adds 'PtNDManager.torchCudaEmptyCache()' that calls the "torch::cuda::empty_cache()" function through JNI.

### DIFF
--- a/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDManager.java
+++ b/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDManager.java
@@ -199,4 +199,12 @@ public class PtNDManager extends BaseNDManager {
             super(null, null);
         }
     }
+
+    public void torchCudaEmptyCache(){
+        if (!this.device.isGpu()) {
+            return;
+        }
+
+        JniUtils.torchCudaEmptyCache();
+    }
 }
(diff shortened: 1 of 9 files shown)


frankfliu (approved):

frankfliu:
@demq I updated your PR, please take a look.

demq:
Thanks a lot @frankfliu , I was about to ask for your help on how to include the CUDA includes during JNI build :) 
I see that making this only a JniUtils static function is sufficient, rather than also adding one to PtNDManager.

Can't wait for this to go to master branch, this functionality will fix an annoying issue we have!

frankfliu (approved):

PR #2320, opened 2023-01-13
add a few openCV image methods

lanking520:
## Description ##

Add a few opencv methods


extensions/opencv/src/main/java/ai/djl/opencv/OpenCVImage.java
```diff
@@ -281,6 +281,43 @@ public List<BoundingBox> findBoundingBoxes() {
                 .collect(Collectors.toList());
     }
 
+    public OpenCVImage bgr2rgb() {
```
frankfliu, line 284: 1. Javadoc
2. Should we return a new OpenCVImage or just replace itself:

```
public void bgr2rgb()
```

frankfliu (approved):

PR #2300, opened 2023-01-06
memory leak in Pt indexing

KexinFeng:
## Description ##

A patch to the memory leak issue https://github.com/deepjavalibrary/djl/issues/2283, https://github.com/deepjavalibrary/djl/issues/2283, https://github.com/deepjavalibrary/djl/issues/2289
As pointed out in https://github.com/deepjavalibrary/djl/issues/2283, the native resources allocated by `ai.djl.pytorch.jni.PyTorchLibrary#torchIndexInit` was not released. This pr fixes it.

PR #2321, opened 2023-01-14
WIP: reference counting

enpasos:
This PR introduces - as suggest by @frankfliu  - reference counting to prevent or get rid of memory leaks on the native side of NDArrays. As suggested by @frankfliu the implementation is derived vom the well established reference counting code of JavaCPP.  It is a deterministic approach and does not rely on Java garbage collection as the approach https://github.com/deepjavalibrary/djl/pull/2273.  

The user interface - in this first attempt - is already easy thanks to the design by JavaCPP author Samuel Audet:
- the feature is opt-in by design
- it has a low-level UI - here in `RCObject` (= reference counted object).
- and a high-level UI - here in `RCScope` (= reference counter scope).

The approach works loosely together with the NDManager hierarchy - with potential to get closer marriaged.


enpasos:
This PR solves the problem https://github.com/deepjavalibrary/djl/issues/2210. As an illustration:

grafik
Example App1 in implementing [introductory DJL Doc Example](https://d2l.djl.ai/chapter_linear-networks/linear-regression-scratch.html) 
```
git clone https://github.com/enpasos/djlperformancetesting.git
cd djlperformancetesting
git checkout refcount
```

To repair the memory leak the only code that had to be changed is introducing a RCScope: 
```
try (RCScope scope = new RCScope()) {
  // the training code running inside a loop
}
```

@KexinFeng The temporary solution https://github.com/deepjavalibrary/djl/pull/2304 that fixed the performance impact of this memory leak could potentially be removed after this PR.

enpasos:
*Garbage collection and reference queues do not provide efficient real-time guarantee of unreachability, while reference-counting provides an alternative mechanism at the cost of slight inconvenience*. [[source ... from netty](https://github.com/netty/netty/wiki/Reference-counted-objects)]

One *cost of slight inconvenience*:  With the high level interface `RCScope` you may be lucky that you just have to set the `RCScope`s right and that's it. But if you are not so lucky  an NDArray that you closed implicitly when leaving a `RCScope` might be needed later. Fortunately you very likely recognize this at programming time when you run the code and get an exception from `NativeResource` when the handle is no more there: "Native resource has been released already". At this point the user needs some help to identify the particular NDArray that is needed later on. One option would be to allow him to start the application again with getting some more info, such as the stacktrace of the NDArray at its creation time.

enpasos:
Performance looks good now. I have looked for hot spots using JProfiler

grafik

None of the methods that consume more than 1% of the time belong to the havily used reference counting code. 
And no memory leak anymore!

KexinFeng:
> @KexinFeng The temporary solution https://github.com/deepjavalibrary/djl/pull/2304 that fixed the performance impact of this memory leak could potentially be removed after this PR.

Thanks for this contribution. Also thanks for mentioning this. But the issue PR https://github.com/deepjavalibrary/djl/pull/2304 originates from the issue https://github.com/deepjavalibrary/djl/issues/2024 regarding the setting of gradient accumulation: add or overwrite. Regarding the subsequent issue https://github.com/deepjavalibrary/djl/issues/2210 you opened, the inefficiency there is due to that the searching scope is too large. So probably this PR still wouldn't heal it.

enpasos:
@KexinFeng Thx for looking into it. I really love how this reference counting deterministically works and solves the memory leak problem ... but the idea comes from your team:  @lanking520 referenced [almson-refcount](https://github.com/almson/almson-refcount/search?q=almson-refcount) and @frankfliu referenced the JavaCPP with the easy to use Scope approach. The approach can work completely without relying on GC and therefore likely avoids problems @zachgk mentioned.

Concerning PR https://github.com/deepjavalibrary/djl/pull/2304 ... you are the expert. I just wanted to mention it because I had the impression that the too large searching scope did not hurt until the memory-leaking NDArrays came into play.

KexinFeng:
@enpasos A quick question about

> This PR can be used to solve the problem https://github.com/deepjavalibrary/djl/issues/2210. As an illustration: [picture]

does this PR also solves the efficiency problem that was found in that same issue https://github.com/deepjavalibrary/djl/issues/2210?

If searching scope being too large were really the root cause, then probably the efficiency issue would still be there. But if not, there might be a different root cause.

enpasos:
> @enpasos A quick question about
> 
> > This PR can be used to solve the problem #2210. As an illustration: [picture]
> 
> does this PR also solves the efficiency problem that was found in that same issue #2210?
> 
> If searching scope being too large were really the root cause, then probably the efficiency issue would still be there. But if not, there might be a different root cause.

@KexinFeng Thx for asking. I didn't test it before ... and with the test I ran into a bug in the `PtNDArrayDeallocator`. After fixing this bug I see

grafik

My interpretation: With the "temporary solution" the performance impact is fixed but the causing memory leak is still there. With "reference counting but not temporary solution" the memory leak is fixed and the main performance impact is fixed, too. The performance numbers of "temporary solution" are slightly better then "reference counting but not temporary solution", so "reference counting with temporary solution" should give the best in terms of memory leak and performance impact.

KexinFeng:
Thanks so much for these test results! It looks like reference counting without temporary solution indeeds solves the issue, the reason of which probably is not known until the true root cause of the previous issue is found. 

From these test results, there are also other bizzare issues. The first is that, temporary the solution https://github.com/deepjavalibrary/djl/pull/2304 is nothing but the revertion of the features PR 2111 and PR2101. So the results indicate that DJL has memory leak issues already before PR 2111 and PR2101. One check is whether it is due to the PtNDArray indexing memory leak issue, which was solved later than this in https://github.com/deepjavalibrary/djl/pull/2300.
On the other hand, if this was indeed the root cause, or if anyother root cause were indentified, then how the reference counting featue solves these issues would be an interesting example to demonstrate its advantage.

The second is that from the efficiency plot, it looks like the global searching doesn't necessarily increase the duration. Probably it is relavant to memory leak, so when reference counting feature is applied, it is solved at the same time.

enpasos:
@KexinFeng The [introductory DJL Doc Example](https://d2l.djl.ai/chapter_linear-networks/linear-regression-scratch.html) as I put it together as an example is for my understanding a from scratch and not a very sophisticated approach - but the memory leaking mistakes made there are on the level I make them - as an example for a DJL user. One "memory leak" there: the `params` belong to the `NDManager manager1`  just below the SystemManager. When the stochastic gradient descent function inside the training loop is called there are intermediate result NDArrays created - such as `param.getGradient().mul(lr)` that also belong to manager1. They are no more used but they are not closed. And it happens again and again on each training loop. With the reference counting scope approach this is easy to solve. 

I think this memory leak is not related to the indexing memory leak issue, solved in https://github.com/deepjavalibrary/djl/pull/2300

KexinFeng:
Hi @enpasos, thanks again for your very nice contribution. We have had several team meetings to discuss your PRs, including this one and the preceding one https://github.com/deepjavalibrary/djl/pull/2273, and we were inspired by your memory management idea and the relevant Java language knowledge. More specifically, we do agree that your implementation of using `RCScope` will work. And, most importantly, this feature indeed frees the users from worrying about which NDManager to attach when doing NDArray computations, which is a great improvement. On the other hand, to keep DJL elegant, we have also simplified your implementation. It turns out that the essential implementation is shown in the following two files: https://gist.github.com/KexinFeng/2ad44daf14f77bc45c7b4c7fd8cf8f33. And since we have `NDManager` already, the reference count won't be essential, thus won't be kept. So this means that a big chunk of your implementation will not be included in this simplified version.

So can we propose the following? We can force-push our simplified version to this PR, so that the feature is credited to you. But before this, you can first backup your current implementation, and if you still want, you can create a new PR and put them there. What do you think?

enpasos:
Perfect! Many thanks to you @KexinFeng  and the DJL team! The solution you found is just right to completely solve my memory leak problems. It is elegant per se and in its integration with DJL. 

Two minor points:
- The hasResource/hasTempResource I added does not always work because equals breaks on NDArray if the handle is null. Perhaps equals could be tweaked to work in this case too.
- NDScopeTest not RCScopeTest

enpasos:
> _Garbage collection and reference queues do not provide efficient real-time guarantee of unreachability, while reference-counting provides an alternative mechanism at the cost of slight inconvenience_. [[source ... from netty](https://github.com/netty/netty/wiki/Reference-counted-objects)]
> 
> One _cost of slight inconvenience_: With the high level interface `RCScope` you may be lucky that you just have to set the `RCScope`s right and that's it. But if you are not so lucky a NDArray that you closed implicitly when leaving a `RCScope` might be needed later on. Fortunately you very likely recognize this at programming time when you run the code and get an exception from `NativeResource` when the handle is no more there: "Native resource has been released already". At this point the user needs some help to identify the particular NDArray that is needed later on. One option would be to allow him to start the application again with getting some more info, such as the stacktrace of the NDArray at its creation time ..(edited 17.01.2023) and the closing time. ... (edited 15.01.2023 after implementing a first solution that I practically used so solve such a problem in an application) Setting the following switch at the start of your app you get the needed info.
> 
> ```
> RCConfig.setVerboseIfResourceAlreadyClosed(true);
> ```

I am missing the feature to find a specific NDArray when I get the error "Native resource has been released already". As this is a feature, I would follow @KexinFeng's suggestion to add a small PR if this PR is on the master.

enpasos:
looks like more problems with unregister ... I will check ...

api/src/test/java/ai/djl/ndarray/NDScopeTest.java
```diff
@@ -0,0 +1,48 @@
...
+                try (NDScope ignore = new NDScope()) {
+                    uninvolved = manager.create(new int[] {1});
+                    uninvolved.close();
+                    inside = manager.create(new int[] {1});
+                    // not tracked by any NDScope, but still managed by NDManager
+                    NDScope.unregister(inside);
+                }
+
+                detached = manager.create(new int[] {1});
+                detached.detach(); // detached from NDManager
```
frankfliu, line 38: 1. In most cases, when `NDArray.detach()` is called, user want to return the NDArray from the function.
2. And usually, NDManager is at the out side if NDScope

Which means user has to call both function in most of time. And we documented that once NDArray is detached, user must manually close it.

So I feel we should unregister from the NDScope when `NDArray.detach()` is called.

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
```diff
@@ -1577,6 +1581,13 @@ public String toString() {
     @Override
     public boolean equals(Object obj) {
         if (obj instanceof NDArray) {
+            // do no compare content if obj is released
+            if (((NDArray) obj).isReleased()) {
+                return this == obj;
+            }
+            if (((NDArray) obj).getManager() != manager) {
```
frankfliu, line 1588: `equals()` compares the content of two NDArray, if their NDManager are different, their content still can be equals

api/src/test/java/ai/djl/ndarray/NDScopeTest.java
frankfliu: Looks like we are abusing `NDArray.detach()` in many internal place (`attache()`, they should only detach from NDManager only. Introducing two `detach(boolean)` function seems too much. Requires user to call both should be fine for now.

enpasos: What is your intention to introduce NDScope?

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
enpasos: NDScope.unregister calls resources.remove which uses equals to identify the object. 
I got an exception from unregister in two cases: one object (not the one to be unregistered) in NDScope was on a different Device and one was already closed. The lines I added - maybe not perfect - were intended to solve both cases.

api/src/test/java/ai/djl/ndarray/NDScopeTest.java
enpasos: My intention would be: Implement it because you need it to close a memory leak that you cannot solve otherwise. So I would not use NDScope if I did not have to. But if I have a leak - which I unfortunately do - I would put the NDScope where I think no NDArray should leak through - no matter if inside or outside of NDManager. And in that case I would not want it to leak through for whatever reason.

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
frankfliu: I see, we should not use `.remove()`, we can call `.pop()` instead

api/src/test/java/ai/djl/ndarray/NDScopeTest.java
enpasos: > Looks like we are abusing `NDArray.detach()` in many internal place (`attache()`, they should only detach from NDManager only. Introducing two `detach(boolean)` function seems too much. Requires user to call both should be fine for now.

I agree that if there were a convenience method where the user exactly knows that she/he also unregister from the NDScope there should be no problem.

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
enpasos: > I see, we should remove the NDArray by it's uid

I think this is not save because the uid from the handle is only unique as long as the handle lives. If you create a uuid by uid plus handleCreationTimestamp it would do.

frankfliu: You are right, uid meant to be unique, but it's possible have a clash if the NDArray is closed and the same memory get reused. We should add a timestamp to it.

enpasos: Do you make the corrections to the equals method?

frankfliu: I should not change `equals()` definition. `equals()` only mean the value are the same. Not the same instance. In NDScope, we must ensure we are deleting the same object. If two NDArray with the same value pushed into NDScope, we might deleted the wrong one.

enpasos: It is an option to solve in NDScope. However, the problems I experience where Exceptions coming from the equals-Method. That is not part of the equals method contract - it should answer yes or no.
If you would like to keep equals as it is. A simple Wrapper of NDArray would do that is only used inside NDScope to put into the list. There you could without an equals implementation just inherit the one from Object. If you go with uid and timestamp equals than the list should be reset by a Set as in a list could exist many NDArrays that are "equal" according to content but the remove method would only remove the first occurance.

frankfliu: Let separate in different issues:
1. NDScope should track the object itself, the ArrayList might not be the right collection, we can use `IdentityHashMap` to resolve this problem
2. uid is not unique. we can solve this problem in a different PR
3. `equals()` function: `isReleased()` is a bug, we can fixed it or in a different PR

enpasos: I comitted a suggestion, but I think it is not correct ... NDArrayWrapper and NDArrayList it a problem ... I will fix it.

api/src/main/java/ai/djl/ndarray/NDScope.java
```diff
@@ -28,7 +28,7 @@ public class NDScope implements AutoCloseable {
     private static final ThreadLocal<Deque<NDScope>> SCOPE_STACK =
             ThreadLocal.withInitial(ArrayDeque::new);
 
-    private List<NDArray> resources;
+    private List<NDArrayWrapper> resources;
```
frankfliu, line 31: We can use IdentityHashMap, we don't need introduce a new NDArrayWrapper class, and we don't need do a linear search.
```suggestion
    private IdentityHashMap<NDArray, NDArray> resources;
```

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
enpasos: Thx. IdentityHashMap works nicely. Better solution than my wrapper suggestion ... and using NDArrayList with the wrapper is simply wrong.

api/src/test/java/ai/djl/ndarray/NDScopeTest.java
enpasos: Should we leave as it is for now?

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
enpasos: 1. done?
Points 2. and 3. in a separate PR?

PR #2334, opened 2023-01-20
[doc] Updates cache menagement adds ONNX and Huggingface cache directory

frankfliu:
## Description ##


carlos-aguayo:
Thank you!

PR #2352, opened 2023-01-27
if native resource closed, on demand info to identify NDArray

enpasos:
If you are trying to use an NDArray that is already closed you get the error message "Native resource has been released already. ".

If you know which NDArray it is - where it was created and where it was closed - you should have no problem solving the problem. But imagine the case where you are not the one trying to use the already closed NDArray, but some generic code does, maybe you have set an NDScope somewhere, so you know who closed it, but still don't know where it was created. In such cases where you do not have a clue it would be helpful if you could set a switch at compile time to get information about the NDArray:
- uid + creation time
- name
- creation stack trace
- closing stack trace

This PR adds code to provide such info if you set the switch `NDScope.setVerboseIfResourceAlreadyClosed(true);`
It is an opt-in feature: If the switch is not set, nothing changes and there is no overhead.

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/NDScope.java
+++ b/api/src/main/java/ai/djl/ndarray/NDScope.java
@@ -26,31 +26,13 @@ public class NDScope implements AutoCloseable {
 
     private static final ThreadLocal<Deque<NDScope>> SCOPE_STACK =
             ThreadLocal.withInitial(ArrayDeque::new);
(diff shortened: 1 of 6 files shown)


enpasos:
Thx. Nice. I think it would be useful to log the uid if (reference == null && TRACK_RESOURCE). Then the user has the now unique ID for the NDArray.

zachgk (approved):

PR #2357, opened 2023-01-30
Fixes PyTorch Java optimizers

zachgk:
PyTorch requires calling "no_autograd_guard{false}" in order to allow the in-place operations required for an optimizer. This moves that call from each individual optimizer into a single call at the end of the gradient collector. This now makes it possible to implement a Java optimizer (like is done in the D2l book).

To verify this, a new test is added that checks that a Java optimizer with in-place operations will work.

Resolves the concerns raised in #2220 by @blue-me.

### DIFF
--- a/engines/pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_optim.cc
+++ b/engines/pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_optim.cc
@@ -22,7 +22,6 @@ JNIEXPORT void JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_adamUpdate(JNIEnv*
     jfloat weight_decay, jfloat rescale_grad, jfloat clip_grad, jfloat beta1, jfloat beta2, jfloat eps,
     jboolean adamw) {
   API_BEGIN()
-  torch::autograd::AutoGradMode no_autograd_guard{false};
   const auto* weight_ptr = reinterpret_cast<torch::Tensor*>(jweight);
   const auto grad = reinterpret_cast<torch::Tensor*>(jgrad)->clone();
   const auto* mean_ptr = reinterpret_cast<torch::Tensor*>(jmean);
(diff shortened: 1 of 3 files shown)


engines/pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_optim.cc
```diff
@@ -52,7 +51,6 @@ JNIEXPORT void JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_sgdUpdate(JNIEnv*
     jfloat momentum) {
   API_BEGIN()
   // disable gradient calculation
```
KexinFeng, line 53:
```suggestion
```

KexinFeng:
I have a native question about calling the `no_autograd_guard`. If the users implement their own java optimizer, then the torch optimizer won't  be called, then the `no_autograd_guard` inside individual optimizer won't be called. So roughly it seems that `no_autograd_guard` is not getting in the way of implementing customized java optimizer.  For example in the unit test in this PR, the optimizer.update doesn't seem to be called.

engines/pytorch/pytorch-native/src/main/native/ai_djl_pytorch_jni_PyTorchLibrary_system.cc
```diff
@@ -58,6 +58,9 @@ JNIEXPORT void JNICALL Java_ai_djl_pytorch_jni_PyTorchLibrary_torchSetGradMode(
   API_BEGIN()
 #if !defined(__ANDROID__)
   c10::GradMode::set_enabled(enable);
+  if (!enable) {
+    torch::autograd::AutoGradMode no_autograd_guard{false};
```
frankfliu, line 62: This doesn't seem has any effect

pdradx:
I have several custom Java optimizers and don't see a problem.
All weight updates routines works without these changes.
The only problem you may encounter if you try to use optimizer INSIDE the scope of Gradient collector - i.e. inside computation graph in autogradient mode. But optimizers should do their work after gradients are computed and GradientCollector is closed.

PR #2386, opened 2023-02-09
Adds back support for getting managed Arrays

zachgk:
The support was removed as part of "A temporary solution to issue 2210  (#2304)". This then reverts that commit while making a smaller change. It keeps the tests designating the desired behavior but notes the problematic engines that fail them.


examples/src/test/java/ai/djl/examples/training/TrainAirfoilWithTabNetTest.java
```diff
@@ -25,6 +25,7 @@ public class TrainAirfoilWithTabNetTest {
 
     @Test
     public void testTrainAirfoilWithTabNet() throws TranslateException, IOException {
+        TestRequirements.nightly();
```
frankfliu, line 28: We already reduced continuous build epoch, should be OK for continuous build now

frankfliu (approved):

PR #2466, opened 2023-03-16
[spark] Add audio predictors

xyang16:
## Description ##


lanking520:
This will not work if user passed in a large audio chunks. I would suggest to build logic on how to split chunks in the DJL into 5-10 second pieces. Some concrete example like this to refer:

https://github.com/gradient-ai/Whisper-AutoCaption

lanking520 (approved):

PR #2480, opened 2023-03-25
[api] Read telemetry metadata from file for sagemaker endpoint

xyang16:
## Description ##

IMDSv2 doesn’t work as sagemaker block IMDSv2 calls due to security reason. So if running on sagemaker, we can read instance id from sagemaker infra metadata file. A sample metadata file looks like this:

```
{
    "ResourceId": "i-0c924c144107d9273",
    "VariantName": "variant1",
    "EndpointArn": "arn:aws:sagemaker:us-east-1:125045733377:endpoint/opt30b-djl20-ds-2023-03-25-00-06-00-824-endpoint"
}
```

### DIFF
--- a/api/src/main/java/ai/djl/util/Ec2Utils.java
+++ b/api/src/main/java/ai/djl/util/Ec2Utils.java
@@ -15,20 +15,20 @@ package ai.djl.util;
 import ai.djl.engine.Engine;
 
 import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.Reader;
 import java.net.HttpURLConnection;
 import java.net.Proxy;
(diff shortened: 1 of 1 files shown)


api/src/main/java/ai/djl/util/Ec2Utils.java
```diff
@@ -37,6 +44,26 @@ public final class Ec2Utils {
 
     private Ec2Utils() {}
 
+    /**
+     * Returns the SageMaker endpoint metadata for specified key.
+     *
+     * @param key the key to retrieve
+     * @return the SageMaker endpoint metadata for specified key
+     */
+    public static String readEndpointMetadata(String key) {
```
frankfliu, line 53: We need update `readMetada()` function. readMetada() should be support for SM as well.

We need to get instance type with readMetada() function

xyang16: OK.

frankfliu (approved):

PR #2502, opened 2023-04-03
Add encode and decode to Input and Output

zachgk:
This adds a full serialization rather than just part of content or properties to the Input and Output. It also requires adding equals and hash to them as well.

### DIFF
--- a/api/src/main/java/ai/djl/modality/Input.java
+++ b/api/src/main/java/ai/djl/modality/Input.java
@@ -19,16 +19,25 @@ import ai.djl.ndarray.NDManager;
 import ai.djl.util.Pair;
 import ai.djl.util.PairList;
 
+import java.io.ByteArrayOutputStream;
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.TreeMap;
 
 /** A class stores the generic input data for inference. */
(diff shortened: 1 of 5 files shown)


api/src/main/java/ai/djl/modality/Input.java
```diff
@@ -328,6 +352,113 @@ public NDList getAsNDList(NDManager manager, int index) {
...
+                }
+                contentRead += newRead;
+            }
+            input.add(key, contents);
+        }
+    }
+
+    /** {@inheritDoc} * */
+    @Override
+    public boolean equals(Object o) {
```
frankfliu, line 444: This is too heavy, we either use a id or just use Identity

api/src/main/java/ai/djl/modality/Output.java
```diff
@@ -70,4 +79,69 @@ public String getMessage() {
...
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        if (!super.equals(o)) {
+            return false;
+        }
+        Output output = (Output) o;
+        return code == output.code && Objects.equals(message, output.message);
```
frankfliu, line 139: We cannot just based on message and code

zachgk: It has a check based on `super.equals()` a few lines above. This is the IntelliJ equals template

api/src/main/java/ai/djl/modality/Input.java
zachgk: I moved it to a separate `deepEquals()` method. So, we can keep the functionality but mostly use the lighter Object equals

frankfliu (approved):

PR #2506, opened 2023-04-06
[spark] Add text2text generation

xyang16:
## Description ##

Add text2text generation, see example: https://github.com/deepjavalibrary/djl-demo/pull/331

### DIFF
--- a/extensions/spark/setup/djl_spark/task/text/text2text_generator.py
+++ b/extensions/spark/setup/djl_spark/task/text/text2text_generator.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+#
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file
+# except in compliance with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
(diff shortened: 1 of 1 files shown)


extensions/spark/setup/djl_spark/task/text/text2text_generator.py
```diff
@@ -0,0 +1,58 @@
...
+        self.model_name = model_name
+
+    def generate(self, dataset, **kwargs):
+        """
+        Performs text2text generation on the provided dataset.
+
+        :param dataset: input dataset
+        :return: output dataset
+        """
+        if not self.model_url and not self.model_name:
```
zachgk, line 46: Maybe check for not both

xyang16: Like: `if not any([self.model_url, self.model_name])`?

zachgk: Basically xor. Must have one or the other but not both. `bool(self.model_url) ^ bool(self.model_name)` should work.

xyang16: I was assuming both should be fine. Just model_url has higher priority than model_name.

zachgk: It is fine, but not great. For cases like this, I tend to ask whether a user would expect that this is the behavior of the API. If they did provide both, would they expect that priority? Instead, an error is clearer because having both properties when only a single one is used already means something is not right. The error will highlight that for the user and ensure they can clarify the situation rather then risk behaving in ways that they don't expect

zachgk (approved):

PR #2508, opened 2023-04-07
Simplify JNI

KexinFeng:
Simplify the JNI where guard appears


zachgk (approved):

frankfliu (commented):
This doesn't work.

If you want to simplify it, you cannot use Guard design pattern.

PR #2509, opened 2023-04-07
POC of LLMDecoder on both Pt trace model and Onnx model (LMAdapter part)

KexinFeng:
Interface `stepGenerator` will be renamed to `LMAdapter`

Description
---
This is a POC which shows that the GPT2 can be traced and exported into DJL for a step inference. Moreover, this step inference allows the following:
1. batch sequence input
2. cached past_key_values input, which is a Tuple[Tuple[Tensor, Tensor], Tuple[Tensor, Tensor], ...] in python.

In this POC, both GPT2_init.pt and GPT2.pt are used since they have different inputs, the former not having past_key_values input.

Run `TextGeneration.java` to test it.

Design
---
The new classes are the following:
```java
public class JavaDecoder {
    private StepGenerator generator;
    
    private DecodeParam decodingParams;
    
    public Text generateText() {};
}
```

JavaDecoder is made engine agnostic. The inference loop can be implemented inside generateText().

```java
public interface StepGenerator {
    private String modelUrl;  // used to load the model
    
    public Token stepGen(inputIds, positionIds, attentionMask, pastKeyValues);  
    // This function is the same as P(w_n|w_{n-1}, w_{n-2}, ...)
}
```
This interface is a wrapper over the model files from different sources, e.g. gpt2.pt, gpt2.onnx, etc.
It can be seen as a Java abstraction of a causal language model, i.e. the conditional probability `p_\theta(v_t | x_1, ..., x_{t-1})`, i.e. given the past tokens up to a certain time x_{< t}, the probability that the next token is v, taken from a vocabulary set V. \theta is the model's weight.

It will be implemented individually to adapt for different models, which applies to different scenarios. 

1. Traced GPT2.pt  + JNI_for_pytorch. Its POC is done.
2. GPT2.onnx + OnnxEngine. This will be the use case corresponding to the graph above.
3. GPT2 + nueronx 
4. FasterTransformer + tensorRTEngine

Model tracing
---
The onnx model gpt2.onnx is loaded from https://huggingface.co/docs/optimum/main/en/exporters/onnx/usage_guides/export_a_model#exporting-a-model-using-past-keysvalues-in-the-decoder.
See also https://github.com/huggingface/optimum/releases.

The gpt2.pt is traced with the following scripts: https://gist.github.com/KexinFeng/4876c6bfb27f40abffe4d5a92c02acff

### DIFF
--- a/api/src/main/java/ai/djl/nn/Block.java
+++ b/api/src/main/java/ai/djl/nn/Block.java
@@ -21,6 +21,7 @@ import ai.djl.ndarray.types.Shape;
 import ai.djl.nn.convolutional.Conv2d;
 import ai.djl.training.ParameterStore;
 import ai.djl.training.initializer.Initializer;
+import ai.djl.util.NativeResource;
 import ai.djl.util.PairList;
 
 import java.io.DataInputStream;
@@ -126,6 +127,8 @@ public interface Block {
     default NDList forward(ParameterStore parameterStore, NDList inputs, boolean training) {
         return forward(parameterStore, inputs, training, null);
     }
+
+    NativeResource<Long> forward(NativeResource<Long>[] inputs);
 
     /**
      * Applies the operating function of the block once. This method should be called only on blocks
--- a/api/src/main/java/ai/djl/translate/LMAdapter.java
+++ b/api/src/main/java/ai/djl/translate/LMAdapter.java
@@ -14,7 +14,6 @@ package ai.djl.translate;
 
 import ai.djl.ndarray.NDList;
 import ai.djl.ndarray.NDManager;
-import ai.djl.util.NativeResource;
 
 /**
  * This is a wrapper over the model files from different sources, e.g. gpt2.pt, gpt2.onnx, etc. This
@@ -26,7 +25,7 @@ import ai.djl.util.NativeResource;
  * range(|inputIds|). This means for each i, the output probability is conditional on the past
  * sequence up to i.
  */
-public interface StepGenerator extends AutoCloseable {
+public interface LMAdapter extends AutoCloseable {
 
     /**
      * @param input input
@@ -34,14 +33,7 @@ public interface StepGenerator extends AutoCloseable {
      * @param manager manager
      * @return CausalLMOutput
      */
-
-    // Will be deprecated
-    default CausalLMOutput stepGeneration(
-            NDList input, NativeResource<Long> pastKeyValues, NDManager manager) {
-        return null;
-    }
-
-    default CausalLMOutput stepGeneration2(NDList ndList, NDList pastKeyValues, NDManager manager) {
+    default CausalLMOutput forward(NDList input, NDList pastKeyValues, NDManager manager) {
         return null;
     }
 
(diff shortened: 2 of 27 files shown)


api/src/main/java/ai/djl/translate/LMAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+ * This is a wrapper over the model files from different sources, e.g. gpt2.pt, gpt2.onnx, etc. This
+ * interface is an abstraction of the causal language model, which in essence is a conditional
+ * probability function: p_\theta(v_t | x_{< t})}, v_t \in V, i.e. given the past tokens up to a
+ * certain time x_{< t}, the probability that the next token is v, taken from a vocabulary set V.
+ * \theta is the model's weight. This function can take an input sequence `inputIds`, whose length
+ * can be greater than one. In this case, the output is still p_\theta(v_i | x_{< i})}, i in
+ * range(|inputIds|). This means for each i, the output probability is conditional on the past
+ * sequence up to i.
+ */
+public interface LMAdapter extends AutoCloseable {
```
zachgk, line 28: Per our discussion, see if you can remake the LMAdapter to be a type of `Block`. Then, you can load it using `Model.load()` rather than requiring the special handling in `Engine`.

api/src/main/java/ai/djl/nn/Block.java
```diff
@@ -127,6 +128,8 @@ default NDList forward(ParameterStore parameterStore, NDList inputs, boolean tra
         return forward(parameterStore, inputs, training, null);
     }
 
+    NativeResource<Long> forward(NativeResource<Long>[] inputs);
```
zachgk, line 131: Instead of adding these additions that may not apply to most of the `Block`s, can you instead modify `GPT2PtLMAdapter` to cast from `Block` to `PtSymbolBlock` before making the `IValue` forward call?

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/IValueUtils.java
```diff
@@ -66,7 +66,7 @@ public static NDList forward(PtSymbolBlock block, NDList inputs, boolean isTrain
      * @param inputs the input {@link IValue}
      * @return the result {@link IValue}
      */
-    public static IValue forward(PtSymbolBlock block, IValue... inputs) {
+    public static IValue forward(PtSymbolBlock block, IValue[] inputs) {
```
zachgk, line 69: I don't understand this change. You should be able to pass in an array of data to the `Ivalue... inputs` anyway

examples/src/main/java/ai/djl/examples/inference/TextGeneration.java
```diff
@@ -0,0 +1,163 @@
...
+public final class TextGeneration {
+
+    private TextGeneration() {}
+
+    public static void main(String[] args) {
+        mainOnnx(args);
+        mainPt(args);
+    }
+
+    public static void mainOnnx(String[] args) {
```
zachgk, line 37: I'm noticing that these examples are almost identical. Is it possible to refactor out a common helper. So, each one would just initialize the `LMAdapter` and then pass it into the helper. Maybe the helper would look something like `public static void generateInternal(LMAdapter generator);`

zachgk (changes requested):

KexinFeng:
Merged in a different PR.

PR #2518, opened 2023-04-10
Adds a way to check if streaming is supported

zachgk:
## Description ##


frankfliu (approved):

integration/src/main/java/ai/djl/integration/tests/inference/StreamingTest.java
```diff
@@ -49,6 +49,7 @@ public void testSequential() throws TranslateException {
 
             try (Predictor<Double, DoubleStream> predictor =
                     model.newPredictor(new TestTranslator())) {
+                Assert.assertTrue(predictor.supportsStreaming());
```
lanking520, line 52: does python engine support streaming by default?

Issue #2558, opened 2023-04-20
UnsupportedOperationException: Expected layout: NCHW, but got: ?????


imkoe:
## Description
(A clear and concise description of what the bug is.)
I want to use DJL to write a cat and dog classification, I used `RandomAccessDataset`。
But when I read the image and convert it to `NDArray` format.
I found that the getShape inside did not have layout values。

```
    @Override
    public Record get(NDManager manager, long index) throws IOException {

        int idx = Math.toIntExact(index);
        Path path = this.imagePaths.get(idx);
        ImageFactory instance = ImageFactory.getInstance();
        Image image1 = instance.fromFile(path);


        NDArray ndArray1 = image1.resize(96, 96, true)
                .toNDArray(manager, Image.Flag.GRAYSCALE)
                .toType(DataType.FLOAT32, false)
//                .transpose(2, 0, 1)
                .div(255.0).expandDims(0);

        NDList dataList = new NDList();
        dataList.add(ndArray1);

        System.out.println(ndArray1.getShape().getLayout());
        System.out.println(dataList.getShapes());
...
```
Layout was  UNKOWN.
![image](https://user-images.githubusercontent.com/19381461/233372855-4901e82c-0976-4525-8338-2cdb6778e8ef.png)

I found that there is no corresponding layout value set here，Is there a problem with my code  or is it a version error。

### Expected Behavior
(what's the expected behavior?)

### Error Message
(Paste the complete error message, including stack trace.)

## How to Reproduce?
(If you developed your own code, please provide a short script that reproduces the error. For existing examples, please provide link.)

### Steps to reproduce
(Paste the commands you ran that produced the error.)

1.
2.

## What have you tried to solve it?

1.
2.

## Environment Info

Please run the command `./gradlew debugEnv` from the root directory of DJL (if necessary, clone DJL first). It will output information about your system, environment, and installation that can help us debug your issue. Paste the output of the command below:

```
PASTE OUTPUT HERE
```

imkoe:
i used `ImageFolder` load folder dateset, and i set 
```
Shape inputShape = new Shape(32, 3, 96, 96);
// initialize trainer with proper input shape
trainer1.initialize(inputShape);
EasyTrain.fit(trainer1, 15, trainerDataSet, testDataSet);
``
running.
thank u.
I think I need to further understand the principles of the RandomAccessDataset

PR #2562, opened 2023-04-21
[fix] Layout initiailization

KexinFeng:
https://github.com/deepjavalibrary/djl/issues/2558


frankfliu (commented):
Layout information is optional. We just provide a way to user the add the information for their own usage.

For the image here, the NDArray is actually HDC, we do transpose based on the pipeline. We don't really need to add layout information here.

PR #2557, opened 2023-04-20
Greedy search and beam search

KexinFeng:
## Description
This PR succeeds PR #2547 and #2509.  The model tracing is shown therein.

Benchmarked with [huggingface transnformers'](https://github.com/huggingface/transformers) output.

Ref. https://huggingface.co/blog/how-to-generate

## Demo output
In the demo TestLMSearch.java, we feed in batch sequence input, using right padding with the space token ' ' (id = 220).
```
["DeepMind Company is",  
 "Memories follow me left and right. I can"]
```
Output of beam search (numBeam=3, maxLength=50):
```
'DeepMind Company is      \xa0 a company that has been around for a long time and has been around for a long time. They have been around for a long time and have been around for a long time. They have been'
'DeepMind Company is      \xa0 a company that has been around for a long time. It has been around for a long time and has been around for a long time. It has been around for a long time and has been'
'DeepMind Company is      \xa0 a company that has been around for a long time and has been around for a long time. They have been around for a long time and have been around for a long time. They are a'

"Memories follow me left and right. I can't tell you how many times I've been told that I'm not a good person. I'm not a good person. I'm not a good person. I'm not a good person. I"
"Memories follow me left and right. I can't tell you how many times I've been told that I'm not a good person. I'm not a good person. I'm not a good person. I'm not a good person.\n"
'Memories follow me left and right. I can\'t tell you how many times I\'ve been told that I\'m not a good person. I\'m not a good person. I\'m not a good person. I\'m not a good person."\n'
```
Output of greedy search (maxLength=50):
```
'DeepMind Company is      \xa0 a company that has been around for over 20 years. We have been around for over 20 years and have been around for over 20 years. We have been around for over 20 years and have been'

"Memories follow me left and right. I can't remember the last time I saw a girl in a dress. I can't remember the last time I saw a girl in a dress. I can't remember the last time I saw a girl in"
```

Notes about the GPT2's behaviour with padding and attention mask:
---
This notes shows that for GPT2, the right padding and left padding can behave very differently, either with attention_mask or without. (The situation with the attention_mask is the most intuitive method). The reason is not totally interpretible yet. But this result will guide the next batching solution.

1. With attention mask which is set 0 on the padded tokens, and 1 everywhere else, for the input_ids that correpsonds to the above input (the space token ' ' id is 220).

  A. Right padding:
```
input_ids = torch.tensor([[29744, 28478, 5834, 318, 220, 220, 220, 220, 220, 220],
[...]

PR #2547, opened 2023-04-17
Contrastive Search

KexinFeng:
## Description
This PR succeeds PR https://github.com/deepjavalibrary/djl/pull/2509.  The model tracing is shown therein.

It implements contrastive search algo based on torchscript gpt2 model. The onnx model support waits for the issue https://github.com/huggingface/optimum/issues/972 to be solved.

Benchmarked with huggingface transformers' output.

Ref.
https://huggingface.co/blog/introducing-csearch

## Demo output
In the demo TestLMSearch.java, we feed in batch sequence input, using right padding with the space token ' ' (id = 220).
```
["DeepMind Company is",  
 "Memories follow me left and right. I can"]
```
Output (topk = 3, maxLength = 50):
```
'DeepMind Company is      \xa0\na small startup that aims to build a better understanding 
of neural networks and how they work. We are currently developing a simple way to create
 a new type of machine learning model that can be used'

"Memories follow me left and right. I can't remember what happened last night, but I know 
that it was very sad. I don't know how long I've been here, but I'm sure there's something
 wrong with me. I'm"
```
The output successfully avoids the repetitive token output, as expected in Ref. https://huggingface.co/blog/introducing-csearch.

Model tracing
---
The onnx model gpt2.onnx is loaded from https://huggingface.co/docs/optimum/main/en/exporters/onnx/usage_guides/export_a_model#exporting-a-model-using-past-keysvalues-in-the-decoder.
See also https://github.com/huggingface/optimum/releases.

The gpt2.pt is traced with the following scripts: https://gist.github.com/KexinFeng/4876c6bfb27f40abffe4d5a92c02acff

PR #2572, opened 2023-04-27
Batch the sequences with ContrastiveSeqBatchScheduler

KexinFeng:
This PR succeeds PR #2547 #2509 #2557, which contains the benchmark outputs of the searching results.

The traced models can be downloaded here: https://d2l-java-resources.s3.amazonaws.com/tmp_model_unittest/model/nlp/text_generation/ai/djl/pytorch/gpt2/metadata.json and https://d2l-java-resources.s3.amazonaws.com/tmp_model_unittest/gpt2_decoder_model_merged.onnx.gz

Design
---
**Overview**

These PRs consists of the following three parts: LMAdapter, LMSearch and SeqBatchScheduler. There are three testing code corresponding to them: `TestLMAdapter.java`, `TestLMSearch.java`, `TestSeqBatchScheduler.java`. The `TestSeqBatchScheduler.java` utilizes all previous features.

**New classes**

```java
/* 
This is a scheduler, serving as an API to the consumer of the systme, allowing for three major actions: initForward, addBatch, fastForward, collectResults. An optimal 
control sequence should be solved, after taking into the time consumption of each action, the batch size and 
sequence length of queueing requests.  Such optimal control solver needs additional effort. Primitive policy is 
setting several thresholds.
*/
SeqBatchScheduler {

    LMAdapter lmAdapter;

    SeqBatcher seqBatcher;

    public SeqBatchScheduler(LMAdapter lmAdapter) {
        this.lmAdapter = lmAdapter;
    }

    public abstract SeqBatcher initForward(
            NDArray inputIds, NDArray batchUids, SearchConfig config);

    public void incrementForward(int count) {}

    public void addRequest(NDArray inputIds, NDArray batchUids, SearchConfig config) {}

    public ConcurrentHashMap<Long, NDArray> collectResults() {}
}
```

```java
// This stores the search state (BatchList), the control variables (eg seqLength, offSets, etc), and batch operations on BatchList.
public class SeqBatcher {

    NDManager manager;

    long batchSize;

    long seqLength;

    NDArray batchUid;

    NDArray offSets;

    BatchList data;

    ConcurrentHashMap<Long, NDArray> finishedSequences;

    List<Pair<Long, Long>> exitIndexEndPosition;

    SeqBatcher(BatchList data, NDArray batchUid, NDArray offSets, NDManager manager) {}

    /** Add new batch. Modify the batch dimension and the left padding. */
    public void addBatch(SeqBatcher seqBatcherNew) {   }

    /** Merge two batchers together. Modify the batch dimension and the left padding. */
    private void merge(SeqBatcher seqBatcher1, SeqBatcher seqBatcher2, long seqDelta) {}

    /** Check which batch needs to exit, according certain criteria like EOS or maxLength. */
    public void exitCriteria(NDArray outputIds, long maxLength, long eosTokenId) {}

    /** Collect the finished sequences and trim the left padding. */
    public void collectAndTrim() { }
}
```

```java
// BatchList is a struct consisting of NDArrays, whose first dimension is batch, and also contains 
// sequence dimension (whose position in tensor's shape is specified by seqDimOrder)
// It represents a search state, and the NDArrays inside are updated in each iteration of the autoregressive loop.
// The batch operators in SeqBatcher will be applied on them.
public abstract class BatchList {

    // [batch, seq_past]. seq-dim-size == |past_seq| + |inputIds|. Will grow.
    public NDArray pastOutputIds;

    // [batch, seq_past]
    // The cache of past attentionMask. seq-dim-size == |past_seq| + |inputIds|. Will grow.
    public NDArray pastAttentionMask;

    // (k, v) * numLayer,
    // kv: [batch, heads, seq_past, kvfeature]
    // The cache of past sequence. seq-dim-size == |past_seq| + |inputIds|. Will grow.
    public NDList pastKeyValues;

    // Sequence dimension order among all dimensions for each element in the batch list.
    public long[] seqDimOrder;

    BatchList() {}

    BatchList(NDList list, long[] seqDimOrder) {}

    BatchList(
            NDArray pastOutputIds,
            NDArray pastAttentionMask,
            NDList pastKeyValues,
            long[] seqDimOrder) {}

    public abstract BatchList fromList(NDList inputList, long[] seqDimOrder);

    // The pastOutputIds has to be the first in the output list
    public abstract NDList getList();

    public long[] getSeqDimOrder() {
        return seqDimOrder;
    }
   } 
}
```

This is partly adapted from the searching algorithms in `TestLMSearch.java`. Currently only the contrastiveSearch has been adapted. The greedySearch and beamSearch will be similarly implemented. On top of that, batching operations are implemented. The system is designed as following, which contains the batching operations like addRequest() and collectAdnTrim()


api/src/main/java/ai/djl/translate/SeqBatchScheduler.java
```diff
@@ -0,0 +1,144 @@
...
+import ai.djl.ndarray.NDArray;
+import ai.djl.ndarray.NDManager;
+import ai.djl.ndarray.index.NDIndex;
+import ai.djl.ndarray.types.DataType;
+import ai.djl.ndarray.types.Shape;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+// This is a scheduler, serving as an API to the consumer of the systme, allowing for three major
```
zachgk, line 12: *system

api/src/main/java/ai/djl/translate/SeqBatchScheduler.java
```diff
@@ -0,0 +1,144 @@
...
+     */
+    public abstract SeqBatcher initForward(
+            NDArray inputIds, NDArray batchUids, SearchConfig config);
+
+    /** Go forward for a given number of iterations. */
+    public void incrementForward(int count) throws NullPointerException {
+        int i = 0;
+        while (i++ < count) {
+            if (seqBatcher == null || seqBatcher.getData() == null) {
+                System.out.println(
```
zachgk, line 45: Should probably throw an Exception here

api/src/main/java/ai/djl/translate/LMSearch.java
```diff
@@ -0,0 +1,617 @@
...
+    public NDArray forward(NDArray inputIds) {
+        switch (searchName) {
+            case "greedy":
+                return greedySearch(inputIds);
+            case "beam":
+                return beamSearch(inputIds);
+            case "contrastive":
+                return contrastiveSearch(inputIds);
+            default:
+                return greedySearch(inputIds);
```
zachgk, line 491: I would consider breaking up `LMSearch` into an abstract `LMSearch` class and four specific search blocks that extend it. You can still share helper code like `prepareAttentionMaskOffset` by putting it in the parent `LMSearch`.

Otherwise, you should probably throw an Exception for the default behavior rather than just use greedySearch. When in doubt when something is probably going wrong, it is better to make sure the user is aware and can fix it rather than risk doing the wrong thing to compensate for the user's mistake

api/src/main/java/ai/djl/translate/LMSearch.java
```diff
@@ -0,0 +1,617 @@
...
+                                .repeat(1, numBeam);
+
+                // [batch, beam, heads, seq_past, kvFeature]
+                Function<NDArray, NDArray> fn = ndarray -> ndarray.expandDims(1).repeat(1, numBeam);
+                NDList pastKeyValues =
+                        new NDList(
+                                modelOutput.pastKeyValuesList.stream()
+                                        .map(fn)
+                                        .collect(Collectors.toList()));
+                // [batch, beam, seq_past]
```
zachgk, line 138: These shape comments are a great addition! It makes it much easier to track the NDArray operations

zachgk (changes requested):

api/src/main/java/ai/djl/engine/Engine.java
```diff
@@ -302,6 +307,11 @@ public SymbolBlock newSymbolBlock(NDManager manager) {
      */
     public abstract NDManager newBaseManager(Device device);
 
+    public LMBlock newLMAdapter(String languageModel, GPTConfig gptConfig, Block[] blocks)
```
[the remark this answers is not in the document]
zachgk: It doesn't need to be engine agnostic. It should be somewhere in `pytorch-model-zoo`. In the `PtModelZoo`, register the GPT2PtLMBlock model. Within the metadata for that model, you can specify the block factory to use. Maybe a nested class like `GPT2PtLMBlock.Factory`. And, that block factory (probably located in `pytorch-model-zoo`) will call the `GPT2PtLMBlock(...)` constructor. Hopefully that makes sense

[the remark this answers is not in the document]
zachgk: It could be kept in the same directory, but would probably be a bit better in the `pytorch-model-zoo` package rather than `pytorch-engines` package. Within the `pytorch-model-zoo`, it still depends with implementation on the `pytorch-engines` so you can still access all of the internal pytorch details

KexinFeng:
Features have been merged in https://github.com/deepjavalibrary/djl/pull/2637

PR #2637, opened 2023-06-06
[api] implements text-generation search algorithm

KexinFeng:
This PR succeeds PR #2547 #2509 #2557, #2572 which contains the benchmark outputs of the searching results. Model tracing is in #2509 's PR description.

This PR contains only the features of LMSearch. 

`djl/examples/src/main/java/ai/djl/examples/inference/GPTInference.java` contains the front_end design.


examples/src/main/java/ai/djl/examples/inference/GPTInference.java
```diff
@@ -0,0 +1,216 @@
...
+import ai.djl.translate.TranslatorContext;
+import ai.djl.util.Pair;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.List;
+
+public final class GPTInference {
```
frankfliu, line 44:
```suggestion
public final class GptInference {
```

See: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

api/src/main/java/ai/djl/modality/nlp/generate/SearchConfig.java
```diff
@@ -0,0 +1,122 @@
...
+    private int k;
+    private float alpha;
+    private int beam;
+    private int maxSeqLength;
+    private long padTokenId;
+    private long eosTokenId;
+    private boolean suffixPadding;
+
+    /** Constructs a new ContrastiveSearchConfig object with default values. */
+    public SearchConfig() {
```
jawaff, line 26: Any plans to support different configurations since not all of the text generation models are the same? I'm personally more interested in T5 than GPT2. T5 in particular is a different beast with both a decoder and encoder in contrast to GPT2's decoder-only approach. T5 also supports over a hundred special tokens. There's 100 "extra" tokens that can be used for a variety of things including fill masks and potentially representing special words/instructions in the generated output.

https://huggingface.co/transformers/v3.0.2/model_doc/t5.html#t5config

There probably needs to be different configurations and generation classes for each of the family of models out there. If we hardcode everything to GPT2, then there's going to be breaking changes in the future. I'd suggest adding support for two different models starting out and coming up with a solution for adding support for others in the future..

jawaff:
I'm just leaving my 2 cents since I'm interested in your work. It's great to see it get added to DJL. My only real fear is that there might need to be a lot of refactoring to get support for other models. Flan-T5 is one of the most powerful open source models (that supports commercial use) we have available and it has a variety of sizes available. I'd be most interested in seeing it be supported.

I'm not super familiar with GPT2 aside from it being a decoder-only model. There's a chance that T5 and GPT2 share some similarities in the decoder aspect, but T5 has an initial encoder pass on the initial inputs. The hidden state of the encoded inputs are then used for each pass of the decoder alongside the ids that have been currently selected for generation.

jawaff:
That's all I've got to add, good work. I just want to see this turn into a bigger feature beyond what you're working on.

frankfliu:
> That's all I've got to add, good work. I just want to see this turn into a bigger feature beyond what you're working on.

We are planning to add T5 model. This is just a starting point to add textgeneration support.

KexinFeng:
@jawaff Thanks for pointing out the encoder-decoder model T5 to us and reminding us of the possible refactoring. 

I think to implement encoder-decoder model, the major edition will be in the search algorithms, where we will need an if (encoderDecoder is true) block, which computes the encoding). The rest part of the code will basically be shared. This structure is seen in huggingface transformer.

frankfliu (approved):

PR #2640, opened 2023-06-07
[api] Add getVenvDir utility function

xyang16:
## Description ##

Add getVenvDir utility function

### DIFF
--- a/api/src/main/java/ai/djl/util/Utils.java
+++ b/api/src/main/java/ai/djl/util/Utils.java
@@ -382,6 +382,23 @@ public final class Utils {
     }
 
     /**
+     * Utility function to get python virtual env directory.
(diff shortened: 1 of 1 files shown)


api/src/main/java/ai/djl/util/Utils.java
```diff
@@ -381,6 +381,23 @@ public static Path getNestedModelDir(Path modelDir) {
         return modelDir.toAbsolutePath();
     }
 
+    /**
+     * Utility function to get python virtual env directory.
+     *
+     * @return DJL venv directory
+     */
+    public static Path getVenvDir() {
```
frankfliu, line 389: This function doesn't seems to belong to DJL

PR #2659, opened 2023-06-17
[xgboost] Adds xgboost aarch64 support

frankfliu:
## Description ##

### DIFF
--- a/.github/workflows/native_s3_xgboost.yml
+++ b/.github/workflows/native_s3_xgboost.yml
@@ -1,0 +1,79 @@
+name: Native S3 XGBoost
+
+on:
+  workflow_dispatch:
+    xgb_version:
+      description: 'xgboost version'
+      required: false
+
+jobs:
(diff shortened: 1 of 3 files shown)


lanking520 (approved):

lanking520:
you can use this: https://github.com/dmlc/xgboost/blob/d8beb517edc8acf1622c13c3086cc8fa5e15a663/jvm-packages/create_jni.py#L143

lanking520:
It brings you some level of maintainability

.github/workflows/native_s3_xgboost.yml
```diff
@@ -0,0 +1,74 @@
+name: Native S3 XGBoost
+
+on:
+  workflow_dispatch:
```
lanking520, line 4: Let's allow customer specify version here

lanking520: Nvm, we go with DJL

PR #2677, opened 2023-06-25
[M1 Macs] add warning for model criteria inexplicably not working when api is loaded as a jar in a new project.

juliangamble:
## Description ##
When you load models on an M1 Mac you can get the error 
 fails with input/output not matching exception. 
This can arise when you're using dlj `api.jar` in a new project as a dependency - but you didn't include `pytorch-model-zoo`

The purpose of this PR is to give a flag to the user to say "hey - just in case you didn't know this" - you're probably looking in the wrong place to solve this problem. Look at project dependencies, not input parameters to `Criteria`. 

### DIFF
--- a/api/src/main/java/ai/djl/repository/zoo/Criteria.java
+++ b/api/src/main/java/ai/djl/repository/zoo/Criteria.java
@@ -181,6 +181,9 @@ public class Criteria<I, O> {
                 }
             }
         }
+        logger.debug(
+                "If you're on an MacOS arch64 (Apple Silicon) you may be getting here "+
+                        "because you do not have the pytorch-model-zoo dependency installed.");
         throw new ModelNotFoundException(
                 "No model with the specified URI or the matching Input/Output type is found.",
                 lastException);


api/src/main/java/ai/djl/repository/zoo/Criteria.java
```diff
@@ -181,6 +181,9 @@ public ZooModel<I, O> loadModel()
                 }
             }
         }
+        logger.debug(
```
frankfliu, line 184: API is an abstract layer which doesn't have any knowledge of underlying engine, platform and model zoo.

This is not the right place to print out this message.

I understand we need to improve model loading message to help troubleshoot model url and model loading related issue. Currently we have to turn on debug log to understand what happened during model loading, this is not a good user experience. We aware of this limitations and looking for an improvement.

I created a PR to address partial of this issue: https://github.com/deepjavalibrary/djl/pull/2678

juliangamble: Pull request looks good. Thankyou.

juliangamble:
Appreciate the feedback and replacement pull request.

PR #2683, opened 2023-06-29
[doco] Model loading - file naming conventions

juliangamble:
## Description ##

Updating doco for file naming conventions when loading models generated from djl.

### DIFF
--- a/docs/load_model.md
+++ b/docs/load_model.md
@@ -106,6 +106,17 @@ You can also use the URL query string to tell DJL how to load model:
 file:///var/models/resnet.zip?model_name=saved_model/resnet-18
 ```
 
+Note that you **can't** load the modelname of a `.params` model file by specifying the full file including file extension. You have to specify the name of the model *without* the file extension eg
+
+```java
+.optModelName("myModelName")
+```
+not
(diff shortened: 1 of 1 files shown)


docs/load_model.md
```diff
@@ -106,6 +106,17 @@ You can also use the URL query string to tell DJL how to load model:
 file:///var/models/resnet.zip?model_name=saved_model/resnet-18
 ```
 
+Note that you **can't** load the modelname of a `.params` model file by specifying the full file including file extension. You have to specify the name of the model *without* the file extension eg
```
frankfliu, line 109: You understanding is not correct:
1. First of all, we support different engines, and the for different engine, the model name extension are different.
2. To support different file name extensions, we recommend to use file name without file extension (file name prefix)
3. You can use full fileName, but the file name should match what engine expected, for example, for a pytorch engine, we are looking for "myModelName.pt" file
4. For a ".params" file, it's DJL trained format, they need use a Block to load these ".params".

juliangamble: Thanks, that's helpful. I'll close this and rethink.

PR #2706, opened 2023-07-13
pytorch support inference on separate cuda stream

jiyuanq:
## Description ##

This PR is an attempt to support running inference on separate cuda streams for pytorch engine. By doing this, we can maximize GPU utilization when running concurrent inference requests on GPU.

Also added a boolean flag `inferenceSeparateCudaStream` controlled through system properties `"ai.djl.pytorch.inference_separate_cuda_stream"` to determine whether this new feature is enabled.

I considered exposing the full cuda stream related pytorch api through JNI but in the end decided to only expose a high level boolean flag, mainly because:

- it's much easier to expose only this boolean flag
- there's no other use case that requires the full cuda stream api

### DIFF
--- a/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/IValueUtils.java
+++ b/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/jni/IValueUtils.java
@@ -35,6 +35,8 @@ public final class IValueUtils {
     private static final Pattern PATTERN_LIST = Pattern.compile("\\w+\\[]");
     private static final Pattern PATTERN_TUPLE = Pattern.compile("\\w+\\(\\)");
     private static final Pattern PATTERN_TUPLE_OF_TUPLE = Pattern.compile("\\w+(\\([\\d,]+\\))");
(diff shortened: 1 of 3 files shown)


jiyuanq:
@frankfliu @zachgk any chance you can take a look and give some early feedback? I checked the failed build looks like it's a coding convention check failure (not sure if I missed any other error message), which I'll fix once there's consensus on the PR itself

frankfliu:
@jiyuanq 
Thanks for your contribution. What you need to do is just run the following command:

```
./gradlew formatJava
```

And then check in updated files.

See: https://github.com/deepjavalibrary/djl/blob/master/docs/development/development_guideline.md#coding-conventions

jiyuanq:
> @jiyuanq Thanks for your contribution. What you need to do is just run the following command:
> 
> ```
> ./gradlew formatJava
> ```
> 
> And then check in updated files.
> 
> See: https://github.com/deepjavalibrary/djl/blob/master/docs/development/development_guideline.md#coding-conventions

thank you! just updated

frankfliu (approved):

PR #2732, opened 2023-08-01
Setup custom ft build for Llama support

rohithkrn:
## Description ##

- Clones FT backend from forked repo - https://github.com/rohithkrn/fastertransformer_backend.git which adds llama support. Backend repo [points](https://github.com/rohithkrn/fastertransformer_backend/blob/llama_support/CMakeLists.txt#L113) to FT repo with model implementation which taken from void-main's PR to FT repo.
- Added `is_llama_build` flag to isolate from existing setup so that it will be easy to switch. This also pushes binaries to a new s3 path (`fastertransformer/llama`) which probably needs creation.

### DIFF
--- a/.github/workflows/build_triton_and_ft.yml
+++ b/.github/workflows/build_triton_and_ft.yml
@@ -11,6 +11,11 @@ on:
         description: 'fastertransformer branch/tag version'
         required: true
         default: 'main'
+      is_llama_build:
+        description: 'whether to build custom llama source'
+        required: false
+        type: boolean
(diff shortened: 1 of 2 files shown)


frankfliu (approved):

rohithkrn:
test runs of the workflow
https://github.com/deepjavalibrary/djl/actions/runs/5741513102/job/15561745671
https://github.com/deepjavalibrary/djl/actions/runs/5732753126/job/15536271103

PR #2793, opened 2023-09-29
Updates LightGBM to 3.3.510

zachgk:
## Description ##


lanking520 (approved):

gradle.properties
```diff
@@ -23,7 +23,7 @@ sentencepiece_version=0.1.97
 tokenizers_version=0.13.3
 fasttext_version=0.9.2
 xgboost_version=1.7.5
-lightgbm_version=3.2.110
+lightgbm_version=3.3.510
```
frankfliu, line 26: LightGBM 3.3 doesn't support AL2

PR #2799, opened 2023-10-03
Prevent model server crash if data is missing in metadata file during callHome

maaquib:
## Description ##


api/src/main/java/ai/djl/util/Ec2Utils.java
```diff
@@ -178,7 +178,7 @@ private static String readEndpointMetadata(String key) {
         Path path = Paths.get(ENDPOINT_METADATA_FILE);
         try (Reader reader = Files.newBufferedReader(path)) {
             JsonObject json = JsonUtils.GSON.fromJson(reader, JsonObject.class);
-            return json.get(key).getAsString();
+            return (json.get(key) != null) ? json.get(key).getAsString() : null;
```
frankfliu, line 181: In which case json is empty?

PR #2803, opened 2023-10-11
Expose numpy encode and decode

zachgk:
This exposes the numpy encoding on the NDArray class and the decoding on the NDManager class. It also refactors the NDSerializer to use a ByteBufferBackedInputStream (added as a utility) to combine the encoding of input streams and byte buffers

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/NDManager.java
+++ b/api/src/main/java/ai/djl/ndarray/NDManager.java
@@ -720,6 +720,27 @@ public interface NDManager extends AutoCloseable {
     }
 
     /**
+     * Decodes a numpy .npy {@link NDArray} through byte array.
+     *
+     * @param bytes byte array to load from
+     * @return {@link NDArray}
+     */
+    default NDArray decodeNumpy(byte[] bytes) {
+        return NDSerializer.decodeNumpy(this, ByteBuffer.wrap(bytes));
+    }
+
+    /**
+     * Decodes a numpy .npy {@link NDArray} through {@link DataInputStream}.
+     *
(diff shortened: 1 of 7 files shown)


frankfliu (commented):
You don't need this change. the decode function will automatically detect the file format.

api/src/main/java/ai/djl/ndarray/NDManager.java
```diff
@@ -719,6 +719,27 @@ default NDArray decode(InputStream is) throws IOException {
         return NDSerializer.decode(this, is);
     }
 
+    /**
+     * Decodes a numpy .npy {@link NDArray} through byte array.
+     *
+     * @param bytes byte array to load from
+     * @return {@link NDArray}
+     */
+    default NDArray decodeNumpy(byte[] bytes) {
```
frankfliu, line 728: 1. We should push auto detection logic from NDList to this level and not expose this API.
2. If we fix the bug in NDList, user can always use `NDList.decode()`

frankfliu: I tried to add auto detection in `NDSerializer.decode()`, but I realized that the performance won't be good.
1. I added ByteBuffer version due the performance issue. It's about 2-3 times faster than InputStream version. I didn't add ByteBuffer version for numpy.
2. I think we should keep it simple and only dealing with numpy and other file format in NDList.
3. If there is need to save .npy file format, we can add `NDList.Encoding.NUMPY` format( .npz file is just a zip file, user can extract .npy from .npz file)

PR #2806, opened 2023-10-12
[api] Added Early stopping configuration (#38)

### LINKED ISSUE #38: Early stopping configuration
## Description
Early stopping configuration: Specifies the various configuration options for running training with early stopping.
- early stopping model saver - only use last best model: How model will be saved (to disk, to memory, etc)
- Termination conditions:
      1. Iteration termination conditions: how many epoch till termination.
      2. score improvement termination condition - terminate training if best model score does not improve for N epochs
      3. best expected score - terminate training once we achieved an expected score.
      4. termination condition after certain time - terminate training after certain time
      5. other termination conditions, if they are logical

### Will this change the current api? How?
We can configure when model training will stop, when one of condition above is met.
Training should be implemented as listener, early stop configuration will listen for any conditions above and terminate training.

### Who will benefit from this feature?
Everybody, we can easily configure when learning will end.

## References
Reference implementation:
https://github.com/eclipse/deeplearning4j/blob/b5f0ec072f3fd0da566e32f82c0e43ca36553f39/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/earlystopping/EarlyStoppingConfiguration.java
There are other implementation in different NN framework.

jagodevreede:
## Description ##

This PR implements a listener for early stopping as described in #38. And takes the code @gforman44  and implements that in a `TrainingListener`

resolves #38

### DIFF
--- a/api/src/main/java/ai/djl/training/listener/EarlyStoppingListener.java
+++ b/api/src/main/java/ai/djl/training/listener/EarlyStoppingListener.java
@@ -1,0 +1,175 @@
+package ai.djl.training.listener;
+
+import ai.djl.training.Trainer;
+
+/**
+ * Listener that allows the training to be stopped early if the validation loss is not improving, or if time has expired.
+ * <br/>
+ * Usage:
+ * Add this listener to the training config, and add it as the last one.
+ * <pre>
+ *  new DefaultTrainingConfig(...)
+ *        .addTrainingListeners(new EarlyStoppingListener()
+ *                .setEpochPatience(1)
+ *                .setEarlyStopPctImprovement(1)
+ *                .setMaxMinutes(60)
+ *                .setMinEpochs(1)
+ *        );
+ * </pre>
+ * Then surround the fit with a try catch that catches the {@link EarlyStoppingListener.EarlyStoppedException}.
+ * <br/>
+ * Example:
+ * <pre>
+ * try {
+ *   EasyTrain.fit(trainer, 5, trainDataset, testDataset);
+ * } catch (EarlyStoppingListener.EarlyStoppedException e) {
+ *   // handle early stopping
+ *   log.info("Stopped early at epoch {} because: {}", e.getEpoch(), e.getMessage());
+ * }
+ * </pre>
+ * <br/>
+ * Note: Ensure that Metrics are set on the trainer.
+ */
+public class EarlyStoppingListener implements TrainingListener {
+    private final double objectiveSuccess;
+
+    /**
+     * after minimum # epochs, consider stopping if:
+     */
+    private int minEpochs;
+    /**
+     * too much time elapsed
+     */
+    private int maxMinutes;
+    /**
+     * consider early stopping if not x% improvement
+     */
+    private double earlyStopPctImprovement;
+    /**
+     * stop if insufficient improvement for x epochs in a row
+     */
+    private int epochPatience;
+
+    private long startTimeMills;
+    private double prevLoss;
+    private int numberOfEpochsWithoutImprovements;
+
+    public EarlyStoppingListener() {
+        this.objectiveSuccess = 0;
+        this.minEpochs = 0;
+        this.maxMinutes = Integer.MAX_VALUE;
+        this.earlyStopPctImprovement = 0;
+        this.epochPatience = 0;
+    }
+
+    public EarlyStoppingListener(double objectiveSuccess, int minEpochs, int maxMinutes, double earlyStopPctImprovement, int earlyStopPatience) {
+        this.objectiveSuccess = objectiveSuccess;
+        this.minEpochs = minEpochs;
+        this.maxMinutes = maxMinutes;
+        this.earlyStopPctImprovement = earlyStopPctImprovement;
+        this.epochPatience = earlyStopPatience;
+    }
+
+    @Override
+    public void onEpoch(Trainer trainer) {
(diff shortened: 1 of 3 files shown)


integration/src/main/java/ai/djl/integration/tests/training/listener/EarlyStoppingListenerTest.java
```diff
@@ -0,0 +1,116 @@
...
+public class EarlyStoppingListenerTest {
+
+    private final Optimizer sgd = Optimizer.sgd().setLearningRateTracker(Tracker.fixed(0.1f)).build();
+
+    private Mnist testMnistDataset;
+    private Mnist trainMnistDataset;
+
+    @BeforeTest
+    public void setUp() throws IOException, TranslateException {
+        testMnistDataset = Mnist.builder()
```
zachgk, line 36: You can set the `limit` on the dataset to make it act like it only has a smaller size. This will let these tests be sped up any further. We try to keep most of our unit tests as quick as possible because that makes our build and CI run faster and we have to spend less time waiting for it

api/src/main/java/ai/djl/training/listener/EarlyStoppingListener.java
```diff
@@ -0,0 +1,175 @@
...
+        this.prevLoss = Double.NaN;
+        this.numberOfEpochsWithoutImprovements = 0;
+    }
+
+    @Override
+    public void onTrainingEnd(Trainer trainer) {
+        // do nothing
+    }
+
+    public EarlyStoppingListener setMinEpochs(int minEpochs) {
```
zachgk, line 140: Instead of adding setters on the listener, you may want to make a builder instead. I don't see a reason that users would be modifying this listener while training. Instead, having a builder will make it easier to work with the number of possible arguments where some of them are defaults and others are specified. 

We typically have our builders called by `EarlyStoppingLister.builder().setters...().build()` with arguments as either `setXXX()` or `optXXX()` if required or optional

api/src/main/java/ai/djl/training/listener/EarlyStoppingListener.java
```diff
@@ -0,0 +1,175 @@
...
+                    }
+                }
+            }
+        }
+        if (Double.isFinite(loss)) {
+            prevLoss = loss;
+        }
+    }
+
+    private static double getLoss(Trainer trainer) {
```
zachgk, line 106: Maybe have `TrainingResult result - trainer.getTrainingResult()` in `onEpoch`. Then, you can make this `getLoss(TrainingResult result)`

api/src/main/java/ai/djl/training/listener/EarlyStoppingListener.java
```diff
@@ -0,0 +1,175 @@
...
+    private final double objectiveSuccess;
+
+    /**
+     * after minimum # epochs, consider stopping if:
+     */
+    private int minEpochs;
+    /**
+     * too much time elapsed
+     */
+    private int maxMinutes;
```
zachgk, line 43: Maybe change this to `maxMillis`. This will let you do the math a bit easier for each epoch. You can still add the helper `setMaxMinutes()` in addition to `setMaxMillis()` in the builder so you can still call it just as easily.

The other benefit for testing is that you can now add a test with it. You can set it to stop after only one ms and then train until timeout.

zachgk (changes requested):
Thank you for contributing this @jagodevreede! I'm sure many training users will appreciate it. Your PR looks pretty good, but I left a few comments.

I'll also run the CI. It will probably complain from our code formatter and checkstyle. See http://docs.djl.ai/docs/development/development_guideline.html#coding-conventions for some explanations and how to check it locally as well

integration/src/main/java/ai/djl/integration/tests/training/listener/EarlyStoppingListenerTest.java
jagodevreede: Set the limit, and reduced the run time to 700ms (was 6 seconds) and now has a extra test as well

api/src/main/java/ai/djl/training/listener/EarlyStoppingListener.java
jagodevreede: Introduced a build, with all optional parameters, as you can pick whatever you want.

api/src/main/java/ai/djl/training/listener/EarlyStoppingListener.java
jagodevreede: `getLoss` now taks the TrainingResult as a param

api/src/main/java/ai/djl/training/listener/EarlyStoppingListener.java
jagodevreede: Introduced a test, and made it millis. I also made a convenient method that takes a duration in the builder that converts it for you. I dislike passing big numbers as I usually train for hours.

jagodevreede:
Local build is happy now, let's see if CI agrees with that. Also fixed your comments.

jagodevreede:
I was able to reproduce the failing ci build locally with JDK11. Fixed it in commit [402df41](https://github.com/deepjavalibrary/djl/pull/2806/commits/402df41f850d6e472a14592c47b649e7924b27f2)

zachgk (approved):

PR #2819, opened 2023-10-24
Creates MultiDevice

zachgk:
This creates an abstraction for combining devices into a single device. The main use case for now is in DJL Serving TP_parallel. It will allow us to create a WorkerGroup and a PyPredictor for a set of devices and then track the usage of devices properly. It could also be used later for multi-gpu training or other multi-device cases.

### DIFF
--- a/api/src/main/java/ai/djl/Device.java
+++ b/api/src/main/java/ai/djl/Device.java
@@ -14,11 +14,16 @@ package ai.djl;
 
 import ai.djl.engine.Engine;
 
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * The {@code Device} class provides the specified assignment for CPU/GPU processing on the {@code
@@ -30,7 +35,7 @@ import java.util.regex.Pattern;
  * @see <a href="https://d2l.djl.ai/chapter_deep-learning-computation/use-gpu.html">The D2L chapter
  *     on GPU devices</a>
  */
-public final class Device {
+public class Device {
 
     private static final Map<String, Device> CACHE = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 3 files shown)


sindhuvahinis (approved):

api/src/main/java/ai/djl/Device.java
```diff
@@ -101,6 +106,13 @@ public static Device fromName(String deviceName, Engine engine) {
             return engine.defaultDevice();
         }
 
+        if (deviceName.contains("+")) {
```
frankfliu, line 109: We need think of the following use cases:
1. A specific device id (existing Device implementation)
2. A continuous range of device: GPU[1-3]
3. Arbitrary device list: GPU1;GPU3
4. Number of device at any free device id exclusively: GPU{2}
5. All available devices exclusively: GPU+
6. All devices sharable: GPU*

zachgk: We actually have two device naming systems. One is the base system used in DJL `Device.fromName()`. The other is the system used in Serving `getLoadOnDevices()`. For example, `*` exists in Serving but not in DJL. The main idea seems to be that all the ones in DJL are absolute descriptions of a device and the ones in serving also contain relative ones. In that case and with your list: DJL would contain 1, 2, 3, 5 and Serving would contain 4, 6.

First, I want to talk about the structure of Device. Here, I changed it to represent anything "Device-like", either real, virtual, a combination of devices, or parts of devices. Device is now open for interpretation. I think this works very well with respect to how it opens possibilities throughout all of the API, even if many would not be supported for now. It helps a lot with multi-device usage, tensor parallel, device sharing, and distributed training. I would support having a clearer recognition of physical devices, though. Would it help to either add a function `device.isPhysicalDevice()` or a class `PhysicalDevice extends Device`?

Also for your list, you need to deal with both levels of lists of device considering tensor parallel. That is, you need something equivalent to "gpu0+gpu1;gpu2+gpu3". Which is, two workers of TP 2. I could also see `{gpu0;gpu1};{gpu2;gpu3}`. We also don't want to use `,` because it is used elsewhere. Then, would we want to have ranges like `gpu[0-3/2]` which would allow for TP? Also, with the current system we could still use a `+` without anything else even with the current system similarly to how we are using `*`. Both of these infer the device.

PR #2847, opened 2023-11-10
only build triton binaries

lanking520:
## Description ##

Only build triton binary and remove FT


rohithkrn:
we can also remove `build_ft_deps.sh` file

sindhuvahinis (approved):

PR #2854, opened 2023-11-15
[docs] Move jupyter notebooks to DJL Demo

zachgk:
This moves the notebooks to the demo repositories. It makes it more clearly follow the trend where the things in the main repo are master build while those in demos track nightly snapshots or releases.

Combined with https://github.com/deepjavalibrary/djl-demo/pull/390


lanking520:
we also have a job that just run those notebook and append output result, does that still work with this change?

zachgk:
> we also have a job that just run those notebook and append output result, does that still work with this change?

Are you referring to the docs action? I updated that one (and it won't pass until both changes are merged) so this PR is not expected to pass CI

frankfliu (approved):

PR #2929, opened 2024-01-09
[pytorch] Allows to load libstdc++.so.6 form different location

frankfliu:
## Description ##

### DIFF
--- a/engines/pytorch/pytorch-engine/src/test/java/ai/djl/pytorch/integration/ALibUtilsTest.java
+++ b/engines/pytorch/pytorch-engine/src/test/java/ai/djl/pytorch/integration/ALibUtilsTest.java
@@ -18,17 +18,21 @@ import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
-public class LibUtilsTest {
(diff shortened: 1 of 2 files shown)


engines/pytorch/pytorch-engine/src/test/java/ai/djl/pytorch/integration/ALibUtilsTest.java
```diff
@@ -18,17 +18,21 @@
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
-public class LibUtilsTest {
+// Ensure this test run first
+public class ALibUtilsTest {
```
siddvenk, line 22: Why the change from `LibUtils` to `ALibUtils`?

frankfliu: The testng use alphabetical order to run the unittest classes. The engine is only initialized once in the JVM. We want make sure this test executed first in all tests.

rahulsmit (approved):
Thanks this will provide added flexibility

siddvenk (approved):

PR #2935, opened 2024-01-11
Suppress serial warning for JDK21

zachgk:
In JDK21, it now throws the serial warning for including potentially unserializable instance variables. This includes the standard Java data structures like List, Set, and Map. This changes the JDK21 support from https://github.com/deepjavalibrary/djl/pull/2903 to suppress the warning rather than no longer serializing the variables.

### DIFF
--- a/api/src/main/java/ai/djl/modality/cv/output/CategoryMask.java
+++ b/api/src/main/java/ai/djl/modality/cv/output/CategoryMask.java
@@ -43,7 +43,9 @@ public class CategoryMask implements JsonSerializable {
                     .registerTypeAdapter(CategoryMask.class, new SegmentationSerializer())
                     .create();
 
-    private transient List<String> classes;
+    @SuppressWarnings("serial")
+    private List<String> classes;
+
     private int[][] mask;
 
     /**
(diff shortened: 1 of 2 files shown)


api/src/main/java/ai/djl/modality/cv/output/CategoryMask.java
```diff
@@ -43,7 +43,9 @@ public class CategoryMask implements JsonSerializable {
                     .registerTypeAdapter(CategoryMask.class, new SegmentationSerializer())
                     .create();
 
-    private transient List<String> classes;
+    @SuppressWarnings("serial")
+    private List<String> classes;
```
frankfliu, line 47: We still need transient to ensure they are not get serialized by Gson

zachgk: Is this supposed to be transient? I am reverting the ones that were marked as transient in #2903 back to non-transient

frankfliu: Yes, they are. Otherwise, 1000 classes will be serialized as output.

frankfliu (approved):

PR #2957, opened 2024-01-22
[tokenizer] Return if exceed max token length

frankfliu:
## Description ##


zachgk (approved):

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/Encoding.java
```diff
@@ -127,6 +130,15 @@ public CharSpan[] getCharTokenSpans() {
         return charTokenSpans;
     }
 
+    /**
+     * Returns if tokens exceed max length.
+     *
+     * @return {@code true} if tokens exceed max length
+     */
+    public boolean exceedMaxLength() {
```
ylwu-amzn, line 138: Will be even better if we can get how many tokens exceeded.

PR #2971, opened 2024-01-30
Switch to cheaper M1

zachgk:
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

### DIFF
--- a/.github/workflows/native_jni_s3_pytorch.yml
+++ b/.github/workflows/native_jni_s3_pytorch.yml
@@ -206,7 +206,7 @@ jobs:
 
   build-pytorch-jni-arm64-macos:
     if: github.repository == 'deepjavalibrary/djl'
-    runs-on: macos-latest-xlarge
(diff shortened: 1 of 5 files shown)


.github/workflows/native_jni_s3_pytorch.yml
```diff
@@ -206,7 +206,7 @@ jobs:
 
   build-pytorch-jni-arm64-macos:
     if: github.repository == 'deepjavalibrary/djl'
-    runs-on: macos-latest-xlarge
+    runs-on: macos-14 #M1
```
frankfliu, line 209: We should use lower version os macos

frankfliu: If we compile the JNI with macos-14, it can only run on macoss-14+
We can use macos-13 to run test, but compile JNI has to use lowest version.

PR #2986, opened 2024-02-10
fix #2968

### LINKED ISSUE #2968: Hundreds wrong detection on yolov8
Using code writed for yolov5, after some adjustment for yolov8, when I try to use torchscript model generated I got ~200 detection with very high score (10000+), and they are mostly focused in left top corner. My model probably is created correctly using yolov8s/yolov8n model (Ultralytics tool) and `yolo predict` work great on that model. Then I exported to torchscript format (`yolo export model=... format=torchscript` like v5 version).

1-Figure1-1

I updated djl to 0.26 and changed translator to `YoloV8Translator`.
My yolo usage looks like:
```
        Image picture = ImageFactory.getInstance().fromFile(image.toPath());
        
        Pipeline pipeline = new Pipeline();
        pipeline.add(new Resize(640));
        pipeline.add(new ToTensor());

        List<String> synset = new ArrayList<>(4);
        synset.add(DETECTION_LABEL_0);
        synset.add(DETECTION_LABEL_1);
        synset.add(DETECTION_LABEL_2);
        synset.add(DETECTION_LABEL_3);

        YoloV8Translator translator = YoloV8Translator
                .builder()
                .setPipeline(pipeline)
                .optSynset(synset)
                .optThreshold(0.7f)
                .build();

        Criteria criteria = Criteria.builder()
                .setTypes(Image.class, DetectedObjects.class)
                .optDevice(Device.cpu())
                .optModelPath(Paths.get(modelLocation))
                .optTranslator(translator)
                .optEngine("PyTorch")
                .build();

        ZooModel<Image, DetectedObjects> model = criteria.loadModel();
        Predictor<Image, DetectedObjects> predictor = model.newPredictor();
        predictor.predict(picture);
```

Same code with YoloV5Translator, djl 0.19.0 and model built on yolov5 works fine.
What I forgot, or what I do wrong? Can you help me migrate to v8?

SidneyLann:
fix #2968

### DIFF
--- a/api/src/main/java/ai/djl/modality/cv/translator/YoloV8Translator.java
+++ b/api/src/main/java/ai/djl/modality/cv/translator/YoloV8Translator.java
@@ -64,6 +64,11 @@ public class YoloV8Translator extends YoloV5Translator {
         float[] buf = reshapedResult.toFloatArray();
         int numberRows = Math.toIntExact(shape.get(0));
         int nClasses = Math.toIntExact(shape.get(1));
+        int padding = nClasses - classes.size();
+        if (padding != 0 && padding != 4) {
+            throw new IllegalStateException(
+                    "Expected classes: " + (nClasses - 4) + ", got " + classes.size());
+        }
 
         ArrayList<IntermediateResult> intermediateResults = new ArrayList<>();
         // reverse order search in heap; searches through #maxBoxes for optimization when set
@@ -75,9 +80,10 @@ public class YoloV8Translator extends YoloV5Translator {
                 float classProb = buf[index + c];
                 if (classProb > maxClassProb) {
                     maxClassProb = classProb;
-                    maxIndex = c - 4;
+                    maxIndex = c;
                 }
             }
+            maxIndex -= padding;
 
             if (maxClassProb > threshold) {
                 float xPos = buf[index]; // center x
--- a/api/src/main/java/ai/djl/modality/cv/translator/YoloV8Translator.java
+++ b/api/src/main/java/ai/djl/modality/cv/translator/YoloV8Translator.java
@@ -75,7 +75,7 @@ public class YoloV8Translator extends YoloV5Translator {
                 float classProb = buf[index + c];
                 if (classProb > maxClassProb) {
                     maxClassProb = classProb;
-                    maxIndex = c;
+                    maxIndex = c - 4;
                 }
             }
 


api/src/main/java/ai/djl/modality/cv/translator/YoloV8Translator.java
```diff
@@ -75,7 +75,7 @@ protected DetectedObjects processFromBoxOutput(NDList list) {
                 float classProb = buf[index + c];
                 if (classProb > maxClassProb) {
                     maxClassProb = classProb;
-                    maxIndex = c;
+                    maxIndex = c - 4;
```
frankfliu, line 78: maxIndex = c is corret.

nClasses should be the same as classes.length. We might add a check to ensure we provide correct error message.

frankfliu (changes requested):

api/src/main/java/ai/djl/modality/cv/translator/YoloV8Translator.java
SidneyLann: No. maxIndex should start with 0, but not 4!

SidneyLann: The labels 0, 1, 2, 3 should never be detected for the bug.

SidneyLann: For the existing logic, the max value of maxIndex is classes.length + 3, but not classes.length -1

SidneyLann: Or, for (int c = 4; c < nClasses; c++) should change to for (int c = 0; c < nClasses; c++)?

SidneyLann: maybe this is correct:
            for (int c = 0; c < nClasses; c++) {
                float classProb = buf[index + 4 + c];
                if (classProb > maxClassProb) {
                    maxClassProb = classProb;
                    maxIndex = c;
                }
            }

frankfliu: I see where is the problem. Original COCO dataset has 80 classes, but in synset.txt contains 84 rows and first 4 are dummy:

0 = "# Classes for coco dataset on which yelov8 is trained"
1 = "# source config https://github.com/ultralytics/ultralytics/blob/main/ultralytics/cfg/datasets/coco.yaml."
2 = "# COCO dataset website: https://cocodataset.org/#home"
3 = "# Ultralytics Coco doc page: https://docs.ultralytics.com/datasets/detect/coco/"

So if you use YoloV8Translator, you need pad 4 dummy labels.

frankfliu (approved):

api/src/main/java/ai/djl/modality/cv/translator/YoloV8Translator.java
SidneyLann: if (padding != 0 && padding != 4) { //means padding ==0 or padding ==4 is invalid, this is not true. the check should be removed and change for (int c = 4; c < nClasses; c++) { to for (int c = padding; c < nClasses; c++) {?

SidneyLann: if (padding != 0 && padding != 4) { //means padding ==0 or padding ==4 is invalid, this is not true. the check should be removed and change for (int c = 4; c < nClasses; c++) { to for (int c = padding; c < nClasses; c++) {?

SidneyLann:
The merging logic is not correct

frankfliu:
The output shape is (8400, 84) for COCO dataset, we should always ignore first 4 entries in the output.
Ideally, we should not padding classes with dummy classes, the originally implementation is a hack. Keep the 4 padding behavior is purely for backward compatibility. We should not allow other paddings.

SidneyLann:
But 0 padding or 4 padding is not work now

frankfliu:
0 or 4 should work, it throw exception when not 0 **and** not 4
```
if (padding != 0 && padding != 4) {
```

SidneyLann:
If padding is 0, the min label is 4, it should be 0

frankfliu:
I think the variable name `padding` is misleading. It should `diff`,

if `diff` == 0, it mean the classes has already padded with dummy entry, so the index should as is.
If `diff` == 4, it means the classes doesn't have padding, so we should -4 to adjust the index

SidneyLann:
I finetune yolov8 in my dataset and get a new yolov8 model for 96 classes, there are 96 entries in my synset.txt and has no dummy  entries. now the diff is 4(doesn't have padding)? 

int padding = nClasses - classes.size();// nClasses is from synset.txt and equals to 96? classes.size() is from the model and equals to 96? 

So the padding(diff) is 0?

frankfliu:
If you trained your model with 96 classes, the output should should be (8400, 100)

`nClasses - classes.size()` should be 4

PR #3004, opened 2024-02-22
Update continuous OSX to 13

zachgk:
This is to fix the flaky llama engine segfaults on CI.


frankfliu (approved):

frankfliu:
I think we need to brew install libomp for mac13

frankfliu:
see: https://github.com/deepjavalibrary/djl/blob/master/.github/workflows/nightly_publish.yml#L37-L39

PR #3005, opened 2024-02-26
Fixes cases where the getEngine method in the EngineProvider class returns null when called concurrently.

onaple:
## Description ##

When I create models concurrently using the following methods, I always encounter a null pointer exception because of a problem with the Engine.getEngine("PyTorch ") double-check lock. I wonder if the author has any other intention in this way. However, it is necessary for users to deal with outer layer exceptions.

Model localModel = Engine.getEngine("PyTorch").newModel("test", Device.cpu())

### DIFF
--- a/api/src/main/java/ai/djl/engine/Engine.java
+++ b/api/src/main/java/ai/djl/engine/Engine.java
@@ -59,7 +59,7 @@ public abstract class Engine {
 
     private static final Map<String, EngineProvider> ALL_ENGINES = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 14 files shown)


frankfliu:
@onaple 
Thanks for report this issue. This bug was introduced by this PR: #2885

I think we should just revert #2885 completely.

frankfliu:
@onaple 

DJL 0.25.0 doesn't have this issue. You can use 0.25.0 for the mean time.

frankfliu (approved):

Issue #179, opened 2020-09-29
Running two models with same engine at once


tspannhw:
## Description

Cannot run two models in NiFi 1.11.4 if both using Pytorch engine.

### Expected Behavior

Both should be able to use the library

### Error Message

java.lang.UnsatisfiedLinkError: /Users/tspann/.djl.ai/pytorch/1.6.0-cpu-osx-x86_64/0.8.0-SNAPSHOT-cpu-libdjl_torch.dylib (Library is already loaded in another ClassLoader)

## How to Reproduce?

https://github.com/tspannhw/nifi-djlsentimentanalysis-processor

### Steps to reproduce
(Paste the commands you ran that produced the error.)

1.   install NiFi nars for both models
2. run at same time

## What have you tried to solve it?

1.   Tweaked poms
2.  Tried different engines, but that's a stop gap and I want Pytorch for both Bert QA and Sentiment.

## Environment Info

Run my nifi processors
https://github.com/tspannhw/nifi-djlqa-processor
https://github.com/tspannhw/nifi-djlsentimentanalysis-processor

```
java.lang.UnsatisfiedLinkError: /Users/tspann/.djl.ai/pytorch/1.6.0-cpu-osx-x86_64/0.8.0-SNAPSHOT-cpu-libdjl_torch.dylib (Library is already loaded in another ClassLoader)
	at java.base/java.lang.ClassLoader.loadLibraryWithPath(ClassLoader.java:1742)
	at java.base/java.lang.System.load(System.java:585)
	at ai.djl.pytorch.jni.LibUtils.loadLibrary(LibUtils.java:83)
	at ai.djl.pytorch.engine.PtEngine.newInstance(PtEngine.java:42)
	at ai.djl.pytorch.engine.PtEngineProvider.getEngine(PtEngineProvider.java:27)
	at ai.djl.engine.Engine.initEngine(Engine.java:49)
	at ai.djl.engine.Engine.<clinit>(Engine.java:44)
	at ai.djl.repository.zoo.BaseModelLoader.loadModel(BaseModelLoader.java:129)
	at ai.djl.repository.zoo.ModelZoo.loadModel(ModelZoo.java:162)
	at com.dataflowdeveloper.djlsentimentanalysis.SentimentAnalysisService.predict(SentimentAnalysisService.java:69)
	at com.dataflowdeveloper.djlsentimentanalysis.DeepLearningSAProcessor.onTrigger(DeepLearningSAProcessor.java:95)
[...]

PR #3020, opened 2024-03-05
Avoid duplicated loading native library

frankfliu:
## Description ##

### DIFF
--- a/engines/llama/src/main/java/ai/djl/llama/jni/LibUtils.java
+++ b/engines/llama/src/main/java/ai/djl/llama/jni/LibUtils.java
@@ -55,8 +55,9 @@ public final class LibUtils {
             String nativeHelper = System.getProperty("ai.djl.llama.native_helper");
             if (nativeHelper != null && !nativeHelper.isEmpty()) {
                 ClassLoaderUtils.nativeLoad(nativeHelper, path);
+            } else {
+                System.load(path); // NOPMD
             }
-            System.load(path); // NOPMD
         }
     }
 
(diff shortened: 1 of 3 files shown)


zachgk (approved):

humcqc:
Hello, thanks for the fix,
Could you tell me if this will fix the https://github.com/langchain4j/langchain4j/issues/776

I'm  not sure as the exception is throw in the System.load call:
java.lang.UnsatisfiedLinkError: Native Library /Users/gkwan/.djl.ai/tokenizers/0.15.0-0.26.0-osx-aarch64/libtokenizers.dylib already loaded in another classloader
[INFO]  at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:201)
[INFO]  at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:174)
[INFO]  at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2389)
[INFO]  at java.base/java.lang.Runtime.load0(Runtime.java:755)
[INFO]  at java.base/java.lang.System.load(System.java:1953)
[INFO]  at ai.djl.huggingface.tokenizers.jni.LibUtils.loadLibrary(LibUtils.java:76)
[INFO]  at ai.djl.huggingface.tokenizers.jni.LibUtils.<clinit>(LibUtils.java:41)

Could we add a check to see if the library is already loaded or just catch the exception in case of and log a warning ?

Thanks

frankfliu:
@humcqc 

I don't think this will fix your issue. Your issue is a typical java ClassLoader issue. see similar discussion: https://github.com/deepjavalibrary/djl/issues/179

You are loading the DJL tokenizer into two different ClassLoader, which is not allowed by JVM. You need to make sure the .so file is loaded in the common parent class loader to avoid this bug.

Here is something you can do:
1. Use `Thread.currentThread().setContextClassLoader(parentClassLoader)`, to set a common class loader before you load tokenizer in your thread.
2. or you can write a "NativeHelper" class, and put this class in system classpath, and set system property:
```
System.setProperty("ai.djl.huggingface.native_helper", "com.examples.NativeHelper");
```

humcqc:
thanks @frankfliu , do you think we can apply your first suggestion directly to your LibUtils to find a centralized solution ?

PR #3021, opened 2024-03-05
[ci] Fixes nightly build for onnx 1.17.1

frankfliu:
## Description ##

### DIFF
--- a/.github/workflows/nightly_publish.yml
+++ b/.github/workflows/nightly_publish.yml
@@ -123,8 +123,9 @@ jobs:
         run: |
           ./gradlew :integration:test "-Dai.djl.default_engine=PyTorch"
           ./gradlew :integration:clean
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly_publish.yml
```diff
@@ -123,8 +123,9 @@ jobs:
         run: |
           ./gradlew :integration:test "-Dai.djl.default_engine=PyTorch"
           ./gradlew :integration:clean
-          ./gradlew :integration:test "-Dai.djl.default_engine=OnnxRuntime"
-          ./gradlew :integration:clean
+          # OnnxRuntime 1.17.1 requires GLIBC 2.27
```
siddvenk, line 126: Is there another OS we can use when testing with ARM?

GH self-hosted runners will have arm64 machines soon, maybe we just wait for that and we can move away from our self hosted arm64 runners and avoid all these GLIBC version issues

siddvenk (approved):

PR #3045, opened 2024-03-27
Avoid UnsatisfiedLinkError when native library already loaded 

humcqc:
In certain condition the library is already loaded by another classloader. 
And this exception should not stop the app.


frankfliu (commented):
This fix won't work. Native library loaded in one class loader is not visible to sibling class loader.

humcqc:
After some investigation in the java code, you right there is a static set that contains already loaded Libraries name but not the library itself that is store in the specific classloader. Not sure we can use the workaround but will try.

PR #3047, opened 2024-03-28
Update checkstyle tool version to 10.14.2

xyang16:
## Description ##

### DIFF
--- a/tools/conf/checkstyle.xml
+++ b/tools/conf/checkstyle.xml
@@ -156,7 +156,6 @@
                     value="Override, Test, Before, After, BeforeClass, AfterClass"/>
         </module>
         -->
-        <module name="FinalClass"/>
         <module name="HideUtilityClassConstructor"/>
         <module name="InnerTypeLast"/>
         <!--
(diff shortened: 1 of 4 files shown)


tools/conf/checkstyle.xml
```diff
@@ -284,8 +284,8 @@
         <!-- Javadoc Comments -->
         <module name="AtclauseOrder"/>
         <module name="JavadocMethod">
-            <property name="allowUndeclaredRTE" value="true"/>
```
xyang16, line 287: These two options are deprecated.

api/src/main/java/ai/djl/modality/cv/translator/InstanceSegmentationTranslator.java
```diff
@@ -149,7 +149,7 @@ public static Builder builder(Map<String, ?> arguments) {
     }
 
     /** Resizes the image based on the shorter edge or maximum edge length. */
-    private class ResizeShort implements Transform {
```
xyang16, line 152: Fix error: Class ResizeShort should be declared as final.

frankfliu (approved):

PR #3052, opened 2024-03-31
[tokenizer] Apply Dense layer for TextEmbeddingTranslator

### LINKED ISSUE #3044: Prediction from clip-ViT-B-32-multilingual-v1 incorrectly returns 768 elements rather than 512
## Description
Prediction from clip-ViT-B-32-multilingual-v1 incorrectly returns 768 elements.

### Expected Behavior

Returns 512 element vectors, as shown by this Python code:

```python
from sentence_transformers import SentenceTransformer, util
import numpy as np

model = SentenceTransformer('sentence-transformers/clip-ViT-B-32-multilingual-v1')
emb = model.encode('fast blue car')
np_array = np.array(emb)

print("Length: ")
print(len(np_array))

print("Result: ")
print(' '.join(map(str, np_array)))
```
This returns:

```
Length: 
512
Result: 
0.12500623 -0.40152556 -0.044941768 ...
```

## How to Reproduce?
Here is an associated unit test:

```java
    @Test
    public void testIssue() throws Exception
    {
        Criteria<String, float[]> criteria =
            Criteria.builder()
                    .setTypes(String.class, float[].class)
                    .optModelUrls("djl://ai.djl.huggingface.pytorch/sentence-transformers/clip-ViT-B-32-multilingual-v1")
                    .optTranslatorFactory(new TextEmbeddingTranslatorFactory())
                    .optEngine("PyTorch")
                    .build();
        try (ZooModel<String, float[]> model = criteria.loadModel();
             Predictor<String, float[]> predictor = model.newPredictor())
        {
            float[] embeddings = predictor.predict("fast blue car");
            assertThat(embeddings.length, is(512));
            assertThat((double) embeddings[0], is(closeTo(0.12500623, 0.0001)));
        }
    }
```

### Steps to reproduce
Run the above unit test.

### Thoughts

https://huggingface.co/sentence-transformers/clip-ViT-B-32-multilingual-v1/discussions/4 reports a similar issue when trying to use Optimum with this model, where apparently it was ignoring the pooling and dense layer inclusion (which reduces the dimensions to 512).  Could a similar issue be happening here?

frankfliu:
Fixes #3044

## Description ##

### DIFF
--- a/extensions/tokenizers/src/main/java/ai/djl/huggingface/translator/TextEmbeddingTranslator.java
+++ b/extensions/tokenizers/src/main/java/ai/djl/huggingface/translator/TextEmbeddingTranslator.java
@@ -12,18 +12,24 @@
  */
 package ai.djl.huggingface.translator;
 
+import ai.djl.Device;
 import ai.djl.huggingface.tokenizers.Encoding;
 import ai.djl.huggingface.tokenizers.HuggingFaceTokenizer;
 import ai.djl.ndarray.NDArray;
 import ai.djl.ndarray.NDList;
 import ai.djl.ndarray.NDManager;
 import ai.djl.ndarray.types.DataType;
+import ai.djl.ndarray.types.Shape;
 import ai.djl.translate.ArgumentsUtil;
 import ai.djl.translate.Batchifier;
 import ai.djl.translate.Translator;
 import ai.djl.translate.TranslatorContext;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Map;
 
 /** The translator for Huggingface text embedding model. */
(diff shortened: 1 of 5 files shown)


extensions/tokenizers/src/main/java/ai/djl/huggingface/translator/TextEmbeddingTranslatorFactory.java
```diff
@@ -62,8 +64,17 @@ public <I, O> Translator<I, O> newInstance(
                             .optTokenizerPath(modelPath)
                             .optManager(model.getNDManager())
                             .build();
-            TextEmbeddingTranslator translator =
-                    TextEmbeddingTranslator.builder(tokenizer, arguments).build();
+            TextEmbeddingTranslator.Builder builder =
+                    TextEmbeddingTranslator.builder(tokenizer, arguments);
+
+            String mode = Utils.getEnvOrSystemProperty("DJL_SENTENCE_TRANSFORMER_MODE");
```
zachgk, line 70: Why is this an env variable and not an argument or model property? Because this mode seems to be local to the model rather than apply to all models.

zachgk (commented):
Can you add some documentation about this option somewhere?

extensions/tokenizers/src/main/java/ai/djl/huggingface/translator/TextEmbeddingTranslatorFactory.java
frankfliu: Our current model zoo model doesn't apply this Dense layer. If we modify the serving.properties in the existing model zoo model, will break existing behavior. With this env var, we can enable new behavior without impact existing user.

zachgk (approved):

PR #3055, opened 2024-04-01
[llama.cpp] Update llama.cpp to latest release b2581 (2024-03-30)

howlger:
In order to get support for BERT based sentence embedding models like [BAAI/bge-base-en-v1.5](https://huggingface.co/BAAI/bge-base-en-v1.5), [mixedbread-ai/mxbai-embed-large-v1](https://huggingface.co/mixedbread-ai/mxbai-embed-large-v1), or others, update [llama.cpp](https://github.com/ggerganov/llama.cpp) from [b1696](https://github.com/ggerganov/llama.cpp/releases/tag/b1696) (2023-12-12) to the current latest release [b2581](https://github.com/ggerganov/llama.cpp/releases/tag/b2581) (2024-03-30).

BERT support was added to llama.cpp in February 2024: https://github.com/ggerganov/llama.cpp/pull/5423

This change has not yet been tested. Maybe updating the Gradle property `llamacpp_version` is not enough and [`ai_djl_llama.cpp`](https://github.com/deepjavalibrary/djl/blob/master/engines/llama/src/main/native/ai_djl_llama.cpp) needs to be adapted as well. If so, please do so.


frankfliu (changes requested):
Llama.cpp implementation has changed, just bump up version won't work, we have to make JNI code change to make it compile

howlger:
I see. Thanks for taking the time to try it.

The native libraries are built for all platforms with [ _Native S3 llama.cpp_](https://github.com/deepjavalibrary/djl/blob/master/.github/workflows/native_s3_llama.yml), right? I can't find the log of the failed build. Could you please share it?

PR #3084, opened 2024-04-13
[api] Add batch support to TextEmbeddingServingTranslator

frankfliu:
- **[api] Support "inputs" key in json request for text embedding translator**
- **[api] Add batch support to TextEmbeddingServingTranslator**

## Description ##

### DIFF
--- a/api/src/main/java/ai/djl/inference/Predictor.java
+++ b/api/src/main/java/ai/djl/inference/Predictor.java
@@ -34,7 +34,9 @@ import ai.djl.translate.TranslatorContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.lang.reflect.Array;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


api/src/main/java/ai/djl/inference/Predictor.java
```diff
@@ -302,40 +308,30 @@ private NDList processInputs(TranslatorContext ctx, List<I> inputs) throws Excep
...
-        return outputs;
-    }
-
-    private void preprocessEnd(NDList list) {
+    private void preprocessEnd(NDList list, int batchSize) {
         if (metrics != null) {
             waitToRead(list);
             long tmp = System.nanoTime();
-            long duration = (tmp - timestamp) / 1000;
+            long duration = (tmp - timestamp) / 1000 / batchSize;
```
siddvenk, line 315: does it make sense to report the average latency per item as the preprocess latency? do we also report the batch size metric somewhere?

I feel like we should report the total time for the full batch, but if this is consistent with how we're doing it elsewhere, or what users typically expect, i'm ok with it. same foes for the other methods we report a similar metric for

xyang16 (approved):

api/src/main/java/ai/djl/inference/Predictor.java
frankfliu: try to align with TEI for batch case. We do report DynamicBatchSize in metrics. We have total inference time as well

PR #3088, opened 2024-04-15
[fix] Fix HfModelZoo and LlamaModelZoo concurrent init problem

ewan0x79:
## Description ##

- This PR addresses a potential thread safety issue in the **ai.djl.huggingface.zoo.HfModelZoo** and **ai.djl.llama.zoo.LlamaModelZoo** classes. Both classes utilize an init method for initialization and are designed as singletons(**ai/djl/repository/zoo/ModelZoo.java:39**), which may be accessed by multiple threads simultaneously. However, the init methods do not currently implement any mechanism to ensure thread safety. This PR proposes changes to synchronize the init methods to prevent concurrent initialization issues.

If this change is a backward incompatible change, why must this change be made?
- This change is not a backward incompatible change. It is a necessary enhancement to ensure that the initialization process of the singletons is thread-safe, thus preventing potential race conditions and data corruption when used in a multi-threaded environment.

### DIFF
--- a/engines/llama/src/main/java/ai/djl/llama/zoo/LlamaModelZoo.java
+++ b/engines/llama/src/main/java/ai/djl/llama/zoo/LlamaModelZoo.java
@@ -39,6 +39,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.zip.GZIPInputStream;
 
 /** LlamaModelZoo is a repository that contains llama.cpp models. */
@@ -52,7 +53,7 @@ public class LlamaModelZoo extends ModelZoo {
 
     private static final long ONE_DAY = Duration.ofDays(1).toMillis();
 
-    private boolean initialized;
+    private final AtomicBoolean initialized = new AtomicBoolean(false);
 
     LlamaModelZoo() {}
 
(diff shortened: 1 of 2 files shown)


frankfliu (commented):
Are you observing any concurrent issue with HfModelZoo?

You PR won't resolve concurrent initialization issue, it actually make it worse.

1. original implementation may cause unnecessary initialization when multiple threads calling `getModelLoaders()` for the first, but it should not causing any issue.
2. on the other hand, your implementation may cause return empty modelLoaders if first call is still processing, but it already marked initialized as true

ewan0x79:
> Are you observing any concurrent issue with HfModelZoo?
> 
> You PR won't resolve concurrent initialization issue, it actually make it worse.
> 
> 1. original implementation may cause unnecessary initialization when multiple threads calling `getModelLoaders()` for the first, but it should not causing any issue.
> 2. on the other hand, your implementation may cause return empty modelLoaders if first call is still processing, but it already marked initialized as true


Thank you for your reply. You are correct; my intention was to address the issue of resource wastage due to redundant initializations that can occur when multiple threads load models simultaneously. However, my approach did indeed overlook the possibility of returning a null object. A double-checked locking pattern should be used to properly address the issue.    
```java
private volatile boolean initialized = false;

private void init() {
    if (!initialized) {
        synchronized (HfModelZoo.class) {
            if (!initialized) {
                Version version = new Version(Engine.getDjlVersion());
                addModels(NLP.FILL_MASK, version);
                addModels(NLP.QUESTION_ANSWER, version);
                addModels(NLP.TEXT_CLASSIFICATION, version);
                addModels(NLP.TEXT_EMBEDDING, version);
                addModels(NLP.TOKEN_CLASSIFICATION, version);
                initialized = true;
            }
        }
    }
}
```
Do you think this is a necessary fix?

frankfliu (commented):
Thanks, this should work.

But double lock singleton is anti-pattern. We'd better use InstanceHolder, see: https://github.com/deepjavalibrary/djl/blob/master/engines/tflite/tflite-engine/src/main/java/ai/djl/tflite/engine/TfLiteEngineProvider.java#L39-L41

frankfliu (approved):

PR #3120, opened 2024-04-24
[tokenizer] add optional tokenizerPath Prior to modelPath

ewan0x79:
## Description ##

In our actual usage, there are scenarios where `tokenizerPath` and `modelPath` differ. In the `TranslatorFactory`, we check if the `tokenizerPath` field is set, which has a higher priority than `modelPath`. If `tokenizerPath` is not configured, `modelPath` will be used by default, ensuring no impact on existing code.


frankfliu:
@ewan0x79 

We actually only need make changes to HuggingfaceTokenizer.builder(), the arguments in criteria will override .optTokenizerPath().

frankfliu (approved):

PR #3130, opened 2024-04-25
[api] Closes file stream

### LINKED ISSUE #3127: Resource Leak Issue with Files.list in ai.djl.util.Utils
## Description

In the ai.djl.util.Utils class of the ai.djl:api version 0.26.0, there appears to be an issue with the usage of Files.list in the getNestedModelDir method. Specifically, the stream returned by Files.list(modelDir) is not being closed properly, which may lead to file descriptor leaks.

### Expected Behavior

The expected behavior is that any resources opened via Files.list(modelDir) should be closed after their use to prevent file descriptor exhaustion and ensure proper resource management.

### Error Message

N/A (This is a potential resource leak issue rather than a runtime error that generates a message.)

## How to Reproduce?

The issue can be observed in the code snippet provided but might not manifest as a direct error during execution. It is a best practice concern rather than a functional bug that crashes the application.

### Steps to reproduce

Since this is a resource management issue, it doesn't produce a direct error trace but can be verified through code inspection or static analysis tools that check for resource leaks.

1. Inspect the code of ai.djl.util.Utils.getNestedModelDir, ai.djl.util.Utils.getCurrentEpoch, ai.djl.util.ClassLoaderUtils.scanDirectory, and ai.djl.util.ClassLoaderUtils.findImplementation.
2. Notice the lack of try-with-resources or explicit .close() calls on the Stream<Path> returned by Files.list().

I propose incorporating try-with-resources to ensure the Stream is closed automatically:

~~~java
public static Path getNestedModelDir(Path modelDir) {
    if (Files.isDirectory(modelDir)) {
        try (Stream<Path> paths = Files.list(modelDir)) {
            // handle actual model directory is subdirectory case
            List<Path> files = paths
                    .filter(p -> !p.getFileName().toString().startsWith("."))
                    .collect(Collectors.toList());
            if (files.size() == 1 && Files.isDirectory(files.get(0))) {
                return files.get(0);
            }
        } catch (IOException e) {
            throw new AssertionError("Failed to list files: " + modelDir, e);
        }
    }
}
~~~

Similar modifications should be applied to the mentioned methods in Utils and ClassLoaderUtils.

## Environment Info

N/A

frankfliu:
Fixes #3127

## Description ##

### DIFF
--- a/api/src/main/java/ai/djl/repository/LocalRepository.java
+++ b/api/src/main/java/ai/djl/repository/LocalRepository.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 
 /**
  * A {@code LocalRepository} is a {@link Repository} located in a filesystem directory.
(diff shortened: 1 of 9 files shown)


xyang16:
This PR has a lot of other commits in it.

xyang16 (approved):

PR #3133, opened 2024-04-26
[rust] Avoid panic in error case

frankfliu:
## Description ##

### DIFF
--- a/extensions/tokenizers/src/main/java/ai/djl/engine/rust/RsModel.java
+++ b/extensions/tokenizers/src/main/java/ai/djl/engine/rust/RsModel.java
@@ -49,6 +49,14 @@ public class RsModel extends BaseModel {
                     "Model directory doesn't exist: " + modelPath.toAbsolutePath());
         }
         modelDir = modelPath.toAbsolutePath();
+        Path config = modelDir.resolve("config.json");
+        if (!Files.isRegularFile(config)) {
+            throw new FileNotFoundException("config.json file not found");
+        }
+        Path file = modelDir.resolve("model.safetensors");
+        if (!Files.isRegularFile(file)) {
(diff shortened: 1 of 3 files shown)


extensions/tokenizers/src/main/java/ai/djl/engine/rust/RsModel.java
```diff
@@ -49,6 +49,14 @@ public void load(Path modelPath, String prefix, Map<String, ?> options)
                     "Model directory doesn't exist: " + modelPath.toAbsolutePath());
         }
         modelDir = modelPath.toAbsolutePath();
+        Path config = modelDir.resolve("config.json");
+        if (!Files.isRegularFile(config)) {
+            throw new FileNotFoundException("config.json file not found");
+        }
+        Path file = modelDir.resolve("model.safetensors");
```
siddvenk, line 56: Do we need to handle sharded checkpoints here?

frankfliu: We only support safetensors

siddvenk: Is it always the case for text-embedding models that they are provided in a single checkpoint? It seems from huggingface models that is overwhelmingly true, but there are some big text-embedding models in the single of billions parameter range (5b-7b) that are typically sharded. Example https://huggingface.co/Alibaba-NLP/gte-Qwen1.5-7B-instruct, https://huggingface.co/intfloat/e5-mistral-7b-instruct

It's not super important for now, but at some point we'll probably need to handle sharded checkpoints even for embedding models since the user is free to specify how many shards/max file size of checkpoints

xyang16 (approved):

PR #3138, opened 2024-04-29
Ensure the alternative ND manager can use GPUs

david-sitsky:
This makes a massive performance difference when using OnnxRuntime, which delegates to PyTorch for some NDArray operations.  This almost makes OnnxRuntime run as fast compared to PyTorch.  Without this change, OnnxRuntime was over 3 times slower.

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
+++ b/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
@@ -62,7 +62,13 @@ public abstract class BaseNDManager implements NDManager {
         uid = UUID.randomUUID().toString();
         Engine engine = getEngine().getAlternativeEngine();
         if (engine != null) {
-            alternativeManager = engine.newBaseManager(Device.cpu());
+            try {
+                // Try to use the same device for efficiency.
+                alternativeManager = engine.newBaseManager(device);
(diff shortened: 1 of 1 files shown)


zachgk (approved):

david-sitsky:
@frankfliu - on my tests with my application (which does a lot of non-GPU work too), my original change ran in 6m 8s, but with the change not to keep the device, this went to 6m 25s.  So it is a significant performance hit on multi-GPU systems, and for other workloads the discepancy will be much higher.  

I also saw GPU 0 having way higher (too high) utilisation compared to the other GPUs since this is the default device set for the alternative ND manager, for all my worker processes (16 on my test).

Can't we keep my original change without the catch for UnsatisfiedLinkError?  Otherwise this puts OnnxRuntime further away from PyTorch again, which won't have this issue.

frankfliu:
@david-sitsky 

You are right, in multiple GPU case, it does make sense to keep original device. I just doesn't like try catch. Let me think a bit, there should a simple way.

david-sitsky:
@frankfliu - any ideas with this?  I am pretty keen to get that multi-GPU performance back. :wink:

frankfliu:
@david-sitsky 

I created a PR: #3146

david-sitsky:
Awesome - thank you.  That looks like a better way of achieving it too.

PR #3139, opened 2024-04-29
[pytorch] optimize model load process, add last chance for .pt model

ewan0x79:
## Description ##
- To optimize the PyTorch model loading process, an additional 'last chance' to load a .pt model can be introduced into the model loading workflow. This means that after all other methods have been attempted without success, we will try to find any .pt file within the specified directory.

### DIFF
--- a/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtModel.java
+++ b/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtModel.java
@@ -38,6 +38,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * {@code PtModel} is the PyTorch implementation of {@link Model}.
@@ -50,7 +51,7 @@ public class PtModel extends BaseModel {
     /**
      * Constructs a new Model on a given device.
      *
-     * @param name the model name
+     * @param name   the model name
      * @param device the device the model should be located on
      */
     PtModel(String name, Device device) {
@@ -60,7 +61,9 @@ public class PtModel extends BaseModel {
         dataType = DataType.FLOAT32;
     }
 
-    /** {@inheritDoc} */
+    /**
+     * {@inheritDoc}
+     */
     @Override
     public void load(Path modelPath, String prefix, Map<String, ?> options)
             throws IOException, MalformedModelException {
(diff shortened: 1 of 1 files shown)


engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtModel.java
```diff
@@ -78,8 +81,18 @@ public void load(Path modelPath, String prefix, Map<String, ?> options)
...
-                String fileName = prefix.endsWith(".pt") ? prefix : prefix + ".pt";
-                throw new FileNotFoundException(fileName + " file not found in: " + modelDir);
+                /*
+                last chance to load model
+                This is the last attempt to load the model. If no specific model file has been successfully identified
+                by previous steps, this call attempts to find any .pt file within the model directory.
+                If exactly one .pt file is found, it will be used as the model file.
+                If there are none or more than one .pt files, an exception will be thrown or null will be returned,
+                */
+                modelFile = findAnyModelFile(modelDir);
```
frankfliu, line 91: I don't think this is a good idea to treat arbitrary  .pt file as model file. We already provided many ways to identify the model file:
1. You can specify `option.modelName` in `serving.properties` file
2. You can pass `?modelName=xxx` in modelUrl
3. You can point the modelPath to the `.pt` file in `.optModelPath(modelPath)`
4. Your .pt file has the same name as the folder name, the .zip file or the modelName in the model
5. You can save the model file as `model.pt`

In DJLServing project, we use the same rule to auto detect an engine, allowing arbitrary .pt file will name the auto detection much compilcated.

ewan0x79: You made a valid point that adding a "last chance" to load any model file would increase complexity. However, when using `ai.djl.huggingface.translator.TokenClassificationTranslator`, I noticed that during the preparation process, it attempts to load a configuration file from the model path with the following code:

```java
/** {@inheritDoc} */
@Override
public void prepare(TranslatorContext ctx) throws IOException {
    Path path = ctx.getModel().getModelPath();
    Path file = path.resolve("config.json");
    try (Reader reader = Files.newBufferedReader(file)) {
        config = JsonUtils.GSON.fromJson(reader, PretrainedConfig.class);
    }
}
```

This means I cannot use the absolute path of a `.pt` file to load the model. Instead, I need to use the directory containing the `.pt` file when calling `ai.djl.Model#load(java.nio.file.Path)` to ensure the preparation process can find the configuration file. However, the process of searching for a `.pt` file with the model name, folder name, or "model.pt" seems to assume the scenario where the model is saved by the user. In reality, for instance, when deploying DJL to a Spark job, the names of the model files we pull, including the zip archive and the `.pt` file, are beyond our control. Different users save model files with different names, and when models are uploaded as zip archives, they are often encoded or appended with additional information, such as timestamps, rather than retaining their original names. This has led to issues during loading. You mentioned that setting the modelName method could solve some problems. Currently, I've resolved this by unzipping the archive, searching for the `.pt` file, and renaming it to "model.pt".Upon further observation, it seems that mainstream frameworks do not support loading any model from a model directory, indicating that this approach might indeed be unsuitable. It appears I may have to continue resolving this issue by renaming the model files.

frankfliu:
1. You still can use absolute path of a `.pt` file, the folder contains `.pt` will be treated as `modelPath`, we are looking for `config.json` in the same folder.
2. If you put a `serving.properties` file in the model folder, you can specify the `option.modelName`
3. You can also specify `.optModelName()` in `Criteria.builder()`, you can keep your scan logic out side the DJL if want to.

PR #3144, opened 2024-04-30
[ci] Update github action runner to macOS x86_64 instance

frankfliu:
## Description ##

### DIFF
--- a/.github/workflows/native_build_mxnet_osx.yml
+++ b/.github/workflows/native_build_mxnet_osx.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   build:
-    runs-on: macos-latest
+    runs-on: macos-13
 
     steps:
       - name: Checkout Apache MXNet repository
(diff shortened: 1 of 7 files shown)


siddvenk (approved):

justin-octo:
There is no description on this.  The change involves version locking instead of using latest.  Also removing macos from the nightly?
Seems a valid explanation of why this is a good idea should be in the description.

zachgk:
@justin-octo This PR is intended to retain a number of our engine build scripts and the nightly tensorflow on OSX x86_64 rather than arm, mostly because these scripts were not supposed to be using arm. It was a change on github actions that latest now refers to arm rather than having x86 remain the default. In the future, these scripts may need to be migrated to further expand our osx arm support, especially as all osx support moves to arm. This change fixes them for the moment.

For nightly, this is only removing the tensorflow part of nightly for which we have never supported osx arm. As a temporary measure, we are going to stop testing tf on osx. If we do add the engine support, we will add this back to nightly as well.

PR #3187, opened 2024-05-14
Updates on djl_converter

xyang16:
## Description ##

### DIFF
--- a/extensions/tokenizers/src/main/python/requirements.txt
+++ b/extensions/tokenizers/src/main/python/requirements.txt
@@ -3,5 +3,5 @@ transformers
 sentence_transformers
 torch
 protobuf==3.20.2
-optimum[exporters,onnxruntime]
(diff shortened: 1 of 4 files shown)


extensions/tokenizers/src/main/python/requirements.txt
```diff
@@ -3,5 +3,5 @@ transformers
 sentence_transformers
 torch
 protobuf==3.20.2
-optimum[exporters,onnxruntime]
```
frankfliu, line 6: Why remove onnxruntime?

xyang16: This will install onnxruntime cpu version, this will make O4 optimization fail.

xyang16: Changed.

frankfliu (approved):

PR #3213, opened 2024-05-23
[tokenizer] Process text embedding input and output in stacked NDArray

xyang16:
## Description ##

### DIFF
--- a/extensions/tokenizers/src/main/java/ai/djl/huggingface/translator/TextEmbeddingTranslator.java
+++ b/extensions/tokenizers/src/main/java/ai/djl/huggingface/translator/TextEmbeddingTranslator.java
@@ -237,7 +237,7 @@ public class TextEmbeddingTranslator implements Translator<String, float[]> {
         embeddings = embeddings.duplicate();
         embeddings.set(inputAttentionMask, -1e9); // Set padding tokens to large negative value
 
-        return embeddings.max(AXIS, true);
+        return embeddings.max(AXIS, false);
     }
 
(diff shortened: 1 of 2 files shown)


extensions/tokenizers/src/main/java/ai/djl/huggingface/translator/TextEmbeddingTranslator.java
```diff
@@ -36,7 +36,7 @@
 /** The translator for Huggingface text embedding model. */
 public class TextEmbeddingTranslator implements Translator<String, float[]> {
 
-    private static final int[] AXIS = {0};
+    private static final int[] AXIS = {1};
```
frankfliu, line 39: Should this be -2?

xyang16: -2 should work.

extensions/tokenizers/src/main/java/ai/djl/huggingface/translator/TextEmbeddingTranslator.java
```diff
@@ -193,7 +212,7 @@ private NDArray processEmbedding(NDManager manager, NDList list, Encoding encodi
                             .get(0);
         }
         if (normalize) {
-            embedding = embedding.normalize(2, 0);
+            embedding = embedding.normalize(2, 1);
```
frankfliu, line 215: This should also be -2, this function should work for both batch and unbatched case.

frankfliu (approved):

PR #3225, opened 2024-05-29
add close method to all nd manager

lanking520:
## Description ##

Close all alternative ndmanager


extensions/tokenizers/src/main/java/ai/djl/engine/rust/RsNDManager.java
```diff
@@ -256,6 +256,16 @@ int toRustDataType(DataType dataType) {
         }
     }
 
+    /** {@inheritDoc} */
```
frankfliu, line 259: Rust is a full engine, it doesn't have alternativeManager

PR #3226, opened 2024-05-30
Improved build performance for repeated builds

benjie332:
## Description ##

Some Gradle tasks were configured to prevent Gradle from using up-to-date checks (and build caching, if enabled). This meant that some tasks were being executed even if the output was not different, wasting build time.

This PR proposes changes to the build scripts to make the builds more efficient for repeated builds on the same machine. This should save a few seconds, and if build caching were enabled, there would be more improvements.

Some key highlights:

- tasks that share output locations (eg. tasks that downloaded files directly into build/classes) would prevent Gradle from identifying the owning tasks, which disables caching/up-to-date checks. These were split into separate locations and left to the processResources/jar tasks to collect. The file was placed directly into the source for some modules, and the version property was injected to ensure repeatability in the build. Feedback on this change is welcome!

- tasks that do not declare inputs/outputs or have no outputs also disable the checks. These tasks were updated to include the right inputs/outputs. In the case of the verifyJava task, a new output file was created to let Gradle know when the task was completed successfully previously. Gradle will automatically skip these tasks if inputs do not change, as the output would have remained the same.

- absolute paths were changed to relative paths (or normalized to only consider relative paths). If remote build caches are enabled, the build cache will be used for these tasks across any connected build machine.

The improvements were discovered through analyzing the build with [Develocity](https://gradle.com/develocity/). This [Build Scan](https://scans.gradle.com/s/w2yxmjzm2ciqm) is of a second build run after the improvements were implemented.

The [timeline view](https://scans.gradle.com/s/w2yxmjzm2ciqm/timeline) shows a breakdown of where time was spent in the build. In this case, as it was performed on a machine that had already run the build task, you can see [here](https://scans.gradle.com/s/w2yxmjzm2ciqm/timeline?outcome=success,failed) that no tasks were executed as they were all up to date.

Feedback and comments are more than welcome!

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -17,12 +17,8 @@ dependencies {
 compileJava.dependsOn(processResources)
 
 processResources {
-    outputs.file file("${project.buildDir}/classes/java/main/ai/djl/engine/api.properties")
-    doFirst {
-        def classesDir = file("${project.buildDir}/classes/java/main/ai/djl/engine/")
-        classesDir.mkdirs()
-        def propFile = new File(classesDir, "api.properties");
-        propFile.text = "djl_version=${project.version}"
+    filesMatching("**/api.properties") {
+        expand projectVersion: project.version
     }
 }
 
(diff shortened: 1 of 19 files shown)


benjie332:
Looks like bad timing on my part here 😄 Let me review my changes

PR #3231, opened 2024-05-31
Improved incremental build

benjie332:
## Description ##

Some Gradle tasks were configured to prevent Gradle from using up-to-date checks (and build caching, if enabled). This meant that some tasks were being executed even if the output was not different, wasting build time.

This PR proposes changes to the build scripts to make the builds more efficient for repeated builds on the same machine. This should save a few seconds, with more improvements if build caching were to be enabled.

Some key highlights:

- tasks that share output locations (eg. tasks that downloaded files directly into build/classes) would prevent Gradle from identifying the owning tasks, which disables caching/up-to-date checks. These were split into separate locations and left to the `processResources`/`jar` tasks to collect. The property files are now part of the source tree, and the properties are now injected by the `processResources` task.

- the Spotbugs plugin was reconfigured to have separate paths for the main and test code, for the same reasons as above.

- tasks that do not declare inputs/outputs or have no outputs also disable the checks. These tasks were updated to include the right inputs/outputs. In the case of the verifyJava task, a new output file was created to let Gradle know when the task was completed successfully previously. Gradle will automatically skip these tasks if inputs do not change, as the output would have remained the same.

- properties were kept constant where possible to allow the incremental checks to work. I removed the system properties pass-through that was present in the tests. I could not find why this was added and the tests passed after removal in any case. Feedback is welcome on this.

The improvements were discovered through analyzing the build with [Develocity](https://gradle.com/develocity/). 

This [Build Scan](https://scans.gradle.com/s/mra7iypfm3yic) is of a second build run after the improvements were implemented. The [timeline view](https://scans.gradle.com/s/mra7iypfm3yic/timeline) shows a breakdown of where time was spent in the build. In this case, as it was performed on a machine that had already run the build task, you can see [here](https://scans.gradle.com/s/mra7iypfm3yic/timeline?outcome=success,failed) that no tasks were executed as they were all up to date.


integration/build.gradle.kts
```diff
@@ -40,20 +40,12 @@ tasks {
 
     run.configure {
         environment("TF_CPP_MIN_LOG_LEVEL" to "1") // turn off TensorFlow print out
-        // @Niels Doucet
-        // Just a heads-up: gradle support warned me about systemProperties System.getProperties(). It's really
-        // dangerous to just copy over all system properties to a task invocation. You should really be specific about
-        // the properties you'd like to expose inside the task, or you might get very strange issues.
-        systemProperties = System.getProperties().toMap() as Map<String, Any>
```
frankfliu, line 47: What's the right way to pass system properties in command line when run gradle?

```
gradle run -Dai.djl.offline=true
```

There are many DJL specific system properties, it's hard for use the list them

frankfliu (approved):

integration/build.gradle.kts
benjie332: If they are all namespaced like that example, you can copy over the properties that start with `ai.djl.offline` instead of blanket copying over all System properties. That should be much safer overall, and also avoid copying properties which change frequently and 'break' the incremental builds.

engines/llama/build.gradle.kts
```diff
@@ -49,15 +51,15 @@ tasks {
             }
             copy {
                 from(jnilibDir)
-                into(project.projectDir / "build/classes/java/main/native/lib")
+                into("$path/native/lib")
             }
 
             // write properties
-            val propFile = project.projectDir / "build/classes/java/main/native/lib/llama.properties"
+            val propFile = file("$path/native/lib/llama.properties")
```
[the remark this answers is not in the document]
benjie332: Nothing as far as I know, just my personal preference 😄 `File` would be the constructor, and `file()` would be the option provided by the Gradle DSL

benjie332:
Thanks for the merge!

PR #3243, opened 2024-06-06
[enhance] add ByteBufferAdapter to construct NDArray

ewan0x79:
## Description ##

This PR introduces an optimization for constructing model input `NDArray` within the DJL framework. Previously, for two-dimensional model inputs, we would utilize `ai.djl.ndarray.NDManager#create(float[][])` to build the `NDArray`. However, this process involved copying the data into a temporary array, then from non-direct memory to direct memory, which introduced unnecessary overhead.

To address this, we have introduced `ByteBufferAdapter`. This allows users to directly write their data into direct memory, which can then be used by the underlying engine to construct the tensor. This approach eliminates the need for two rounds of data copying, thereby reducing overhead.

For memory safety, the direct memory utilized will be managed by `NDManager`, ensuring that resources are appropriately released. Additionally, a `clean` method is provided, giving users the option to manage memory manually if they choose to do so.

### Key Points ###
- Introduction of `ByteBufferAdapter` for efficient data handling.
- Elimination of unnecessary data copying, reducing overhead.
- Memory safety ensured through `NDManager`.
- Optional manual memory management with the `clean` method.

This optimization is crucial for performance-sensitive applications and represents a significant improvement in the efficiency of data handling within the DJL framework.

### DIFF
--- a/api/src/main/java/ai/djl/translate/ByteBufferAdapter.java
+++ b/api/src/main/java/ai/djl/translate/ByteBufferAdapter.java
@@ -1,0 +1,600 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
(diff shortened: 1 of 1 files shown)


ewan0x79:
@frankfliu 
Indeed, as you mentioned, the data is not additionally copied into non-direct memory but rather the array is wrapped as non-direct memory. Initially, my thought was that the input might not be of array type, possibly another type such as Row in Spark. In such cases, when manually stacking a batch of data, it would be necessary to construct a temporary two-dimensional array, copying certain data from the Row into the array, and then using this array to build an NDArray. I assumed that directly copying the data into direct memory and then using this direct memory to construct an NDArray could reduce the overhead of copying. However, to my surprise, in my actual tests, copying the data into a temporary array and then using this temporary array to build an NDArray turned out to be slightly faster. Well, this indeed turned out to be a negative optimization.

PR #3244, opened 2024-06-06
[api] Adds metric type to Metic class

frankfliu:
## Description ##


api/src/main/java/ai/djl/metric/Metric.java
```diff
@@ -218,6 +253,9 @@ public String toString() {
         if (timestamp != null) {
             sb.append('|').append(timestamp);
         }
+        if (metricType != null) {
```
zachgk, line 256: How does this work if you have a metricType but no timestamp?

frankfliu: It should work

zachgk (approved):

PR #3273, opened 2024-06-24
[pytorch] Split maven publish into two parts

frankfliu:
The pytorch native jar file become two large, and sonatype auto staging break the artifacts into multiple staging repo and can not close properly.

### DIFF
--- a/.github/workflows/native_publish_pytorch.yml
+++ b/.github/workflows/native_publish_pytorch.yml
@@ -7,6 +7,10 @@ on:
           description: 'staging/snapshot, default is snapshot'
           required: true
           default: 'snapshot'
+        package_type:
+          description: 'cpu/gpu, default is gpu'
+          required: true
+          default: 'cpu'
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/native_publish_pytorch.yml
```diff
@@ -7,6 +7,10 @@ on:
           description: 'staging/snapshot, default is snapshot'
           required: true
           default: 'snapshot'
+        package_type:
```
zachgk, line 10: Rather than making this an input, it might be better to run as a matrix and have both package builds in one run?

.github/workflows/native_publish_pytorch.yml
```diff
@@ -27,7 +31,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
       - name: Download native binaries
-        run: ./gradlew :engines:pytorch:pytorch-native:dPTNL
+        run: ./gradlew :engines:pytorch:pytorch-native:dPTNL -Ppackage_type=${{ github.event.inputs.package_type }}
```
zachgk, line 34:
```suggestion
        run: ./gradlew :engines:pytorch:pytorch-native:dPTNL -Ppackage_type=${{ inputs.package_type }}
```

zachgk (approved):
Left some optional comments. Overall LGTM

.github/workflows/native_publish_pytorch.yml
frankfliu: The problem is on sonatype side, in gradle we are publishing each package sequentially, ideally, each package should landed in the same repo or each with its own repo. We don't have control the close the repo for each package. Even we manually run two publish command, it may or may not in the same repo. Currently, I manually publish one, login to sonatype, close the repo and run another publish.

PR #3283, opened 2024-06-27
[tensorflow] Updates tensorflow to 2.16.1

frankfliu:
## Description ##


.github/workflows/nightly_publish.yml
```diff
@@ -133,15 +133,15 @@ jobs:
     if: github.repository == 'deepjavalibrary/djl'
     runs-on: [ self-hosted, gpu ]
     container:
-      image: nvidia/cuda:12.1.1-cudnn8-runtime-ubuntu20.04
+      image: nvidia/cuda:12.4.1-cudnn-devel-ubuntu22.04
```
siddvenk, line 136: do we need the development image?

siddvenk (approved):

.github/workflows/nightly_publish.yml
frankfliu: Yes, TF need nvcc. If the TF is not compiled with cuda arch, it will compile on the fly.

PR #3312, opened 2024-07-10
[llamacpp] Removing llamacpp support in DJL

frankfliu:
## Description ##


vaiju1981:
Just trying to understand why we are removing llamacpp engine? is there a replacement planned to support LLM's via DJL?

frankfliu:
> Just trying to understand why we are removing llamacpp engine? is there a replacement planned to support LLM's via DJL?

1. We don't have maintainer can actively maintain this module. 
2. We we focused on supporting LLM via Python engine in DJLServing repo
3. We working on rust engine and will support rust based LLM support in future.

xyang16 (approved):

PR #3317, opened 2024-07-10
[tokenizers] Fixes memory leak when there is overflowing tokens

### LINKED ISSUE #3316: Memory leak in HF tokenizer when using truncation and optWithOverflowingTokens(false)
We have observed a native memory leak when using `ai.djl.huggingface.tokenizers.HuggingFaceTokenizer`. Using the default (`false`) option of `optWithOverflowingTokens` we see a significant increase in memory usage over time when we have long strings that are truncated to shorter token sequences. When we set `optWithOverFlowingTokens` to `true`, we do not see this memory increase. This is particularly evident when very long strings are truncated to short token sequences.

Testing back, this behaviour started in version 0.27.0, and tracing that to the release notes, it seems that this PR might be the culprit: https://github.com/deepjavalibrary/djl/pull/2957.

Particularly: https://github.com/deepjavalibrary/djl/pull/2957/files#diff-62d10f278a5a7644ce30deff638cf6ead21457bca60b9cc7430d115dd2fa2b38R533-R537

It seems that by calling `TokenizersLibrary.LIB.getOverflowing(encoding)`, a clone will be created that is only cleaned up when `withOverflowTokens` is `true`, as `toEncoding` is then called recursively on the overflowing handles which eventually calls `TokenizersLibrary.LIB.deleteEncoding(encoding);` on this copy. 

So when `withOverflowTokens` is `false`, this cleanup does not occur.

baldersheim:
Fixes #3316

If you call TokenizersLibrary.LIB.getOverflowing you must also clean up all overflow encodings.

If withOverflowingTokens was false no Encodings where generated leaving jni Encoding handles that would not be properly deleted.

This introduces a new native method where you can inquire about number of overflow tokens without using any jni resources. And you will now only call TokenizersLibrary.LIB.getOverflowing(encoding) if withOverflowingTokens is true.

## Description ##

### DIFF
--- a/extensions/tokenizers/rust/src/lib.rs
+++ b/extensions/tokenizers/rust/src/lib.rs
@@ -498,6 +498,17 @@ pub extern "system" fn Java_ai_djl_huggingface_tokenizers_jni_TokenizersLibrary_
 }
 
 #[no_mangle]
+pub extern "system" fn Java_ai_djl_huggingface_tokenizers_jni_TokenizersLibrary_getOverflowCount(
+    _: JNIEnv,
+    _: JObject,
+    handle: jlong,
+) -> jint {
+    let encoding = cast_handle::<Encoding>(handle);
+    let count = encoding.get_overflowing().len();
+    count as jint
+}
+
+#[no_mangle]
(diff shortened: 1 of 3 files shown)


baldersheim:
@frankfliu This PR suggests a solution to issue #3316

frankfliu (approved):

PR #3326, opened 2024-07-11
[tokenizers] Download jni lib files for cuda

xyang16:
## Description ##


extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/jni/LibUtils.java
```diff
@@ -114,4 +134,38 @@ private static Path copyJniLibraryFromClasspath(String[] libs) {
...
+                        + version
+                        + "/jnilib/"
+                        + djlVersion
+                        + '/'
+                        + classifier
+                        + '/'
+                        + flavor
+                        + '/'
+                        + LIB_NAME;
+        logger.info("Downloading jni {} to cache ...", url);
```
frankfliu, line 156: 1. We don't have cuda for windows
2. We may only want to support cuda12

xyang16: Changed.

frankfliu (approved):

PR #3336, opened 2024-07-12
[rust] Change loging level to debug

xyang16:
## Description ##

### DIFF
--- a/extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/jni/LibUtils.java
+++ b/extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/jni/LibUtils.java
@@ -158,6 +158,10 @@ public final class LibUtils {
 
             for (String libName : libs) {
                 String libPath = "native/lib/" + classifier + "/" + flavor + "/" + libName;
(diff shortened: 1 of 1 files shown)


extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/jni/LibUtils.java
```diff
@@ -167,7 +167,7 @@ private static boolean copyJniLibraryFromClasspath(
             Utils.moveQuietly(tmp, dir);
             return true;
         } catch (IOException e) {
-            logger.error("Cannot copy jni files", e);
+            logger.debug("Cannot copy jni files: {}", e.getMessage());
```
frankfliu, line 170: 1. We should check null for cu122 case at line 162, it will throw and NPE.
2. We should keep it as warning at least. IOException should not happen.

xyang16: IOException will happen if it's cu122, it will need to download.

frankfliu (approved):

PR #3344, opened 2024-07-17
[release] Bump up versions to 0.29.0 in documents to point to new url

ydm-amazon:
[MCM 0.29.0] Bump up versions in documents to point to new url

Notebooks CI will fail because there is no version in sonatype - re-run docs CI and merge it after release in sonatype is complete.

### DIFF
--- a/engines/pytorch/pytorch-engine/README.md
+++ b/engines/pytorch/pytorch-engine/README.md
@@ -46,6 +46,7 @@ The following table illustrates which pytorch version that DJL supports:
 
 | PyTorch engine version | PyTorch native library version            |
 |------------------------|-------------------------------------------|
+| pytorch-engine:0.29.0  | 1.13.1, 2.1.2, 2.2.2, **2.3.1**           |
 | pytorch-engine:0.28.0  | 1.13.1, 2.1.2, **2.2.2**                  |
 | pytorch-engine:0.27.0  | 1.13.1, **2.1.1**                         |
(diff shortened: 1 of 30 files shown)


siddvenk:
I think we need to update the pytorch version first in the docs.

I'm seeing usage of pytorch 2.2.2, but pretty sure we're going to release 2.3.1 as the default version here for 0.29.0. @frankfliu is that accurate?

frankfliu:
> I think we need to update the pytorch version first in the docs.
> 
> I'm seeing usage of pytorch 2.2.2, but pretty sure we're going to release 2.3.1 as the default version here for 0.29.0. @frankfliu is that accurate?

We need manually update this README

engines/pytorch/pytorch-engine/README.md
```diff
@@ -115,7 +115,7 @@ export PYTORCH_FLAVOR=cpu
 ### macOS
 For macOS, you can use the following library:
 
-- ai.djl.pytorch:pytorch-jni:2.2.2-0.28.0
+- ai.djl.pytorch:pytorch-jni:2.2.2-0.29.0
```
siddvenk, line 118: here and subsequently - should this be 2.3.1-0.29.0?

frankfliu (approved):

siddvenk (approved):

engines/pytorch/pytorch-engine/README.md
ydm-amazon: Nice catch, fixed

PR #3365, opened 2024-07-26
Update docs.yml and docs workflow to support versioning

Varun-Dutta:
Chages to mkdocs.yml

1. Added the parameter "site_dir" as required by mike 
2. Added mike to the required plugins 

Changes to docs workflow

1. Added five parameters to support versioning 
2. Added caching to reduce set-up time for runner 
3. Added authentication using aws-actions/configure-aws-credentials@v4 to access S3 bucket in Gamma account 
4. Added to step to checkout previous SHAs when deploying new versions 
5. Use mike instead of mkdocs to build new website
6. Add option to upload new website to Gamma S3 Bucket 

Testing:

1. Downloaded the updated website from Gamma S3 Bucket to local directory. Ran the website locally.


.github/workflows/docs.yml
```diff
@@ -8,61 +8,231 @@ on:
...
+        type: boolean
+        description: Is this a test run that will be built into gamma account?
+        default: false
+        required: false
+      is-new-version:
+        type: boolean
+        description: Are you deploying new version? If so, enter the new version number (ex. 28.0). If not, skip the rest of the parameters. 
+        default: false
+        required: true
+      version-number:
```
frankfliu, line 22: Can we infer version from tag?

.github/workflows/docs.yml
```diff
@@ -8,61 +8,231 @@ on:
...
+        type: string 
+        description: What is the new version number for the website?
+        default: dev
+        required: false
+      djl-tag:
```
frankfliu, line 27: better use one tag for all repos

.github/workflows/docs.yml
```diff
@@ -8,61 +8,231 @@ on:
...
   documentation:
-    if: github.repository == 'deepjavalibrary/djl'
+    # Change line below 
+    if: github.repository == 'Varun-Dutta/djl'  
     runs-on: ubuntu-latest
     steps:
+        
+    #Phase I: Set-up
+    #This phase sets up all of the necessary dependencies. This includes authenticating the workflow, installing Java and Python, and adding the tools used to build the documentation website.  
+      - name: Configure AWS Test Credentials
```
frankfliu, line 55: This should be in later steps to avoid credential accidentally leaked to build process

.github/workflows/docs.yml
```diff
@@ -8,61 +8,231 @@ on:
...
+      - name: Delete index.html 
         run: |
-          python3 tools/scripts/add_online_runner.py
-      - name: clone demos
+          rm -f index.html    
+          git add -u 
+          git commit -m "Removed index.html" || echo "No changes to commit"
+        shell: bash
+
+      - name: Clone DJL-Serving
```
frankfliu, line 130: Can we use checkout actions?

.github/workflows/docs.yml
```diff
@@ -8,61 +8,231 @@ on:
...
+
+        #This step is necessary because the index.html file in djl/index.html overwrites the index.html file Mike create for the landing page
+        #This is problematic because this index.html corresponds to the djl website rather than the docs website.
+      - name: Delete index.html 
         run: |
-          python3 tools/scripts/add_online_runner.py
-      - name: clone demos
+          rm -f index.html    
+          git add -u 
+          git commit -m "Removed index.html" || echo "No changes to commit"
```
frankfliu, line 127: Why do we nee commit to git locally?

.github/workflows/docs.yml
```diff
@@ -8,61 +8,231 @@ on:
...
+        
       - name: run Notebooks
         run: |
           cd docs/demos/jupyter
           bash test_notebook.sh
+
+      #Phase III: Retrive Current Website
+      #This phase retrieves the current version of the website that is stored in S3. The key pieces are the index.html and versions.json which serve as the landing page and the tracker of previous versions. 
+      #Potential Improvement: See if mike still works if only downloading the index.html file and versions.json rather than the entire previous website. 
+      - name: Sync gh-pages
```
frankfliu, line 169: Can we just create a shell script for all these steps?, ideally this script can be executed locally

.github/workflows/docs.yml
Varun-Dutta: Implemented this change.

.github/workflows/docs.yml
Varun-Dutta: We can, I updated the workflow to use actions/checkout.

.github/workflows/docs.yml
Varun-Dutta: We don't, I removed the commit.

.github/workflows/docs.yml
Varun-Dutta: Implemented this change.

.github/workflows/docs.yml
Varun-Dutta: I updated the workflow to infer this.

.github/workflows/docs.yml
```diff
@@ -75,22 +71,22 @@ jobs:
...
-
       - name: Configure Deployment AWS Credentials
         uses: aws-actions/configure-aws-credentials@v4
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-2
+
+      - name: Build docs website
+        run: tools/scripts/build-website.sh "${{ github.event.inputs.version-number || 'master' }}"
```
frankfliu, line 82: build-website doesn't need aws credentials, why move to here?

Varun-Dutta: I am working on section in build-website that would download the previous versions.json and then prepend the new version info to it. I moved the aws credentials up to download the versions.json file

tools/scripts/build-website.sh
```diff
@@ -0,0 +1,31 @@
...
+# Configure Git User, mike will create gh-pages branch
+git config user.name "nobody"
+git config user.email "nobody@localhost"
+
+echo "Building docs for $VERSION_NUMBER"
+mike deploy "$VERSION_NUMBER" -F docs/mkdocs.yml -b gh-pages
+git checkout gh-pages
+
+echo "Attempting to update versions.json"
+aws s3 cp s3://djl-ai/documentation/nightly/versions.json .
```
frankfliu, line 24: 1. Can we just generate this `versions.json` file without download it?
2. We don't need use awscli to download it, we can use curl:
```
curl -O https://docs.djl.ai/versions.json
```

Varun-Dutta: Probably, I'll look into using curl, I am unfamiliar with it

frankfliu (approved):

PR #3367, opened 2024-07-29
[rust] Add reranker and text classification models

xyang16:
## Description ##

### DIFF
--- a/extensions/tokenizers/rust/src/layers/mod.rs
+++ b/extensions/tokenizers/rust/src/layers/mod.rs
@@ -5,6 +5,7 @@ mod linear;
 #[allow(dead_code, unused)]
 mod rms_norm;
 
+#[allow(unused_imports)]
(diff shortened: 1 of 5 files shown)


extensions/tokenizers/rust/src/models/bert.rs
```diff
@@ -63,57 +65,8 @@ pub struct BertConfig {
     #[serde(default)]
     use_cache: bool,
     classifier_dropout: Option<f64>,
+    pub id2label: Option<HashMap<String, String>>,
     pub use_flash_attn: Option<bool>,
-    model_type: Option<String>,
-}
-
-impl Default for BertConfig {
-    fn default() -> Self {
```
frankfliu, line 71: So we don't need all these default value any more?

xyang16: Added back.

frankfliu (approved):

PR #3382, opened 2024-08-02
[tokenizer] Make normalize default to false

xyang16:
## Description ##

Make normalize default to false. normalize take a lot of time, we can let use enable if they want.


extensions/tokenizers/src/main/java/ai/djl/huggingface/translator/TextEmbeddingTranslator.java
```diff
@@ -280,7 +280,7 @@ public static final class Builder {
 
         private HuggingFaceTokenizer tokenizer;
         private Batchifier batchifier = Batchifier.STACK;
-        private boolean normalize = true;
+        private boolean normalize;
```
frankfliu, line 283: We should change the value in djl-convert for rust, not default value here

PR #3385, opened 2024-08-05
[docs] Updates djl model converter document

frankfliu:
## Description ##

### DIFF
--- a/extensions/tokenizers/README.md
+++ b/extensions/tokenizers/README.md
@@ -13,6 +13,7 @@ You can also build the latest javadocs locally using the following command:
 ```sh
 ./gradlew javadoc
 ```
+
 The javadocs output is built in the `build/doc/javadoc` folder.
 
 ## Installation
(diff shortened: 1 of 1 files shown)


extensions/tokenizers/README.md
```diff
@@ -29,71 +31,95 @@ You can pull the module from the central Maven repository by including the follo
...
 
-# import models as DJL Model Zoo
-djl-import --help
-```
+#### Load converted model
 
 Then, all you need to do, is to load this model in DJL:
 
 ```java
 Criteria<QAInput, String> criteria = Criteria.builder()
```
lanking520, line 102: could we provide a link to this sample Java code?

frankfliu: We don't have example like this, we might need create on in djl-demo repo

lanking520: ok, will take an action

lanking520 (approved):

PR #3412, opened 2024-08-14
[rust] Add mistral embedding model

xyang16:
## Description ##


extensions/tokenizers/rust/src/layers/linear.rs
```diff
@@ -27,9 +27,14 @@ impl Linear {
         out_dim: usize,
         act: Option<HiddenAct>,
     ) -> Result<Self> {
+        let bias = match vb.get(out_dim, "bias")
```
frankfliu, line 30: reformat your code with:

```
cargo fmt
```

xyang16: Done

frankfliu (approved):

PR #3422, opened 2024-08-17
[rust] Add GTE and Gemma2 model

xyang16:
## Description ##

### DIFF
--- a/extensions/tokenizers/src/main/python/djl_converter/huggingface_converter.py
+++ b/extensions/tokenizers/src/main/python/djl_converter/huggingface_converter.py
@@ -119,7 +119,8 @@ class HuggingfaceConverter:
         if hasattr(config, "model_type"):
             if config.model_type not in [
                     "bert", "camembert", "distilbert", "xlm-roberta",
-                    "roberta", "nomic_bert", "mistral", "qwen2"
+                    "roberta", "nomic_bert", "mistral", "qwen2", "new",
+                    "gemma2"
             ]:
                 return False, f"Unsupported model_type: {config.model_type}", -1
         else:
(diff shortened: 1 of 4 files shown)


extensions/tokenizers/src/main/python/djl_converter/huggingface_converter.py
```diff
@@ -119,7 +119,8 @@ def save_rust_model(self, model_info, args: Namespace, temp_dir: str,
         if hasattr(config, "model_type"):
             if config.model_type not in [
                     "bert", "camembert", "distilbert", "xlm-roberta",
-                    "roberta", "nomic_bert", "mistral", "qwen2"
+                    "roberta", "nomic_bert", "mistral", "qwen2", "new",
```
frankfliu, line 122: Add a TODO here, we need monitor if Transformer will give a better model type in few, or if there is conflict.

extensions/tokenizers/src/main/python/djl_converter/huggingface_converter.py
```diff
@@ -119,7 +119,7 @@ def save_rust_model(self, model_info, args: Namespace, temp_dir: str,
         if hasattr(config, "model_type"):
             if config.model_type not in [
                     "bert", "camembert", "distilbert", "xlm-roberta",
-                    "roberta", "nomic_bert", "mistral", "qwen2"
+                    "roberta", "nomic_bert", "mistral", "qwen2", "new"
```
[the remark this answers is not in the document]
xyang16: https://huggingface.co/Alibaba-NLP/gte-large-en-v1.5/blob/main/config.json

GTE model can be detected either "new" model type or "NewModel" architecture.

extensions/tokenizers/src/main/python/djl_converter/huggingface_converter.py
xyang16: Added

frankfliu (approved):

PR #3434, opened 2024-08-21
[rust] Make cublaslt wrapper non static

xyang16:
## Description ##

Make cublaslt wrapper non static to fix illegal memory access.

### DIFF
--- a/extensions/tokenizers/rust/src/layers/cublaslt.rs
+++ b/extensions/tokenizers/rust/src/layers/cublaslt.rs
@@ -5,29 +5,23 @@ use std::sync::Once;
 #[cfg(feature = "cuda")]
 use candle_cublaslt::{fused_batch_matmul, fused_matmul, Activation, CublasLt};
 
-static INIT: Once = Once::new();
-static mut CUBLASLT: Option<CublasLtWrapper> = None;
-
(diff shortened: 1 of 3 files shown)


extensions/tokenizers/rust/src/layers/cublaslt.rs
```diff
@@ -5,29 +5,23 @@ use std::sync::Once;
 #[cfg(feature = "cuda")]
 use candle_cublaslt::{fused_batch_matmul, fused_matmul, Activation, CublasLt};
 
-static INIT: Once = Once::new();
-static mut CUBLASLT: Option<CublasLtWrapper> = None;
-
-pub fn get_cublas_lt_wrapper() -> Option<&'static CublasLtWrapper> {
+pub fn get_cublas_lt_wrapper(device: &Device) -> Option<CublasLtWrapper> {
```
frankfliu, line 8: Why static initialization cause issue?

xyang16: Because static cublaslt is created on device 0, if device 1 use it, it will cause illegal memory access. This is causing the integration test failure now.

frankfliu (approved):

PR #3476, opened 2024-09-16
[tokenizers] Use tokenizers from rust.io

xyang16:
## Description ##


siddvenk (approved):

siddvenk:
was there a reason we weren't doing this before?

xyang16:
I think because tokenizers was not available in rust.io.

frankfliu (approved):

PR #3478, opened 2024-09-17
[ci] Fixes fasttext native build for nodejs20 issue

xyang16:
## Description ##

### DIFF
--- a/.github/workflows/native_s3_fasttext.yml
+++ b/.github/workflows/native_s3_fasttext.yml
@@ -10,27 +10,20 @@ permissions:
 jobs:
   build-fasttext-jni-linux:
     runs-on: ubuntu-latest
-    container: amazonlinux:2
+    container:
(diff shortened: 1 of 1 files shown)


.github/workflows/native_s3_fasttext.yml
```diff
@@ -10,14 +10,14 @@ permissions:
 jobs:
   build-fasttext-jni-linux:
     runs-on: ubuntu-latest
-    container: amazonlinux:2
+    container: amazonlinux:2023
```
frankfliu, line 13: We cannot upgrade to amazonlinux:2023 yet

frankfliu: All internal AWS customers are using AL2

xyang16: OK. Changed the checkout action.

PR #3543, opened 2024-11-25
[ci] fix ci to test windows properly

siddvenk:
## Description ##

The continuous workflow for windows is busted, and we've had a silent failure in that test for a long time. 

The issue is that we run 3 separate tests in the final github step of continuous/build-windows, but the success/failure outcome is only dependent on the final command we run. Specifically, we run:
```
./gradlew -Pjni "-Dai.djl.default_engine=PyTorch" test -x examples:test
./gradlew -Pjni --rerun-tasks "-Dai.djl.default_engine=TensorFlow" test -x examples:test
./gradlew --rerun-tasks "-Dai.djl.default_engine=OnnxRuntime" :integration:test
```

When this runs if the 3rd line succeeds, it hides any failure that occurs in the previous two lines.

This change refactors the continuous workflow by unifying the windows test into the existing mac/linux tests, which function as expected. The separate stage for windows may have made sense sometime in the past, it no longer needs to be separate.

I have also modified the nightly-publish workflow to ensure that we test all 3 major OS's across each of the suites. I have also specified fail-fast: false to ensure no short-circuiting of tests due to failures.

### DIFF
--- a/.github/workflows/continuous.yml
+++ b/.github/workflows/continuous.yml
@@ -25,8 +25,9 @@ jobs:
     if: github.repository == 'deepjavalibrary/djl'
     runs-on: ${{ matrix.operating-system }}
     strategy:
+      fail-fast: false
       matrix:
-        operating-system: [ ubuntu-latest, macos-latest ]
+        operating-system: [ ubuntu-latest, macos-latest, windows-latest ]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


siddvenk:
Note:

The continuous/build (windows-latest) is expected to fail as a result of this change. We are fixing the workflow so that failures like this are exposed. I will be raising another PR shortly that will address that issue. I want to merge this first so that we can see that the change I make actually resolves the issue.

Same case for the nightly-publish. I have the run based on the changes made here available at https://github.com/deepjavalibrary/djl/actions/runs/12020282449

Issue #2949, opened 2024-01-18
Make commons-compress an optional dependency


StefanOltmann:
I'm experiencing difficulties with ProGuard and commons-compress. This has led me to consider the possibility of excluding the commons-compress dependency, allowing the application to function smoothly unless a zipped model that requires unzipping is explicitly specified. This approach would enable me to circumvent ProGuard issues associated with commons-compress while reducing unnecessary dependencies.

frankfliu:
Can you exclude `commons-compress` from your project if you are not using any `.tar` file?

This should be pretty easy from gradle:

```
    implementation platform("ai.djl:bom:0.26.0")
    implementation("ai.djl:api") {
          exclude group: "org.apache.commons", module: "commons-compress"
    }
```

StefanOltmann:
@frankfliu Do you have an estimate when the next release including this fix will be available?

I see you seem to have monthly releases.

frankfliu:
We just released 0.26.0, 0.27.0 will be in March, see: https://docs.djl.ai/index.html#release-notes

For the meantime, you can use our nightly snapshot release: https://docs.djl.ai/docs/get.html#nightly-snapshots

StefanOltmann:
@frankfliu I've tested the latest snapshot version and found that it functions smoothly with ProGuard. There seems to be something off with commons-compress that hinders optimization. I'm glad I don't need that library as long as I'm not using TAR files. Thank you for quickly fixing this!

Considering March is a considerable time away, I was wondering if you could consider releasing a `v0.26.1` fix version. I'm a bit hesitant to deploy a snapshot version in a production environment, as it might introduce breaking changes at any moment.

StefanOltmann:
@zachgk Thank you for showing me a workaround. I guess I need to try that then.

I still must say I’m a bit surprised that you have so long and fixes release schedules and that it’s such an effort to do a release. Sounds to me like something is off.

Maybe you should consider reworking your CI pipeline for more automation. I maintain two libraries with Maven Central artifacts and releasing a new one is just a matter creating a GitHub release (which builds everything using GitHub actions and uploads to Maven Central) and confirming in the Sonatype Nexus that the artifacts should be released to Maven Central.

frankfliu:
@StefanOltmann 
Releasing DJL is non-trivial work. We have 11 engines and need to test on to following platforms:
- macOS
- macOS M1
- ubuntu
    - x86_64
    - aarch64
[...]

PR #3544, opened 2024-11-26
[api] fix issue in Tar/Zip Utils that resulted in incorrect artifact …

siddvenk:
…extraction

## Description ##

This change fixes an issue with `ZipUtils.unzip`, and `TarUtils.untar`. This issue was obfuscated by the fact that our ci tests were not failing when they should have been. See https://github.com/deepjavalibrary/djl/pull/3543 for details.

Rather than stripping leading file separators and still extracting the archive, this change validates that the tar entry will extract into the expected output directory. 

While this change means that some tars that previously worked (such as those with entries starting with `/` or `\`) will no longer work, I argue that the new behavior is correct and those tar entries are invalid/incorrect.


api/src/main/java/ai/djl/util/ZipUtils.java
```diff
@@ -121,14 +118,16 @@ private static void addToZip(Path root, Path file, ZipOutputStream zos) throws I
...
+        Path expectedOutputPath = destination.resolve(name).normalize();
+        if (!expectedOutputPath.startsWith(destination.normalize())) {
+            throw new IOException(
+                    "Bad archive entry "
+                            + name
+                            + ". Attempted write outside destination "
+                            + destination);
         }
-        return name.substring(index);
+        return name;
```
si2d, line 130: Why return the input variable as is? Maybe this method should be `getAbsolutePath` and the method is expected to throw if the path is invalid. 
We are using the output in `vis.validate`. Do we need that after this change?

api/src/test/java/ai/djl/util/ZipUtilsTest.java
```diff
@@ -49,13 +51,20 @@ public void testEmptyZipFile() throws IOException {
...
-            TarUtils.untar(is, output, false);
+            Assert.assertThrows(() -> TarUtils.untar(is, output, false));
+        }
+    }
+
+    @Test
+    public void testLinuxCreatedWindowsUsedOffendingTar() throws IOException {
+        TestRequirements.windows();
+        Path tarPath = Paths.get("src/test/resources/linux_create_windows_use.tar");
+        Path output = Paths.get("C:/out");
```
frankfliu, line 64: Why choose "C:/out"?, better output to `build` folder, although the test throws exception before access to file system but using "c:/out" is a bit confusion:
1. not all machine has c driver
2. not all environment has write access to /out directory

api/src/main/java/ai/djl/util/ZipUtils.java
frankfliu: Agree, this method can be `void`

siddvenk:
@si2d @frankfliu what do you think about this method? We would use `sanitizeAndValidateArchiveEntry` where we currently use `validateArchiveEntry`.

1. We sanitize the archive entry by converting the entry name to a system aware path (convert a linux archive entry path to the equivalent windows path, or vis versa). Then we can remove the leading file separator as before (but it works now cross platform) and handles the concern @frankfliu mentioned about `/` being a valid starting char for zip
2. Validate that with this santized path, the expected output path exists under the expected output dir


```
static String sanitizeAndValidateArchiveEntry(String name, Path destination) throws IOException {
        String sanitizedName = removeLeadingFileSeparator(name);
        Path expectedOutputPath = destination.resolve(sanitizedName).normalize();
        if (!expectedOutputPath.startsWith(destination.normalize())) {
            throw new IOException(
                    "Bad archive entry "
                            + name
                            + ". Attempted write outside destination "
                            + destination);
        }
        return name;
    }

    static String removeLeadingFileSeparator(String name) {
        String osAwareArchiveEntryName = FilenameUtils.separatorsToSystem(name);
        int index = 0;
        for (; index < name.length(); index++) {
            if (name.charAt(index) != File.separatorChar) {
                break;
            }
        }
        return osAwareArchiveEntryName.substring(index);
    }
```

api/src/main/java/ai/djl/util/ZipUtils.java
siddvenk: I think we still need `vis.validate` since that method is doing some deeper validation specific to zips. It's validating the entries in the header match what is in the archive

frankfliu:
@si2d 

in archive file, the path is always linux style, we don't really need use os specific Path to validate it. And we don't have to support special cases (even they are valid). We can just check if the entry name starts with "/" or contains "..",  we treat them as invalid. This was the original algorithm. I'm curious why that cause test failure on windows.

siddvenk:
> @si2d
> 
> in archive file, the path is always linux style, we don't really need use os specific Path to validate it. And we don't have to support special cases (even they are valid). We can just check if the entry name starts with "/" or contains "..", we treat them as invalid. This was the original algorithm. I'm curious why that cause test failure on windows.

File.separatorChar is os specific. On mac/linux it is `/`, on windows it is `\`. So on windows, if you had an archive entry like `/tmp/test.txt`, the removeLeadingFileSeparator would not actually remove `/` from the entry and cause the write at root

siddvenk:
@frankfliu with the existing logic (ignoring any of my changes here), this is what the testOffendingTar unit test produces on windows (i added some print statements, but otherwise logic is same).

```
Gradle suite > Gradle test > ai.djl.util.ZipUtilsTest > testOffendingTar STANDARD_OUT
    Entry before removing leading separator: /tmp/empty.txt
    Entry after removing leading separator: /tmp/empty.txt
    Writing archive entry to D:\tmp\empty.txt
```

The output dir in that test is `build/output`, and we would expect that we write to `build/output/tmp/empty.txt`, but that's not what happens because of the separatorChar issue i mentioned

si2d:
> For Zip file, entry name starts with "/" is valid, and will be ignored by common zip utils. For Tar file, entry name starts with "/" is a security vulnerability, `tar` cli will untar to root folder.
> 
> I think it's OK we make it more strict that doesn't allows "/", but your changes makes `a/../b.txt` valid, which can overwrite `b.txt` in the same archive file.

I wonder - instead of trying to block this, should we only disallow overwrites? so even if the path is a/../b.txt, it will be allowed as long as nothing else writes to b.txt. We are explicitly setting REPLACE_EXISTING when unarchiving, but is overwriting files a valid use case? (Also, this change might break customers if they are somehow using it)

siddvenk:
The simplest change that solves the issue is to keep everything the same, except for this modification to `removeLeadingFileSeparator`.

```
static String removeLeadingFileSeparator(String name) {
        // Below single line is the only change
        String osAwareArchiveEntryName = FilenameUtils.separatorsToSystem(name);
        int index = 0;
        for (; index < osAwareArchiveEntryName.length(); index++) {
            if (osAwareArchiveEntryName.charAt(index) != File.separatorChar) {
                break;
            }
        }
        return osAwareArchiveEntryName.substring(index);
    }
```

frankfliu:
> The simplest change that solves the issue is to keep everything the same, except for this modification to `removeLeadingFileSeparator`.
> 
> ```
> static String removeLeadingFileSeparator(String name) {
>         // Below single line is the only change
>         String osAwareArchiveEntryName = FilenameUtils.separatorsToSystem(name);
>         int index = 0;
>         for (; index < osAwareArchiveEntryName.length(); index++) {
>             if (osAwareArchiveEntryName.charAt(index) != File.separatorChar) {
>                 break;
>             }
>         }
>         return osAwareArchiveEntryName.substring(index);
>     }
> ```

Right, we should not use File.separatorChar. Archive file format is not os specific

frankfliu:
`vis.validate(set)` is till required. There is a CVE. Basically, when user use winzip to open the file, the see: `a.txt`, but when they unzip to file system, they got `b.txt`. We need make sure the ZipEntry matches the ZipInputStream name.

siddvenk:
> > The simplest change that solves the issue is to keep everything the same, except for this modification to `removeLeadingFileSeparator`.
> > ```
> > static String removeLeadingFileSeparator(String name) {
> >         // Below single line is the only change
> >         String osAwareArchiveEntryName = FilenameUtils.separatorsToSystem(name);
> >         int index = 0;
> >         for (; index < osAwareArchiveEntryName.length(); index++) {
> >             if (osAwareArchiveEntryName.charAt(index) != File.separatorChar) {
> >                 break;
> >             }
> >         }
> >         return osAwareArchiveEntryName.substring(index);
> >     }
> > ```
> 
> Right, we should not use File.separatorChar. Archive file format is not os specific

The original algorithm was using File.separatorChar, not `/`. I think it's still possible to craft a tar with `\` as the path separator. Converting the entry name with separatorsToSystem, and then using File.separatorChar will work. We could also modify the condition to something like
```
if (name.charAt(index) != '/' && name.charAt(index) != '\\') {
 ...
}
```

si2d:
> > > The simplest change that solves the issue is to keep everything the same, except for this modification to `removeLeadingFileSeparator`.
> > > ```
> > > static String removeLeadingFileSeparator(String name) {
> > >         // Below single line is the only change
> > >         String osAwareArchiveEntryName = FilenameUtils.separatorsToSystem(name);
> > >         int index = 0;
> > >         for (; index < osAwareArchiveEntryName.length(); index++) {
> > >             if (osAwareArchiveEntryName.charAt(index) != File.separatorChar) {
> > >                 break;
> > >             }
> > >         }
> > >         return osAwareArchiveEntryName.substring(index);
> > >     }
> > > ```
> > 
> > 
> > Right, we should not use File.separatorChar. Archive file format is not os specific
> 
> The original algorithm was using File.separatorChar, not `/`. I think it's still possible to craft a tar with `\` as the path separator. Converting the entry name with separatorsToSystem, and then using File.separatorChar will work. We could also modify the condition to something like
> 
> ```
> if (name.charAt(index) != '/' && name.charAt(index) != '\\') {
>  ...
> }
> ```

To me, checking that the path resolves to be within the destination is the more correct change rather than modifying our custom method. What is a reason to keep this change to a _simplest_ change?

siddvenk:
@si2d 
> To me, checking that the path resolves to be within the destination is the more correct change rather than modifying our custom method. What is a reason to keep this change to a _simplest_ change?

I agree with you. I think we should be checking that the path resolves to be within the destination. I've kept that portion in.

The part that still seems open is whether we want to sanitize the archive entry at all. If it's possible and valid for zip archives to start with `/`, then we probably should be sanitizing each entry to remove leading file separators. Or, we can decide to be more strict and not allow that (which is handled implicitly by ensuring we're only writing under the expected output dir).

siddvenk:
@si2d @frankfliu I have updated the PR based on the above discussions. The changes are now

1. convert the archive entry to system specific (convert path separators to the one used by the current system). we remove leading file separators using File.separatorChar as this should work post the initial conversion.
2. validate that the expected output path for the entry exists under the specified output directory

This will fix the issue with the current code. Some additional things we may consider:
- remove step 1 entirely. This will be more strict than what we have today, but I think that's ok.
- don't replace existing files, or provide an option for user to specify whether to overwrite or not

api/src/main/java/ai/djl/util/ZipUtils.java
```diff
@@ -121,13 +120,31 @@ private static void addToZip(Path root, Path file, ZipOutputStream zos) throws I
...
+                    "Bad archive entry "
+                            + name
+                            + ". Attempted write outside destination "
+                            + destination);
+        }
+        return sanitizedName;
+    }
+
     static String removeLeadingFileSeparator(String name) {
+        String osAwareArchiveEntryName = FilenameUtils.separatorsToSystem(name);
```
frankfliu, line 138: 1. try to avoid dependency on `commons.io`
2. If we always run `sanitizeAndValidateArchiveEntry()`, we only need to remove "/" char here, "\" issue will be caught by `sanitizeAndValidateArchiveEntry()`

siddvenk: why avoid commons.io? we're using it in a few places in this code path already

frankfliu: 1. in `api` module, the only place use commons.io is in `TarUtils.java`, `commons-io` is transient dependency from `commons-compression` recently added in 1.27.x, which is not intention of api project
2. there are customers they don't use `tar` files, they can exclude `commons-compression` from their project. see: #2949

siddvenk:
Updated the PR - i've opted for the more strict approach where entries that start with '/' are invalid. 

We validate that each archive entry will be written to a location under the provided output directory. If it won't (either because it starts with `/`, or containers `..` traversal elements that would go outside the expected dir), we fail. Users should be expected to provide valid archives, and we don't do any special processing on them.

api/src/main/java/ai/djl/util/ZipUtils.java
```diff
@@ -52,12 +52,10 @@ public static void unzip(InputStream is, Path dest) throws IOException {
...
-            String name = removeLeadingFileSeparator(entry.getName());
-            if (name.contains("..")) {
-                throw new IOException("Malicious zip entry: " + name);
-            }
-            set.add(name);
-            Path file = dest.resolve(name).toAbsolutePath();
+            String entryName = entry.getName();
+            validateArchiveEntry(entry.getName(), dest);
+            set.add(entryName);
+            Path file = dest.resolve(entryName).toAbsolutePath();
```
si2d, line 58: A bit subjective, feel free to reject:

I know I was the one who said you can make this void, but what if we return the absolute path instead?

```
Path file = getAbsolutePathInDestination(entry.getName(), dest);
set.add(file.toString());
```

```
static Path getAbsolutePathInDestination(String name, Path destination) throws IOException {
    Path outputPath = destination.resolve(name).normalize();
    if (!outputPath.startsWith(destination.normalize())) {
        throw new IOException(
                "Bad archive entry "
                        + name
                        + ". Attempted write outside destination "
                        + destination);
    }
    return outputPath;
}
```

si2d (approved):

api/src/main/java/ai/djl/util/ZipUtils.java
```diff
@@ -121,14 +119,15 @@ private static void addToZip(Path root, Path file, ZipOutputStream zos) throws I
...
 
-    static String removeLeadingFileSeparator(String name) {
-        int index = 0;
-        for (; index < name.length(); index++) {
-            if (name.charAt(index) != File.separatorChar) {
-                break;
-            }
+    static void validateArchiveEntry(String name, Path destination) throws IOException {
+        Path expectedOutputPath = destination.resolve(name).normalize();
+        if (!expectedOutputPath.startsWith(destination.normalize())) {
```
frankfliu, line 124: I think we'd better block ".." as well in this method. it prevent file overwrite inside the destination folder. In original version, we already blocking "..", and nobody complained about it.

siddvenk: The default behavior of extracting a tar is to overwrite isn't it? I'm not sure why we need to differ.

If an archive had (in order)
```
b.txt
a/../b.txt
```

a/../b.txt would overwrite b.txt

siddvenk: I've gone ahead and added it back to keep in line with what we had before, but curious to know what you think about my point above.

frankfliu (approved):

PR #3558, opened 2024-12-10
[xgb] Fix xgb intern to close replaced array

ewan0x79:
## Description ##

close replaced array for xgb engine

### DIFF
--- a/engines/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
+++ b/engines/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
@@ -92,7 +92,8 @@ public class XgbNDArray extends NDArrayAdapter {
             throw new IllegalArgumentException("The replaced NDArray cannot be null.");
         }
         if (!(replaced instanceof XgbNDArray)) {
-            throw new IllegalArgumentException("The replaced NDArray must be an instance of XgbNDArray.");
+            throw new IllegalArgumentException(
+                    "The replaced NDArray must be an instance of XgbNDArray.");
         }
         XgbNDArray array = (XgbNDArray) replaced;
 
--- a/engines/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
+++ b/engines/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
@@ -100,8 +100,10 @@ public class XgbNDArray extends NDArrayAdapter {
             throw new IllegalArgumentException("This target array is already closed");
         }
 
-        long pointer = handle.getAndSet(0L);
-        JniUtils.deleteDMatrix(pointer);
+        if (handle != null && handle.get() != 0L) {
+            long pointer = handle.getAndSet(0L);
+            JniUtils.deleteDMatrix(pointer);
+        }
         if (alternativeArray != null) {
             alternativeArray.close();
         }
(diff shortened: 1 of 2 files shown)


engines/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
```diff
@@ -88,14 +88,36 @@ public ByteBuffer toByteBuffer(boolean tryDirect) {
     /** {@inheritDoc} */
     @Override
     public void intern(NDArray replaced) {
-        if (handle != null && handle.get() != 0L) {
-            long pointer = handle.getAndSet(0L);
-            JniUtils.deleteDMatrix(pointer);
+        if (replaced == null) {
```
frankfliu, line 91: This is not necessary, the `instanceof` will check null:
```
if (!(replaced instanceof XgbNDArray))
```

engines/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
```diff
@@ -88,14 +88,36 @@ public ByteBuffer toByteBuffer(boolean tryDirect) {
...
+        if (!(replaced instanceof XgbNDArray)) {
+            throw new IllegalArgumentException(
+                    "The replaced NDArray must be an instance of XgbNDArray.");
         }
         XgbNDArray array = (XgbNDArray) replaced;
-        data = array.data;
-        handle = array.handle;
-        format = array.format;
+
+        synchronized (this) {
```
frankfliu, line 100: NDArray operation is not thread-safe. Adding synchronized here won't improve thread-safty

engines/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
```diff
@@ -88,14 +88,36 @@ public ByteBuffer toByteBuffer(boolean tryDirect) {
...
+            if (handle != null && handle.get() != 0L) {
+                long pointer = handle.getAndSet(0L);
+                JniUtils.deleteDMatrix(pointer);
+            }
+
+            data = array.data;
+            format = array.format;
+
+            if (array.handle != null && array.handle.get() != 0L) {
```
frankfliu, line 109: Not sure about this null check, if we want to prevent a closed NDArray, we should just do:
```
if (this.closed()
```

engines/ml/xgboost/src/main/java/ai/djl/ml/xgboost/XgbNDArray.java
```diff
@@ -88,14 +88,36 @@ public ByteBuffer toByteBuffer(boolean tryDirect) {
...
+                    handle = new AtomicLong();
+                }
+                handle.set(array.handle.getAndSet(0L));
+            }
+        }
+
+        array.data = null;
+        array.handle = null;
+        array.format = null;
+        array.close();
```
frankfliu, line 120: The memory leak issue may caused by alternativeArray leak, I think we should close current array's alternative array as will. The code can be re-implemented as the following:

```
        if (!(replaced instanceof XgbNDArray)) {
            throw new IllegalArgumentException(
                    "The replaced NDArray must be an instance of XgbNDArray.");
        }
        XgbNDArray array = (XgbNDArray) replaced;
        if (isReleased()) {
            throw new IllegalArgumentException("This array is already closed");
        }
        if (replaced.isReleased()) {
            throw new IllegalArgumentException("This target array is already closed");
        }

        long pointer = handle.getAndSet(0L);
        JniUtils.deleteDMatrix(pointer);
        if (alternativeArray != null) {
            alternativeArray.close();
        }

        data = array.data;
        handle = array.handle;
        format = array.format;
        alternativeArray = array.alternativeArray;
        array.handle = null;
        array.alternativeArray = null;
        array.close();

```

frankfliu (approved):

xyang16 (approved):

PR #3571, opened 2024-12-30
[tokenizers] Add int32 option to encoding

xyang16:
## Description ##

### DIFF
--- a/extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/Encoding.java
+++ b/extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/Encoding.java
@@ -13,6 +13,7 @@
 package ai.djl.huggingface.tokenizers;
 
 import ai.djl.huggingface.tokenizers.jni.CharSpan;
+import ai.djl.ndarray.NDArray;
 import ai.djl.ndarray.NDList;
 import ai.djl.ndarray.NDManager;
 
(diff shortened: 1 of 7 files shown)


frankfliu (approved):

frankfliu:
Fixes: #3566

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/Encoding.java
```diff
@@ -55,23 +56,85 @@ protected Encoding(
...
+        int[][] attentionMask = new int[encodings.length][];
+        int[][] typeIds = new int[encodings.length][];
+        for (int i = 0; i < encodings.length; i++) {
+            ids[i] = Arrays.stream(encodings[i].getIds()).mapToInt(l -> (int) l).toArray();
+            attentionMask[i] =
+                    Arrays.stream(encodings[i].getAttentionMask()).mapToInt(l -> (int) l).toArray();
+            if (withTokenType) {
+                typeIds[i] =
+                        Arrays.stream(encodings[i].getTypeIds()).mapToInt(l -> (int) l).toArray();
+            }
```
siddvenk, line 101: This looks like the only part of this function that differs based on dtype -> can we refactor so only the logic in the the loop differs?

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/Encoding.java
```diff
@@ -55,23 +56,85 @@ protected Encoding(
...
+     * Returns the {@link NDList} representation of the encodings.
+     *
+     * @param encodings the {@code Encoding} batch
+     * @param manager the {@link NDManager} to create the NDList
+     * @param withTokenType true to include the token type id
+     * @param int32 true to use int32 datatype
+     * @return the {@link NDList}
+     */
+    public static NDList toNDList(
+            Encoding[] encodings, NDManager manager, boolean withTokenType, boolean int32) {
```
siddvenk, line 69: could we leverage the instance method toNdList here? It seems like we should be able to loop over all the encondings and call encoding.toNdList(...)?

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/Encoding.java
```diff
@@ -55,23 +56,85 @@ protected Encoding(
         this.overflowing = overflowing;
     }
 
+    /**
+     * Returns the {@link NDList} representation of the encodings.
+     *
+     * @param encodings the {@code Encoding} batch
+     * @param manager the {@link NDManager} to create the NDList
+     * @param withTokenType true to include the token type id
+     * @param int32 true to use int32 datatype
```
siddvenk, line 65: for my understanding, why do we have a boolean for int32? Should these methods accept a dtype argument for more flexibility? Is there an issue with Rust/Candle that prevents us from doing so?

frankfliu: the token ids are always integer, it actually should always be int64, but some of the rust implementation currently only accept int32, that's a workaround for candle.

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/Encoding.java
frankfliu: the variable data type are different (`ids`, `attentionMask`, `typeIds`), they has to be declared separately. I'm not sure if there is easy way to combine them in the loop

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/Encoding.java
frankfliu: They are different, the static one is batched (`Encoding[]`)

siddvenk: yes, but per encoding they are the same. Why can't we do something like this?

```
public static NDList toNDList(Encoding[] encodings, NDManager manager, boolean withTokenType, boolean int32) {
    NDList list = new NDList();
    for (int i = 0; i < encodings.length; i++) {
        NDList encoding = encodings[i].toNDList(manager, withTokenType, int32);
        list.addAll(encoding);
    }
    return list;
}
```

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/Encoding.java
siddvenk: oh yea i didn't catch that. makes sense

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/Encoding.java
frankfliu: This is an optimization for performance (was trying to compete with TEI). It reduced number of memory copies to create a batched NDArray. This way it only invoke JNI 3 times, while the loop one will copy N times and then use stack to batchify them.

I actually not sure if this is necessary.

siddvenk (approved):

siddvenk:
> @siddvenk Any more comments?

lgtm, thanks for the explanations!

PR #3579, opened 2025-01-14
[api] Adds ZeroShotClassification support

bryanktliu:
## Description ##

Brief description of what this PR is about

- If this change is a backward incompatible change, why must this change be made?
- Interesting edge cases to note here

PR #3582, opened 2025-01-15
fix issue with predictor not validating null NDList

siddvenk:
## Description ##

This is a temporary workaround to immediately unblock CI on the djl-serving side. The issues we are facing on djl-serving were introduced in https://github.com/deepjavalibrary/djl/pull/3579, but i'm not clear on why the changes to Predictor introduced in that change are needed. 

It also seems that NDList in predictInternal can be null, but that hasn't been an issue so far from what I can tell. This should (hopefully) only be needed as a temporary fix to unblock some PRs on the serving side

### DIFF
--- a/api/src/main/java/ai/djl/inference/Predictor.java
+++ b/api/src/main/java/ai/djl/inference/Predictor.java
@@ -144,7 +144,10 @@ public class Predictor<I, O> implements AutoCloseable {
     protected NDList predictInternal(TranslatorContext ctx, NDList ndList)
             throws TranslateException {
         logger.trace("Predictor input data: {}", ndList);
-        if (ndList.isEmpty()) {
+        // TODO: this check was (partially) introduced in
+        // https://github.com/deepjavalibrary/djl/pull/3579,
+        // but should ideally not be needed.
+        if (ndList != null && ndList.isEmpty()) {
             return new NDList();
         }
         return block.forward(parameterStore, ndList, false);


xyang16 (approved):

api/src/main/java/ai/djl/inference/Predictor.java
```diff
@@ -144,7 +144,10 @@ public O predict(I input) throws TranslateException {
     protected NDList predictInternal(TranslatorContext ctx, NDList ndList)
             throws TranslateException {
         logger.trace("Predictor input data: {}", ndList);
-        if (ndList.isEmpty()) {
+        // TODO: this check was (partially) introduced in
+        // https://github.com/deepjavalibrary/djl/pull/3579,
+        // but should ideally not be needed.
+        if (ndList != null && ndList.isEmpty()) {
```
frankfliu, line 150: Looks like some `Translator` returned null in djl-serving test.

The question here would be:
1. does `Translator.preprocessInput()` allows null return?
2. does `Translator.preprocessInput()` allows empty NDList return?

Ideally, NDList is a collection type, we should not allow null return, and we can use empty NDList to indicate bypass block forwarding.

frankfliu: Looks like `EchoTranslator` returns null: https://github.com/deepjavalibrary/djl-serving/blob/7315729019480b004784b3f38c474509e2953e0e/serving/src/test/java/ai/djl/serving/translator/EchoTranslator.java#L40

frankfliu: I thought it over, we we 2 use cases here:
1. We want to bypass block forward, this is a processing only predictor, if `ndlist == null`, we should just return null. 
2. The model block has no input parameters (rare but valid),  we should just pass ndlist down to the block

If you agree above proposal, we should make the following changes:
1. We should change the code here to:
```
     if (ndList == null) {
           return null;
     }
```
2. in [ZeroShotClassificationTranslator](https://github.com/deepjavalibrary/djl/blob/a4e61eaa58b3fc35d4f6ce5d10e1982e5d63f99f/extensions/tokenizers/src/main/java/ai/djl/huggingface/translator/ZeroShotClassificationTranslator.java#L64), change `return new NDList();` to:
```
return null;
```

3. optionally, change `EchoTranslator` to `return new NDList();`, `return null;` in this test case should also work

siddvenk: That sounds good to me @frankfliu - let me raise a PR

siddvenk: Actually, should we return empty NDList from predictInternal when input is null?

PR #3605, opened 2025-02-07
[examples] fixing some dependencies and test requirements

sindhuvahinis:
## Description ##

I was doing some testing the examples with ObjectDetection example, with four of the engines our examples support. PyTorch, TensorFlow, OnnxRuntime and MXNet. 

```
export DJL_DEFAULT_ENGINE=[PyTorch, OnnxRuntime, MXNet, TensorFlow]
./gradlew run && mvn package && mvn exec:java -Dexec.mainClass="ai.djl.examples.inference.cv.ObjectDetection"

```

- mvn package was failing because it did not have the common-io version. The error I got was ` NoClassDefFound org/apache/commons/io/input/UnsynchronizedBufferedReader`
- Some of the tests are not supported by some of the engines, because the operations that were used by these examples are not implemented for these engines. Since the default engine is PyTorch, it was succeeding in the CI. Excluding the tests the are not supported by the engines.

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -39,6 +39,11 @@
             <version>1.9.0</version>
         </dependency>
         <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>2.17.0</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-slf4j2-impl</artifactId>
             <version>2.24.1</version>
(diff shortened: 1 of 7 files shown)


frankfliu (approved):

examples/pom.xml
```diff
@@ -38,6 +38,11 @@
             <artifactId>commons-cli</artifactId>
             <version>1.9.0</version>
         </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
```
ethnzhng, line 42: Why was the missing dependency only discovered now? Wondering if we need to add some more dep verification checks to the CI

frankfliu: This is transient dependency, it should work even without this.

Issue #2844, opened 2023-11-09
Add NDArray delete and diff


squallliu:
Add NDArray delete and diff, just like numpy's np.delete and np.diff

frankfliu:
@saedmanaf Really appreciate your contribution, will take a look.

saedmanaf:
I just wanted to understand how implemented is right, so I can implemented others

danyal-imtiaz99:
Hi! 
Since diff was merged in #3663, I'd like to implement delete. Is this still available?

joaquinescalante23:
Hey! I'd like to pick up the rest of this issue. diff landed in #3663, and meshgrid / delete don't seem to have a PR or assignee yet.

@danyal-imtiaz99 — you mentioned wanting to do delete back in March. If you're still on it, no worries, I'll take meshgrid; if you've moved on, I can do both.

For delete I was thinking NDArray.delete(indices, axis) plus a no-axis overload that flattens first, like np.delete. torch doesn't have a native delete, so I'd build it on top of gather using the kept indices — works on all engines without touching the JNI. Negative indices, duplicates, bounds and empty results handled the numpy way.

For meshgrid, NDArrays.meshgrid(...) returning an NDList felt like the natural place. I can compose it with reshape + broadcast (works everywhere, no JNI), or do it natively with torch::meshgrid like diff and roll — any preference? Same for the default indexing: I'd go with 'ij' like torch and allow 'xy' as an option, but happy to match numpy's default if you'd rather.

Tests would go in NDArrayShapesManipulationOpTest on the pytorch engine. Can split into one PR per op if that's easier to review.

PR #3663, opened 2025-04-08
[pytorch] Add diff support

saedmanaf:
## Description ##

- this is the issue https://github.com/deepjavalibrary/djl/issues/2844


saedmanaf:
@frankfliu can you please take a look at this now?
Also, I'm getting this null pointer exception while running the test case? what could be the root case?
image

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
```diff
@@ -1658,6 +1658,11 @@ public NDArray conj() {
         return JniUtils.conj(this);
     }
 
+    @Override
```
xyang16, line 1661:
```suggestion
    /** {@inheritDoc} */
    @Override
```

xyang16:
LGTM, thanks for adding this!

xyang16 (approved):

PR #3677, opened 2025-04-26
Correct 'Load Model' documentation: method ImageClassificationTranslator.builder().setSynsetArtifactName() does not exist

xinhuagu:
## Description ##

In the [Load models from the local file system](https://djl.ai/docs/load_model.html) section, the provided code example incorrectly uses the method _**.setSynsetArtifactName("synset.txt")**_  It should be  **_.optSynsetArtifactName("synset.txt")_**.


xyang16 (approved):

xyang16:
LGTM, thanks for fixing this!

xinhuagu:
There were two places on the same page where this wrong method was used, and I submitted another commit to fix the second place.  Sorry about that.

Issue #2835, opened 2023-11-03
Add NDArray diagonal


tklinchik:
## NDArray diagonal

It appears as diagonal vector/matrix accessor is missing in the API. Would be great to expose what engines already capable of providing.

`NDArray NDArray.diagonal()`

gad2103:
is this still relevant? if so i'd be interested in picking up a few issues.

PR #3724, opened 2025-06-29
Add NDArray diagonal

dev-jonghoonpark:
## Description ##

Add NDArray diagonal

- In the case of TensorFlow, [DiagPart](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/diag-part) was used instead of [Diag](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/diag) to produce results in the same format as PyTorch.
- In the case of Apache MXNet, It is identified as a retired project, so it was handled with `UnsupportedOperationException` without additional searching.

close https://github.com/deepjavalibrary/djl/issues/2835.

### DIFF
--- a/integration/src/main/java/ai/djl/integration/tests/ndarray/NDArrayOtherOpTest.java
+++ b/integration/src/main/java/ai/djl/integration/tests/ndarray/NDArrayOtherOpTest.java
@@ -527,6 +527,32 @@ public class NDArrayOtherOpTest {
         }
     }
 
+    @Test
+    public void testDiagonal() {
+        try (NDManager manager = NDManager.newBaseManager(TestUtils.getEngine())) {
+            NDArray array = manager.arange(4.0f).reshape(2, 2);
+            NDArray expected = manager.create(new float[] {0f, 3f});
+            Assert.assertEquals(array.diagonal(), expected);
+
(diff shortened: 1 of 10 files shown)


integration/src/main/java/ai/djl/integration/tests/ndarray/NDArrayOtherOpTest.java
```diff
@@ -527,6 +527,31 @@ public void testCumsum() {
...
+            array = manager.arange(9.0f).reshape(3, 3);
+            expected = manager.create(new float[] {1f, 5f});
+            Assert.assertEquals(array.diagonal(1), expected);
+
+            array = manager.arange(9.0f).reshape(3, 3);
+            expected = manager.create(new float[] {3f, 7f});
+            Assert.assertEquals(array.diagonal(-1), expected);
+
+            array = manager.arange(27f).reshape(3, 3, 3);
+            expected = manager.create(new float[][] {{0f, 4f, 8f}, {9f, 13f, 17f}, {18f, 22f, 26f}});
```
frankfliu, line 550: Please format java code, run:
```
./gradlew fJ
```

frankfliu (approved):
Thanks for your contribution

xyang16 (approved):

PR #3757, opened 2025-08-06
Issues related to the DownloadUtils vulnerability

freedom-of-the-mind:
## Description ##

Hello, I'm the person who reported the related vulnerability on huntr.com.

As you mentioned, due to the if (Files.exists(output)) line, existing files are indeed not overwritten. 
- https://github.com/deepjavalibrary/djl/pull/3753

I apologize for the incorrect information in my initial report.

However, since there is no validation for the url and output parameters, an attacker can still create arbitrary files in arbitrary locations. The overwrite prevention only limits part of the attack. This still allows for malicious files to be placed on the server or for large files to be downloaded, potentially leading to denial-of-service conditions. Additionally, depending on the server's configuration, an attacker could download a malicious file to execute a web shell.

I just finished updating and resubmitting the issue I previously reported on huntr.com.
I would appreciate it if you could review this issue once again.

### DIFF
--- a/api/src/main/java/ai/djl/training/util/DownloadUtils.java
+++ b/api/src/main/java/ai/djl/training/util/DownloadUtils.java
@@ -24,6 +24,9 @@ import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.util.zip.GZIPInputStream;
 
+import java.net.MalformedURLException;
+import java.nio.file.InvalidPathException;
+
 /** A utility class downloads the file from specified url. */
 public final class DownloadUtils {
 


xyang16:
@siriusbellatrix Could you please look at the build failures here: https://github.com/deepjavalibrary/djl/actions/runs/16778679781/job/47517204012?pr=3757#step:16:69

api/src/main/java/ai/djl/training/util/DownloadUtils.java
```diff
@@ -86,6 +97,42 @@ public static void download(URL url, Path output, Progress progress) throws IOEx
...
+    /**
+     * [Sample] Validates the safety and validity of a URL.
+     * It now specifically checks if the URL starts with the allowed domain.
+     * @param url The URL object to validate.
+     * @throws IOException If the URL is not from the allowed domain or is unsafe.
+     */
+    private static void validateUrl(URL url) throws IOException {
+        String urlString = url.toString();
+        // Allow only URLs that start with the specific domain.
+        if (!urlString.startsWith("https://djl-ai.s3.amazonaws.com/")) {
```
frankfliu, line 109: We cannot limit the url
1. in our Jupyter notebook, we use DownloadUtils to download images from many different location
2. If our user use Jupyter notebook, they should be able to use `DownloadUtils` do anything they want to
It's up to caller to ensure the URL is validate, not the library.

api/src/main/java/ai/djl/training/util/DownloadUtils.java
```diff
@@ -86,6 +97,42 @@ public static void download(URL url, Path output, Progress progress) throws IOEx
...
+    /**
+     * [Sample] Validates the safety and validity of an output path.
+     * It prevents directory traversal attacks by checking for parent directory access (e.g., ../).
+     * @param output The Path object to validate.
+     * @throws IOException If the path is invalid or unsafe.
+     */
+    private static void validatePath(Path output) throws IOException {
+        try {
+            // Normalize the path and check for ".." to prevent directory traversal attacks.
+            if (output.normalize().toString().contains("..")) {
```
frankfliu, line 128: 1. This validation is not enough, should block absolute path as well if we really want to
2. We have test cases (in other repository) write file to tmp directory, this will break the existing code.

frankfliu (commented):
I don't think this change is necessary

api/src/main/java/ai/djl/training/util/DownloadUtils.java
siriusbellatrix: For security reasons, it is advisable to restrict the use of absolute paths by default.
If absolute paths are required in specific cases, it is safer and clearer to handle those cases with targeted code modifications or exceptions.

Rather than relaxing restrictions globally, prioritizing security by default and allowing exceptions only when necessary is a better approach in terms of maintainability and system integrity.

api/src/main/java/ai/djl/training/util/DownloadUtils.java
siriusbellatrix: While it's understandable that DownloadUtils needs to support a variety of URLs,
that doesn't mean it should allow any URL without restriction.

As a shared utility used by many developers, the library should enforce at least minimal security validation to reduce the risk of misuse or accidental security issues.

Placing the entire burden of validation on the caller can lead to security vulnerabilities and may compromise trust in the library itself. Instead, it's more robust for the library to offer safe defaults, while still allowing flexibility when explicitly needed.

frankfliu: This class is not meant for download model, DJL download model using `Criteria` and `Repository` class.
This class is designed to download arbitrary files for Jupyter notebook usage (per jupyter notebook user's request). It is designed for general purpose file download for Jupyter notebook users.
It just happened to be used in unit test code to download some testing data. There is no usage of this class in DJL own production code at all.

frankfliu: This is direct user facing API, designed for general file download. There is not indirect usage of this class. When developer call this function, they know what they are doing. If they want to save file in `/etc` folder, there is no reason for us to block them.

siriusbellatrix: 1. The DJL official documentation (https://docs.djl.ai/master/docs/demos/jupyter/load_pytorch_model.html) clearly provides an example of using the DownloadUtils class to download model files. The statement "First, use the DownloadUtils to download the model files..." directly contradicts the claim that this class is not meant for model downloads. It seems reasonable to conclude, based on the official documentation, that this class is indeed intended for downloading models.

2. Limiting file size is just one example of a secure coding practice. There are many other techniques available, such as restricting file extensions or validating file content. Therefore, instead of arguing that file size restrictions are inappropriate, it's more important to apply a combination of these security measures to ensure system stability.

3. The term "direct user facing API" is not clearly understood. I would appreciate an explanation. Furthermore, I find it hard to agree with the assumption that a developer using DownloadUtils would intentionally save a model in a critical system folder like /etc for training. It's highly unlikely that anyone would try to save a model in a sensitive system folder (/shadow/, /drivers/, etc...). Developers typically save files in safe, authorized locations, considering security and system stability. Assuming a developer would deliberately try to save a file in a sensitive folder and then arguing against blocking them seems like a risky approach. I believe that restricting access to critical system folders is a safer practice.

siriusbellatrix: According to the documentation, .txt files are used to store class labels for classification purposes. Therefore, if necessary, the .txt extension can also be added to the whitelist.

It’s unlikely that extensions such as .jsp or .exe, which could potentially be used as malicious files, would be required for using PyTorch models. Wouldn’t it be reasonable to manage such extensions through a whitelist?

Ultimately, whether to apply secure coding practices should be decided by the maintainer of the code. If the reasoning is that secure coding is not necessary because we cannot predict how developers will use the DownloadUtils class, then by that logic, no library would be obligated to enforce secure coding—since no library author can fully anticipate how their code will be used.

As mentioned earlier, the current implementation of the DownloadUtils class poses a risk of saving arbitrary files to arbitrary paths. Whether to accept that risk or mitigate it through secure coding should be decided after reviewing the points I’ve raised.

frankfliu: The link is for Jupyter notebook. There isn't an built-in command that has a nice UI in Java version of Jupyter notebook. That's why we create this wrapper. You can consider this is a Jupyter notebook version of `curl` command. I didn't see any reason to restrict an `curl` command.

`direct user facing` means, user use this command interactively in Jupyter notebook. It's not indirectly used when calling some other functions.

We do have many cases to use this to download executables (in model server example jupyters), here is an example: we use `DownloadUtils` to download `awscurl` command (anther tool we make) and same it in `/usr/local/bin` folder, and we use it for benchmark

If we rename the class to `FileUtils` and copy apach `FileUtils`'s implementation in our library (or use shadow to rename the package to `ai.djl` and include in `djl.jar` file), do you still consider that as a security vulnerability?

siriusbellatrix: The discussion seems to be going in circles, and my opinion remains the same. If a feature is needed, I believe it's the library developer's responsibility to provide a minimum level of security to the end developer.

I think it's right to implement at least some secure coding practices. As I've said, no developer would intentionally save a model in a critical system folder like /shadow/ or /drivers/ for model training. Similarly, no developer wants files with malicious extensions like .jsp or .exe to be downloaded. So, couldn't a basic whitelist be provided at the very least?

The developer keeps insisting that the person calling this function is solely responsible for input validation. If we follow that logic, then there would've been no reason to issue a security update for the 2021 Log4j vulnerability, would there? After all, the library developer doesn't know how a user will call the library or what functionality they'll need.

siriusbellatrix: The discussion seems to be going in circles. My opinion remains the same: if a feature is needed, I believe it is the library developer’s responsibility to provide at least a minimum level of security to the developer.

I think it is right to implement at least some basic security measures (secure coding). As I mentioned earlier, no developer would intentionally save a model in critical system folders like /shadow/ or /drivers/ for model training, and no developer would want files with potentially malicious extensions like .jsp or .exe to be downloaded. In that case, couldn’t at least a whitelist be provided for these?

The developer keeps insisting that the person calling this function is fully responsible for input validation. But if we follow that logic, then there wouldn’t have been any reason to release a security update for the 2021 Log4j vulnerability, would there?

Why do so many forum-style libraries restrict major scripts with a whitelist to prevent XSS, while still providing the source code to developers? Shouldn’t they allow all scripts since developers might need them?

frankfliu: First of all, I don't think this is a security vulnerability. If DJL use this class internally and didn't do input validation, and can cause unexpected behavior (log4j case), then it's a vulnerability, but clearly this is NOT the case. If the side effect is not obvious and can cause damage (XSS case), then we should consider to improve it. But `DownloadUtils` api is very obvious.

If we release a jar that only contains `DownloadUtils`, do you still think this is a security vulnerability?

We have a lot of powerful features, like load a shared library from arbitrary location, execute arbitrary code (model is code), even reboot GPU. It's up to developer to ensure a trusted environment. If we follow your logic, all those features itself is security vulnerability.

siriusbellatrix: if you believe that using external input values is the responsibility of the caller or developer, and that this is the intended purpose of the design, then I don’t think further discussion will convince you that this is a vulnerability.
I only hope that all developers or callers using DownloadUtils are safe and have no malicious intent.

PR #3785, opened 2025-08-29
Fix preference of tokenizer_config.json and remove doLowerCase from TokenizerConfig

Soha-Agarwal:
# Fix preference of tokenizer_config.json and remove doLowerCase from TokenizerConfig

## Description

This PR improves the HuggingFace tokenizer configuration handling by:

1. **Fixing configuration precedence**: Options now take priority over `tokenizer_config.json` values, allowing runtime overrides of config file settings
2. **Removing doLowerCase from TokenizerConfig**: The `doLowerCase` parameter is now handled exclusively through options, simplifying the configuration model
3. **Adding modelMaxLength support**: Users can now set `modelMaxLength` via options with proper fallback to config values

## Changes

- **Enhanced `applyConfig()` method**: Only applies config values when not explicitly set in options
- **Improved parameter precedence**: Options → TokenizerConfig → Defaults
- **Better modelMaxLength handling**: Supports runtime override of config's `model_max_length`

## Backward Compatibility

This change is backward compatible. Existing code will continue to work as before, but now has additional flexibility to override config file values at runtime.

## Edge Cases

- When `modelMaxLength` is set in both options and config, options take precedence
- Config values are only applied if the corresponding option key is not present
- Maintains existing default behavior when neither options nor config specify values

Reference Discussion - https://github.com/deepjavalibrary/djl/discussions/3730

### DIFF
--- a/extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
+++ b/extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
@@ -100,31 +100,36 @@ public final class HuggingFaceTokenizer extends NativeResource<Long> implements 
             modelMaxLength = 512;
         }
         if (config != null) {
-            applyConfig(config);
+            applyConfig(config, options);
         }
         updateTruncationAndPadding(padInfo);
     }
 
-    private void applyConfig(TokenizerConfig config) {
-        this.modelMaxLength = config.getModelMaxLength();
-        if (config.hasExplicitDoLowerCase() && config.isDoLowerCase()) {
-            this.doLowerCase = Locale.getDefault();
+    private void applyConfig(TokenizerConfig config, Map<String, String> options) {
+        if (options != null && !options.containsKey("modelMaxLength")) {
+            this.modelMaxLength = config.getModelMaxLength();
         }
         this.cleanupTokenizationSpaces = config.isCleanUpTokenizationSpaces();
-        if (Stream.of(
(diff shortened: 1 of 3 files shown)


extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
```diff
@@ -61,13 +62,15 @@ public final class HuggingFaceTokenizer extends NativeResource<Long> implements
     private boolean cleanupTokenizationSpaces;
     private boolean stripAccents;
     private boolean addPrefixSpace;
+    private final Map<String, String> options;
```
frankfliu, line 65: We don't need keep options as a member variable, we can pass it down to `applyConfig()`

extensions/tokenizers/src/main/java/ai/djl/huggingface/tokenizers/HuggingFaceTokenizer.java
```diff
@@ -61,13 +62,15 @@ public final class HuggingFaceTokenizer extends NativeResource<Long> implements
...
 
     private HuggingFaceTokenizer(
             long handle,
             Map<String, String> options,
             TokenizerConfig config,
             PadTokenResolver.PadInfo padInfo) {
         super(handle);
+        this.options = options != null ? new HashMap<>(options) : new HashMap<>();
```
frankfliu, line 73: We don't need create an empty HashMap, just check null in `applyConfig()`

xyang16 (approved):

Issue #3799, opened 2025-09-16
Add percentile implementation for Pytorch engine


tipame:
## Description
Thre is unimplemented percentile method for PtNDArray. 
At the same time Torch library has quantile moethod (which actually provides the same functionality)

Will this change the current api? How?
No

Who will benefit from this enhancement?
humanity

## References
https://docs.pytorch.org/docs/stable/generated/torch.quantile.html

PR #3800, opened 2025-09-16
Added percentile implementation for Pytorch engine

tipame:
## Description ##
https://github.com/deepjavalibrary/djl/issues/3799


frankfliu (commented):
Really appreciate your contribution. This looks great.

We don't have integration test for this function. would you mind add an test in `NDArrayElementComparisonOpTest`?

tipame:
I'll add a test

tipame:
It's not clear - the integration test NDArrayElementComparisonOpTest looks like something general, but an implementation has been added only for the Pytorch engine.

frankfliu (approved):

frankfliu:
@tipame 

You are right, there is bug in build.gradle.kts file caused by gradle 9.0.0 upgrade.

Would you please apply the following patch:

```
diff --git a/engines/pytorch/pytorch-native/build.gradle.kts b/engines/pytorch/pytorch-native/build.gradle.kts
index 4e00b1160..b6b989a77 100644
--- a/engines/pytorch/pytorch-native/build.gradle.kts
+++ b/engines/pytorch/pytorch-native/build.gradle.kts
@@ -65,6 +65,10 @@ tasks {
             }
             into(ciDir)
         }
+        val dll = ciDir / "Release/djl_torch.dll"
+        if (dll.exists()) {
+            dll.renameTo(ciDir / "djl_torch.dll")
+        }
     }
 
     fun downloadBuildAndroid() {
```

tipame:
I applied the patch but now it crashed on the Ubuntu build. I didn't understand what the problem was from the logs.
Is there a forced rebuild button here?

PR #3818, opened 2025-12-05
NDManager.create(Number) unsupported type error message improvement

fracpete:
The error message for an unsupported conversion now specifies the type of the supplied Number object rather than assuming Short (in case of BigDecimal etc)

## Description ##

When creating a network from Groovy, I supplied `1e-10` to a method which Groovy converted implicitly into a `BigDecimal` type. However, the error message that was thrown by `NDManager.create(Number)` stated incorrectly `Short conversion not supported!`. Only after looking at the DJL code did I figure out what was going wrong. 
To fix the type problem, I needed to use `Double.valueOf(1e-10)` in the method call to force a conversion to `java.lang.Double`.
To avoid future confusion, the error message now states the actual type rather than assuming `java.lang.Short`.

### DIFF
--- a/api/src/main/java/ai/djl/util/Utils.java
+++ b/api/src/main/java/ai/djl/util/Utils.java
@@ -501,8 +501,25 @@ public final class Utils {
     /**
      * Opens a connection to this URL and returns an InputStream for reading from that connection.
      *
-     * @param url the url to open
-     * @param headers the HTTP headers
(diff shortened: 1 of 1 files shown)


xyang16 (approved):

fracpete:
Apologies, I edited that file in a text editor and didn't run "gradlew formatJava" beforehand. The latest commit fixes that.

PR #3834, opened 2026-02-01
Fix the broken first link in the 'Quick start' documentation

bhq12:
## Description ##

Hi - I was just getting started with DJL and working through the documentation.

I noticed that literally the first link users see when visiting the quick start documentation at https://docs.djl.ai/master/docs/quick_start.html  (to the "beginner tutorial" ) currently sends users to a nonexistent page / 404 due to pointing at README.md instead of the generated mkdocs html page.

I have just bumped the link to point to the same location as the other link to the same tutorial on line 23

### DIFF
--- a/docs/quick_start.md
+++ b/docs/quick_start.md
@@ -1,7 +1,7 @@
 # Quick start
 
 Deep Java Library (DJL) is designed to be easy to get started with and simple to use.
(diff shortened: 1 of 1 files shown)


xyang16 (approved):
LGTM. Thanks for contributing!

PR #3864, opened 2026-06-19
[google-ai] Add Google Cloud Storage repository support

CodeYogiCo:
## Description

Adds a new `google-ai` extension that lets DJL load models directly from a Google Cloud Storage (GCS) bucket using `gs://` URLs, mirroring the existing `aws-ai` (`s3://`) extension.

### What's included
- **`GcsRepository`** — a `Repository` backed by the Google Cloud `Storage` client. Parses `gs://bucket/prefix` URLs, lists objects (with archive auto-extraction and `model_name` / `artifact_id` query-param support), and streams downloads via `Blob.reader()`. Modeled 1:1 on `S3Repository`.
- **`GcsRepositoryFactory`** — registered via `META-INF/services/ai.djl.repository.RepositoryFactory` for auto-discovery; handles the `gs` scheme and falls back to Application Default Credentials when no `Storage` client is supplied.
- **Tests** — `GcsRepositoryTest` and `GcsRepositoryFactoryTest` using Mockito (run offline, no GCP credentials required).
- **Build wiring** — registered `:extensions:google-ai` in `settings.gradle.kts`; added the `com.google.cloud:libraries-bom` and `google-cloud-storage` entries to the version catalog.
- **Docs** — README for the extension plus references in `docs/README.md`, `docs/load_model.md`, `docs/mkdocs.yml`, and `docs/development/dependency_management.md`.

### Usage
```java
Criteria<Image, Classifications> criteria =
    Criteria.builder()
            .optModelUrls("gs://my-bucket/models/resnet18")
            .optModelName("resnet18")
            .build();
ZooModel<Image, Classifications> model = criteria.loadModel();
```

### Testing
`./gradlew :extensions:google-ai:build` passes (compile, tests, checkstyle, PMD, format verification).

> Note: the GCS BOM is pinned to `26.50.0`; happy to adjust to a preferred version. Tests are Mockito-based rather than hitting a live public bucket — can add an integration test if there's a public GCS bucket with test models.


extensions/google-ai/src/main/java/ai/djl/google/cloud/GcsRepository.java
```diff
@@ -0,0 +1,215 @@
...
+            throw new IOException("Failed scan gcs bucket: " + bucket, e);
+        }
+    }
+
+    private Artifact listFiles() {
+        Page<Blob> blobs =
+                storage.list(
+                        bucket,
+                        Storage.BlobListOption.prefix(prefix),
+                        Storage.BlobListOption.currentDirectory(),
```
CodeYogiCo, line 189: `currentDirectory()` makes GCS list with a `/` delimiter, so only objects directly under the prefix are returned (subdirectories come back as zero-byte `name/` blobs, which are filtered out below). This matches the `delimiter("/")` behavior in `S3Repository#listFiles`.

extensions/google-ai/src/main/java/ai/djl/google/cloud/GcsRepository.java
```diff
@@ -0,0 +1,215 @@
...
+    protected void download(Path tmp, URI baseUri, Artifact.Item item, Progress progress)
+            throws IOException {
+        String key = item.getUri();
+        logger.debug("Downloading artifact from: gs://{}/{} ...", bucket, key);
+        Blob blob = storage.get(BlobId.of(bucket, key));
+        if (blob == null) {
+            throw new IOException("Object not found: gs://" + bucket + '/' + key);
+        }
+        try (ReadChannel reader = blob.reader();
+                InputStream is = Channels.newInputStream(reader)) {
```
CodeYogiCo, line 139: GCS exposes a `ReadChannel` rather than an `InputStream`, so we adapt it via `Channels.newInputStream` to reuse the shared `AbstractRepository#save` download path. Both the channel and the stream are closed by the try-with-resources.

extensions/google-ai/src/main/java/ai/djl/google/cloud/GcsRepository.java
```diff
@@ -0,0 +1,215 @@
...
+                item.setUri(key);
+                item.setSize(blob.getSize() == null ? 0 : blob.getSize());
+                item.setArtifact(artifact);
+                if ("dir".equals(item.getType())) {
+                    item.setName(""); // avoid creating extra folder
+                }
+                files.put(key, item);
+            }
+        }
+        if (files.isEmpty()) {
```
CodeYogiCo, line 209: The S3 version returns null when the raw listing is empty; here we check after filtering directory placeholders, so a prefix that contains only subdirectories also resolves to no artifact (returns null) rather than an empty one.

extensions/google-ai/src/main/java/ai/djl/google/cloud/GcsRepositoryFactory.java
```diff
@@ -0,0 +1,61 @@
...
+    /** {@inheritDoc} */
+    @Override
+    public Repository newInstance(String name, URI uri) {
+        String scheme = uri.getScheme();
+        if (!"gs".equalsIgnoreCase(scheme)) {
+            throw new IllegalArgumentException("Invalid gcs url: " + uri);
+        }
+
+        if (storage == null) {
+            storage = StorageOptions.getDefaultInstance().getService();
```
CodeYogiCo, line 51: When no `Storage` client is injected, we fall back to Application Default Credentials. Users who need custom credentials/project (or anonymous access to a public bucket) can register `new GcsRepositoryFactory(storage)` with a preconfigured client — documented in the extension README.

gradle/libs.versions.toml
```diff
@@ -25,6 +25,7 @@ commonsCsv = "1.14.1"
 commonsLogging = "1.3.5"
 
 awssdk = "2.32.19"
+googleCloud = "26.50.0"
```
CodeYogiCo, line 28: Pinned to the Google Cloud `libraries-bom` `26.50.0`. Happy to bump to whatever the maintainers prefer to keep aligned with the rest of the catalog.

CodeYogiCo (commented):
A few inline notes to help reviewers walk through the new `google-ai` extension. The design intentionally mirrors `aws-ai` so the two stay easy to compare.

CodeYogiCo:
@zachgk thank you for  having a look. I have resolved all review comments. Thank you again for your time and patience

CodeYogiCo:
@zachgk is there anything that can help me expediate this

CodeYogiCo:
@zachgk have fixed those build failures. was a silly mistake at my end.. would u approve the workflow again

xyang16 (approved):

PR #3870, opened 2026-08-01
fix(onnxruntime): allow sub-managers after NDManager.cap()

### LINKED ISSUE #3837: Onnx Engine's NDManager cap() prevents subManagers being created
## Description
The javadoc for NDManager::cap says that sub-managers can still be created (makes sense):

https://github.com/deepjavalibrary/djl/blob/3edd20adb4142e883f79a3fc962a09a9d0d43e73/api/src/main/java/ai/djl/ndarray/NDManager.java#L1550

The Pytorch Engine is following this by using `attachUncappedInternal`:

https://github.com/deepjavalibrary/djl/blob/3edd20adb4142e883f79a3fc962a09a9d0d43e73/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDManager.java#L185

But the Onnx engine uses  `attachInternal` which throws when cap()ed:

https://github.com/deepjavalibrary/djl/blob/3edd20adb4142e883f79a3fc962a09a9d0d43e73/engines/onnxruntime/onnxruntime-engine/src/main/java/ai/djl/onnxruntime/engine/OrtNDManager.java#L153

### Expected Behavior
Onnx Engine uses `attachUncappedInternal` just like the Pytorch one, allowingf us to create new sub-managers when capped.

arimu1:
## Description ##

`NDManager.cap()` is documented to still allow attaching **sub-managers** while blocking ordinary resource attachment (to catch memory leaks early).

Onnx Runtime's `OrtNDManager.newSubManager` incorrectly used `attachInternal`, which throws when the manager is capped. PyTorch, MXNet, and Rust engines already use `attachUncappedInternal` for sub-managers.

This change aligns Onnx Runtime with that contract:

- After `cap()`, creating arrays on the capped manager still fails (`IllegalStateException`)
- After `cap()`, `newSubManager()` succeeds and arrays can be created on the sub-manager

Fixes #3837

### Testing
- `./gradlew :engines:onnxruntime:onnxruntime-engine:test --tests "ai.djl.onnxruntime.engine.OrtTest"` (JDK 21, macOS aarch64) — all tests PASSED, including new `testCappedManagerAllowsSubManagers`

### DIFF
--- a/engines/onnxruntime/onnxruntime-engine/src/main/java/ai/djl/onnxruntime/engine/OrtNDManager.java
+++ b/engines/onnxruntime/onnxruntime-engine/src/main/java/ai/djl/onnxruntime/engine/OrtNDManager.java
@@ -150,7 +150,7 @@ public class OrtNDManager extends BaseNDManager {
     @Override
     public OrtNDManager newSubManager(Device device) {
         OrtNDManager manager = new OrtNDManager(this, device, env);
-        attachInternal(manager.uid, manager);
+        attachUncappedInternal(manager.uid, manager);
         return manager;
     }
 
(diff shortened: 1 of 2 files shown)


xyang16 (approved):
LGTM. Thanks for your contribution!

PR #3876, opened 2026-09-03
Use 64-bit arithmetic for tensor buffer size calculations

ethnzhng:
Buffer size calculations of the form dataType.getNumOfBytes() * size were performed in 32-bit int. For large element counts these products overflow and wrap to a small or negative value, so BaseNDManager.validateBuffer compared the supplied buffer against the wrong expected size, and the corresponding allocateDirect calls allocated buffers smaller than the shape requires.

Compute these byte counts in 64-bit and narrow with Math.toIntExact so a request that does not fit in an int throws instead of silently wrapping. The same treatment is applied to the sibling allocation paths in the api module and in each engine, and to the (int) shape.size() narrowings in NDManager.zeros/ones and BaseNDManager.truncatedNormal, which could truncate rather than fail.

Behaviour is unchanged for any shape that fits in an int-sized buffer.

Adds regression coverage for the overflowing element counts.

### DIFF
--- a/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
+++ b/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
@@ -177,7 +177,8 @@ public abstract class BaseNDManager implements NDManager {
     /** {@inheritDoc} */
     @Override
     public NDArray truncatedNormal(float loc, float scale, Shape shape, DataType dataType) {
-        int sampleSize = (int) shape.size();
+        // Fail closed rather than silently truncating an element count beyond int range.
+        int sampleSize = Math.toIntExact(shape.size());
         double[] dist = new double[sampleSize];
 
         for (int i = 0; i < sampleSize; i++) {
@@ -461,7 +462,10 @@ public abstract class BaseNDManager implements NDManager {
         }
 
         int remaining = buffer.remaining();
-        int expectedSize = isByteBuffer ? dataType.getNumOfBytes() * expected : expected;
+        // Compute the expected byte count in 64-bit arithmetic. getNumOfBytes() * expected is
+        // int*int, which overflows for large element counts and wraps to a small or negative
+        // value, causing the check below to compare against the wrong size.
+        long expectedSize = isByteBuffer ? (long) dataType.getNumOfBytes() * expected : expected;
         if (remaining < expectedSize) {
             throw new IllegalArgumentException(
                     "The NDArray size is: " + expected + ", but buffer size is: " + remaining);
@@ -470,7 +474,8 @@ public abstract class BaseNDManager implements NDManager {
             logger.warn(
                     "Input buffer size is greater than the NDArray size, please set limit"
                             + " explicitly.");
-            buffer.limit(expectedSize);
+            // Safe narrowing: in this branch expectedSize < remaining <= Integer.MAX_VALUE.
+            buffer.limit((int) expectedSize);
         }
     }
 
--- a/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
+++ b/api/src/main/java/ai/djl/ndarray/BaseNDManager.java
@@ -462,10 +462,8 @@ public abstract class BaseNDManager implements NDManager {
         }
 
         int remaining = buffer.remaining();
-        // Compute the expected byte count in 64-bit arithmetic. getNumOfBytes() * expected is
-        // int*int, which overflows for large element counts and wraps to a small or negative
-        // value, causing the check below to compare against the wrong size.
-        long expectedSize = isByteBuffer ? (long) dataType.getNumOfBytes() * expected : expected;
+        int expectedSize =
+                isByteBuffer ? Math.multiplyExact(dataType.getNumOfBytes(), expected) : expected;
         if (remaining < expectedSize) {
             throw new IllegalArgumentException(
                     "The NDArray size is: " + expected + ", but buffer size is: " + remaining);
@@ -474,8 +472,7 @@ public abstract class BaseNDManager implements NDManager {
             logger.warn(
                     "Input buffer size is greater than the NDArray size, please set limit"
                             + " explicitly.");
-            // Safe narrowing: in this branch expectedSize < remaining <= Integer.MAX_VALUE.
-            buffer.limit((int) expectedSize);
+            buffer.limit(expectedSize);
         }
     }
 
--- a/api/src/main/java/ai/djl/ndarray/NDManager.java
+++ b/api/src/main/java/ai/djl/ndarray/NDManager.java
@@ -788,8 +788,11 @@ public interface NDManager extends AutoCloseable {
      * @see #zeros(Shape, DataType, Device)
      */
     default NDArray zeros(Shape shape, DataType dataType) {
-        int size = (int) shape.size();
-        ByteBuffer bb = allocateDirect(size * dataType.getNumOfBytes());
+        // Both narrowings fail closed: shape.size() is a long and the byte count is computed in
+        // 64-bit, so an element count or buffer size beyond int range throws instead of silently
+        // truncating or wrapping to an undersized allocation.
+        int size = Math.toIntExact(shape.size());
+        ByteBuffer bb = allocateDirect(Math.toIntExact((long) size * dataType.getNumOfBytes()));
         return create(bb, shape, dataType);
     }
 
@@ -817,8 +820,11 @@ public interface NDManager extends AutoCloseable {
      * @return a new instance of {@link NDArray}
      */
     default NDArray ones(Shape shape, DataType dataType) {
-        int size = (int) shape.size();
-        ByteBuffer bb = allocateDirect(size * dataType.getNumOfBytes());
+        // Both narrowings fail closed: shape.size() is a long and the byte count is computed in
+        // 64-bit, so an element count or buffer size beyond int range throws instead of silently
+        // truncating or wrapping to an undersized allocation.
+        int size = Math.toIntExact(shape.size());
+        ByteBuffer bb = allocateDirect(Math.toIntExact((long) size * dataType.getNumOfBytes()));
         for (int i = 0; i < size; ++i) {
             switch (dataType) {
                 case FLOAT16:
--- a/api/src/main/java/ai/djl/ndarray/NDManager.java
+++ b/api/src/main/java/ai/djl/ndarray/NDManager.java
@@ -792,7 +792,7 @@ public interface NDManager extends AutoCloseable {
         // 64-bit, so an element count or buffer size beyond int range throws instead of silently
         // truncating or wrapping to an undersized allocation.
         int size = Math.toIntExact(shape.size());
-        ByteBuffer bb = allocateDirect(Math.toIntExact((long) size * dataType.getNumOfBytes()));
+        ByteBuffer bb = allocateDirect(Math.multiplyExact(size, dataType.getNumOfBytes()));
         return create(bb, shape, dataType);
     }
 
@@ -824,7 +824,7 @@ public interface NDManager extends AutoCloseable {
         // 64-bit, so an element count or buffer size beyond int range throws instead of silently
         // truncating or wrapping to an undersized allocation.
         int size = Math.toIntExact(shape.size());
-        ByteBuffer bb = allocateDirect(Math.toIntExact((long) size * dataType.getNumOfBytes()));
+        ByteBuffer bb = allocateDirect(Math.multiplyExact(size, dataType.getNumOfBytes()));
         for (int i = 0; i < size; ++i) {
             switch (dataType) {
                 case FLOAT16:
--- a/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
+++ b/engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
@@ -257,7 +257,7 @@ public class PtNDArray extends NativeResource<Long> implements NDArray {
             return;
         }
         // int8, uint8, boolean use ByteBuffer, so need to explicitly input DataType
-        ByteBuffer buf = manager.allocateDirect(size * type.getNumOfBytes());
+        ByteBuffer buf = manager.allocateDirect(Math.toIntExact((long) size * type.getNumOfBytes()));
         BaseNDManager.copyBuffer(buffer, buf);
 
         // If NDArray is on the GPU, it is native code responsibility to control the data life cycle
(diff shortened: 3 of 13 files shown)


engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
```diff
@@ -257,7 +257,7 @@ public void set(Buffer buffer) {
             return;
         }
         // int8, uint8, boolean use ByteBuffer, so need to explicitly input DataType
-        ByteBuffer buf = manager.allocateDirect(size * type.getNumOfBytes());
+        ByteBuffer buf = manager.allocateDirect(Math.toIntExact((long) size * type.getNumOfBytes()));
```
Lokiiiiii: **Formatting will fail the build.** This line is 101 columns; the repo enforces 100 (AOSP / `google-java-format`, and `tools/conf/checkstyle.xml` sets `LineLength max=100`). `buildSrc/src/main/kotlin/ai/djl/javaFormatter.gradle.kts` wires `check` to `dependsOn(verifyJava)`, so `./gradlew build` fails here and at `extensions/tokenizers/src/main/java/ai/djl/engine/rust/RsNDArray.java:242`, which is the same 101-column statement.

Measured both at 101 columns on `af0401e`.

Worth noting the identical construct is already wrapped in this same PR at `engines/tensorflow/.../TfNDArray.java:203-204`, so the two long lines look like a missed formatter run rather than a choice.

```
./gradlew formatJava   # alias: fJ
```

Side effect worth flagging: because `verifyJava` runs under `check`, the new `BaseNDManagerTest` regression coverage this PR adds never executes in CI until this is fixed.

---
*Automated review (review-bot, four-agent pass). Posted as a draft for a human to keep, edit, or discard.*

api/src/main/java/ai/djl/ndarray/NDManager.java
```diff
@@ -788,8 +788,11 @@ default NDArray zeros(Shape shape) {
      * @see #zeros(Shape, DataType, Device)
      */
     default NDArray zeros(Shape shape, DataType dataType) {
-        int size = (int) shape.size();
-        ByteBuffer bb = allocateDirect(size * dataType.getNumOfBytes());
+        // Both narrowings fail closed: shape.size() is a long and the byte count is computed in
+        // 64-bit, so an element count or buffer size beyond int range throws instead of silently
+        // truncating or wrapping to an undersized allocation.
+        int size = Math.toIntExact(shape.size());
+        ByteBuffer bb = allocateDirect(Math.toIntExact((long) size * dataType.getNumOfBytes()));
```
Lokiiiiii: **Ten copies of one rule.** `Math.toIntExact((long) size * <dataType>.getNumOfBytes())` now appears at this line and at `NDManager.java:827`, `PassthroughNDManager.java:100-101`, `MxNDArray.java:316`, `PtNDArray.java:260`, `PtNDManager.java:75`, `TfNDArray.java:203-204`, `TfNDManager.java:96`, `RsNDArray.java:242`, `RsNDManager.java:77`.

Two consequences:

1. The invariant "bytes = numOfBytes x elements, and it must fit an int" has no single point of truth, so the next engine added will omit it — and several sibling sites in these same files were in fact left in 32-bit arithmetic by this PR.
2. The resulting `ArithmeticException: integer overflow` names no shape, data type, or requested byte count, so an operator hitting the new failure mode cannot tell which tensor request was too large.

This repo already centralises engine-shared logic of exactly this kind in `BaseNDManager` statics next to `validateBuffer`/`copyBuffer`. A single helper there — throwing `IllegalArgumentException` naming shape, dtype and byte count — would replace all ten sites and give a better message. It would also shorten the two 101-column lines below the formatter limit.

---
*Automated review (review-bot, four-agent pass). Posted as a draft for a human to keep, edit, or discard.*

api/src/main/java/ai/djl/ndarray/BaseNDManager.java
```diff
@@ -461,7 +462,10 @@ public static void validateBuffer(Buffer buffer, DataType dataType, int expected
         }
 
         int remaining = buffer.remaining();
-        int expectedSize = isByteBuffer ? dataType.getNumOfBytes() * expected : expected;
+        // Compute the expected byte count in 64-bit arithmetic. getNumOfBytes() * expected is
+        // int*int, which overflows for large element counts and wraps to a small or negative
+        // value, causing the check below to compare against the wrong size.
+        long expectedSize = isByteBuffer ? (long) dataType.getNumOfBytes() * expected : expected;
```
Lokiiiiii, line 468: **The gate is hardened in its int step but not its long step.** This computes `expectedSize` in 64-bit, which fixes the `int * int` wrap. But `expected` derives from `Shape.size()`, which multiplies its dimensions in `long` with no overflow check (`api/src/main/java/ai/djl/ndarray/types/Shape.java:188-197` — not touched by this PR). A shape whose dimension product is a multiple of 2^64 still wraps, to 0 or a small value, and passes this comparison.

Why it matters here specifically: `validateBuffer` is the boundary that guards deserialised input. An application accepting untrusted `NDArray`/`NDList` bytes can still drive an out-of-bounds native read or a process crash through the exact check this PR is strengthening.

The adjacent comments say these narrowings "fail closed", which holds for the int step only.

Either validate the element product (`Math.multiplyExact` in `Shape.size()`, or an explicit check at this boundary), or — if that is deliberately out of scope — say so and soften the comments to name the one step they actually cover, so the next reader does not trust a stronger guarantee than exists.

Lokiiiiii (changes requested):

engines/pytorch/pytorch-engine/src/main/java/ai/djl/pytorch/engine/PtNDArray.java
Lokiiiiii: **[major] F1:** Two of the changed lines are 101 columns wide, which google-java-format (AOSP, 100-column) will reformat, so the `verifyJava` task that `check`/`build` depends on will fail and CI will go red. Also: Two of the new lines are 101 characters, which trips the repo's own 100-column gates and will fail the build for pytorch-engine and tokenizers. Also: Two of the edited lines are 101 columns, over the AOSP 100-column limit the project enforces via `verifyJava`, while the identical construct elsewhere in this same CR was wrapped. Also: Two lines the CR introduces exceed the repo's 100-column formatter limit, so verifyJava fails the build and the new regression test never runs in CI.

**Evidence:** `grep -nE '^.{101,}$'` across every touched file matches exactly two code lines — PtNDArray.java:260 and RsNDArray.java:242 — and neither matches `^.{102,}$`, so both are exactly 101 columns. buildSrc/src/main/kotlin/ai/djl/javaFormatter.gradle.kts registers `verifyJava`, which runs `formatter.format("-a", "-n", "--set-exit-if-changed", f)` over `sourceSets.flatMap { it.allSource }` and throws GradleException on any diff, and ends with `named("check") { dependsOn(verifyJava) }`; .github/workflows/continuous.yml runs `./gradlew build`. `-a` selects AOSP style = 4-space indent, 100-column limit. The same statement was handled inconsistently elsewhere in this very diff: MxNDArray.java:316 lands at exactly 100 columns (`bb` instead of `buf`) and TfNDArray.java:203-204 was hand-wrapped onto two lines, so `./gradlew formatJava` clearly was not run. Also: `grep -E '^.{101,}$'` on every changed file returns exactly these two new lines (plus the pre-existing license headers). tools/conf/checkstyle.xml:429-432 sets `LineLength max=100`; buildSrc/.../check.gradle.kts sets `checkstyle { isIgnoreFailures = false }`; buildSrc/.../javaFormatter.gradle.kts registers `verifyJava` (google-java-format `-a -n --set-exit-if-changed`) and wires it into `check`. The author correctly hand-wrapped the identical expression at TfNDArray.java:203-204 and PassthroughNDManager.java:100-101, and MxNDArray.java:316 lands at exactly 100 — only these two overflow (the `buf` vs `bb` identifier adds the extra character). Also: `grep -c -E '^.{101}$'` returns 1 for each of those two files, matching exactly the new lines. buildSrc/src/main/kotlin/ai/djl/javaFormatter.gradle.kts runs google-java-format with `-a` (AOSP, 100 cols) and `--set-exit-if-changed`, and ends with `named("check") { dependsOn(verifyJava) }`; docs/development/development_guideline.md §Coding Conventions says "We also follow the AOSP Java Code Style ... The formatting is checked very strictly." The same expression was wrapped by the formatter in TfNDArray.java:203-204 and PassthroughNDManager.java:100-101, and MxNDArray.java:316 lands at exactly 100. (checkstyle's LineLength module is commented out at tools/conf/checkstyle.xml:424-437, so the gate that fails is verifyJava, not checkstyle.) Also: `grep -n '.\{101,\}'` matches both lines (each is 101 chars: 'ByteBuffer buf = manager.allocateDirect(Math.toIntExact((long) size * type.getNumOfBytes()));'), while the identical MxNDArray.java:316 is exactly 100 and does not match. buildSrc/src/main/kotlin/ai/djl/javaFormatter.gradle.kts runs google-java-format with '-a' (AOSP = 100 cols) and '--set-exit-if-changed', throwing GradleException on any reformat, and registers `named("check") { dependsOn(verifyJava) }`. .github/workflows/continuous.yml runs './gradlew build :jacoco:testCodeCoverageReport' on ubuntu/macos/windows. The only other >100-char lines in these files are the pre-existing license-header block comments, which google-java-format does not reflow.

**Impact:** `./gradlew build` fails on all three CI operating systems, so the PR cannot merge until reformatted. Also: `./gradlew check` (checkstyleMain + verifyJava) fails in two modules, so the fix cannot merge as-is. Also: `./gradlew build` fails at :engines:pytorch:pytorch-engine:verifyJava and :extensions:tokenizers:verifyJava, and the inconsistent wrapping of one identical expression across five files makes the change read as hand-edited rather than formatter-clean. Also: CI fails on all three OSes and the change cannot merge; the new BaseNDManagerTest regression coverage is never exercised by the gate until this is fixed.

**Suggested direction:** Run `./gradlew formatJava` (alias `fJ`) and commit the result; it will wrap those two statements the way TfNDArray.java:203-204 already is. Also: Run `./gradlew formatJava`, or wrap manually the same way TfNDArray.java:203-204 already does: `ByteBuffer buf =\n        manager.allocateDirect(Math.toIntExact((long) size * type.getNumOfBytes()));` Also: Run `./gradlew formatJava` (`fJ`) before pushing; or adopt the helper from L-DES-01, which shortens both lines below the limit without wrapping. Also: Run './gradlew fJ' (formatJava) and commit the reformatting, or hoist the byte count into a local (e.g. 'int nbytes = Math.toIntExact((long) size * type.getNumOfBytes());') so both lines fit.

_Independently reported by 4 reviewers (principal, lens:correctness, lens:design, lens:security)._

---
*Automated review (review-bot). Draft — a human decides whether to submit, edit, or discard.*

api/src/main/java/ai/djl/ndarray/NDManager.java
Lokiiiiii: **[major] F2:** The identical expression `Math.toIntExact((long) size * <dataType>.getNumOfBytes())` is copy-pasted into eight call sites across four engines instead of being added as a shared static helper on `BaseNDManager` alongside `validateBuffer`/`copyBuffer`, and the resulting `ArithmeticException: integer overflow` carries no shape, data type, or requested byte count. Also: The overflow-safe byte-count computation is open-coded at twelve sites across five modules instead of a single shared helper, even though the repo already centralizes exactly this engine-shared logic in BaseNDManager statics.

**Evidence:** `grep -rn 'getNumOfBytes()' --include=*.java` shows ten near-identical `Math.toIntExact((long) size * ...getNumOfBytes())` expressions introduced by this CR. `BaseNDManager` already exists precisely to host this kind of cross-engine buffer helper: `public static void validateBuffer(Buffer, DataType, int)` (BaseNDManager.java:451) and `public static void copyBuffer(Buffer, ByteBuffer)` (BaseNDManager.java:496) are both called from MxNDArray, PtNDArray/PtNDManager, TfNDArray/TfNDManager, RsNDArray/RsNDManager and OrtNDManager. On the diagnostic side, `Math.toIntExact` throws `ArithmeticException("integer overflow")` with no arguments, so a user who calls `manager.zeros(new Shape(1L << 30), DataType.FLOAT32)` now gets a bare `java.lang.ArithmeticException: integer overflow` from inside a default interface method, with nothing naming the shape or the 4294967296-byte request — noticeably worse than the `IllegalArgumentException("The NDArray size is: ..., but buffer size is: ...")` that validateBuffer throws for the analogous case. The comments added at NDManager.java:791-793 and 823-825 also overstate the guard: for a shape with an unknown dimension `Shape.size()` returns -1 (Shape.java:188-197), so the narrowing does not fail closed at all and execution still reaches `allocateDirect(-4)`. Also: The expression `Math.toIntExact((long) size * dataType.getNumOfBytes())` is repeated verbatim (operand order mirrored in NDArrayAdapter:174, variant in NDSerializer.java:57-59, kept as long in BaseNDManager.java:468). At every engine site the line sits within three lines of a call to `BaseNDManager.validateBuffer(...)` and/or `BaseNDManager.copyBuffer(...)` (e.g. PtNDArray.java:247+260-261, TfNDArray.java:198+204-205, MxNDArray.java:310+316-317) — public statics in BaseNDManager are already the established mechanism for logic shared by these exact methods, and BaseNDManager is imported at all of them. Meanwhile sibling byte-size products in the same interface remain 32-bit: NDManager.java:415/430/445/460/475/490 `allocateDirect(data.length * data[0].length * 4)`, plus Float16Utils.java:62, BufferedImageFactory.java:249, LgbmNDManager.java:83/88, XgbSymbolBlock.java:70.

**Impact:** Ten copies of one rule mean the next engine added will forget it; and operators hitting the new failure mode get an exception that does not say which tensor request was too large. Also: The rule 'bytes = numOfBytes x elements, must fit an int' has no single point of truth, so the next reader cannot tell which sites are hardened, the degenerate DataType values (UNKNOWN=0, STRING=-1) are re-derived per site, and any future revision has to be re-applied twelve times across five modules — which is already why several sibling sites in the same files were left in 32-bit arithmetic.

**Suggested direction:** Add one `public static int toBufferSize(long elements, DataType dataType)` (or `(Shape, DataType)`) helper next to `validateBuffer` in `BaseNDManager` that does the 64-bit multiply and throws `IllegalArgumentException` naming shape, data type and byte count, then call it from all ten sites. Also: Add one helper and call it from every site: `public int byteSize(long elements) { return Math.toIntExact(numOfBytes * elements); }` on DataType (natural owner of getNumOfBytes and its 0/-1 edge values), or a `public static int byteSize(DataType, long)` in BaseNDManager next to validateBuffer/copyBuffer. Call sites then read `allocateDirect(type.byteSize(size))`, which also removes the casts and brings both over-long lines back under 100 columns.

_Independently reported by 2 reviewers (principal, lens:design)._

---
*Automated review (review-bot). Draft — a human decides whether to submit, edit, or discard.*

api/src/main/java/ai/djl/ndarray/BaseNDManager.java
xyang16: Since `buffer.limit()` only accepts int, I have updated to `Math.multiplyExact()` that throws an exception if the result overflows an int, rather than silently wrapping.

Lokiiiiii (changes requested):

Lokiiiiii (approved):

PR #3895, opened 2026-09-19
fix(sam2): add multimask_output option to return all candidate masks

### LINKED ISSUE #3794: SAM2Translator only returns one mask instead of the full list of candidates
## Description
ai.djl.modality.cv.translator.SAM2Translator only returns one mask instead of the full list of candidates.

### Expected Behavior
The returned DetectedObjects object should contain a list of all masks detected if the input argument multimask_output is set to true. The current behavior to return the "best" mask is correct if multimask_output is set to false.

### Details
Starting from line 160 of the SAM2Translator source code, it is clear that all masks are retrieved, but in line 168 the result is built as a singleton list, then that singleton list is returned to the user in line 170 with no pathway to returning the full set of masks. 

```
160 NDArray masks = logits.gt(0f).squeeze(0);
161
162 float[][] dist = Mask.toMask(masks.get(best).toType(DataType.FLOAT32, true));
163 Mask mask = new Mask(0, 0, width, height, dist, true);
164 double probability = scores.getFloat(best);
165
166 List<String> classes = Collections.singletonList("");
167 List<Double> probabilities = Collections.singletonList(probability);
168 List<BoundingBox> boxes = Collections.singletonList(mask);
169
170 return new DetectedObjects(classes, probabilities, boxes);

```

Ashfaqbs:
## Summary

`Sam2Translator.processOutput` always picked the single highest-scoring mask (`scores.argMax()`) and discarded the rest, even though the underlying model output (`logits`/`scores`) already contains every candidate mask for the given prompt. There was no way to get the other candidates back out, even though SAM2's own multi-mask behavior is a normal part of its API.

## Fix

Adds a `multimask_output` translator argument (and `Builder.optMultimaskOutput(boolean)`), default `false` so existing behavior/output is unchanged. When set to `true`, `processOutput` returns the full parallel lists of masks and their scores instead of a singleton.

## Test plan

Added `Sam2TranslatorTest` with a stub `LambdaBlock` model returning 3 fixed candidate masks + scores (no real SAM2 weights needed):
- `testDefaultReturnsOnlyBestMask` — confirms unchanged behavior: 1 result, the highest-scoring mask.
- `testMultimaskOutputReturnsAllCandidates` — confirms all 3 masks/scores come back in order when the flag is set.

Both pass locally (`gradle :api:test --tests Sam2TranslatorTest`), and `:api:checkstyleMain`/`:api:checkstyleTest` are clean. Did not run the full module suite (the wider `:api` test target pulls in native engine downloads not practical in this environment) — happy to have CI confirm.

Fixes #3794


Ashfaqbs:
Good catch — fixed in 014c120. Added assertions on each returned object's `Mask.getProbDist()[0][0]`, which is 0/1/0 across the three fixture masks, so the test now fails if every candidate ends up aliasing to the same mask.

